PR #7, opened 2018-09-24
--strip

ijl:
This strips shared libraries through a link arg. It's a common enough case to merit its own flag, I think.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -85,6 +85,7 @@ FLAGS:
     -h, --help               Prints help information
         --release            Pass --release to cargo
         --skip-auditwheel    Don't check for manylinux compliance
+        --strip              Strip the library for minimum file size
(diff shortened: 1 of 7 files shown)


konstin:
I didn't even know that this option existed, but it definitely should be supported!

I've compared the reduction from this and the last PR and the results are truly amazing - The new files are about 25x smaller than the old ones:

wheel | baseline | deflate | strip + deflate
-|-|-|-
get_fourtytwo-2.0.1-cp36-cp36m-manylinux1_x86_64.whl | 2,8M | 771K | 102K
hello_world-0.1.0-py2.py3-none-manylinux1_x86_64.whl | 3,9M | 1,1M | 180K
points-0.1.0-py2.py3-none-manylinux1_x86_64.whl | 2,8M | 752K | 85K

Thank you!

PR #20, opened 2018-11-14
Look for `cpython` instead of `rust-cpython`

jwbuurlage:
This PR does two things:

- It looks for `cpython` as a depencency, since the crate is (renamed to?) [cpython](https://crates.io/crates/cpython).
- It only  passes `--features` flags to `pyo3`, since `rust-cpython`  does not recognize them.

Tested this on my `rust-cpython` package, and called `cargo test` in `pyo3-pack` with no errors.

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -144,7 +144,9 @@ pub fn compile(
 
     let mut shared_args = vec!["--manifest-path", context.manifest_path.to_str().unwrap()];
 
-    if python_feature != "" {
+    if *bindings_crate == BridgeModel::Bindings("pyo3".to_string())
+        && python_feature != ""
+    {
         // This is a workaround for a bug in pyo3's build.rs
         shared_args.extend(&["--features", &python_feature]);
     }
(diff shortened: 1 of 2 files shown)


src/compile.rs
```diff
@@ -144,9 +144,14 @@ pub fn compile(
 
     let mut shared_args = vec!["--manifest-path", context.manifest_path.to_str().unwrap()];
 
-    if python_feature != "" {
-        // This is a workaround for a bug in pyo3's build.rs
-        shared_args.extend(&["--features", &python_feature]);
+    match bindings_crate {
+        BridgeModel::Bindings(name) => {
+            if name == "pyo3" && python_feature != "" {
```
konstin: Nitpick: This could be a single if with `binding_crate == BridgeModel::Bindings("pyo3".to_string())`

jwbuurlage: I want to use fancy Rust pattern matching too badly :smile: I have updated the commit.

PR #52, opened 2018-12-21
Add support for conda envs on windows

paddyhoran:
Hey @konstin,

I have support for building with `conda` on windows working.  I just tested this locally by running `pyo3-pack build` for the `get_fourtytwo` example.

I'm assuming that supporting `conda` on other systems would be pretty much the same but I'm not able to test this.

I'll take a look at updating CI to test this in another PR.

### DIFF
--- a/src/python_interpreter.rs
+++ b/src/python_interpreter.rs
@@ -100,8 +100,12 @@ fn find_all_windows(target: &Target) -> Result<Vec<String>, Error> {
                         .parse::<usize>()
                         .context(context)?;
 
-                    if windows_interpreter_no_build(major, minor, target.pointer_width(), pointer_width)
-                    {
+                    if windows_interpreter_no_build(
+                        major,
+                        minor,
+                        target.pointer_width(),
+                        pointer_width,
+                    ) {
                         continue;
                     }
 
@@ -128,8 +132,8 @@ fn find_all_windows(target: &Target) -> Result<Vec<String>, Error> {
         let lines = str::from_utf8(&output.stdout).unwrap().lines();
         let re = Regex::new(r"(\w|\\|:|-)+$").unwrap();
         let mut paths = vec![];
-        for i in lines.into_iter() {
-            if !i.starts_with("#") {
+        for i in lines {
+            if !i.starts_with('#') {
                 if let Some(capture) = re.captures(&i) {
                     paths.push(String::from(&capture[0]));
                 }
@@ -155,10 +159,14 @@ fn find_all_windows(target: &Target) -> Result<Vec<String>, Error> {
                         32_usize
                     };
 
-                    if windows_interpreter_no_build(major, minor, target.pointer_width(), pointer_width)
-                        {
-                            continue;
-                        }
+                    if windows_interpreter_no_build(
+                        major,
+                        minor,
+                        target.pointer_width(),
+                        pointer_width,
+                    ) {
+                        continue;
+                    }
 
                     interpreter.push(executable);
                     versions_found.insert((major, minor));
--- a/src/python_interpreter.rs
+++ b/src/python_interpreter.rs
@@ -173,7 +173,7 @@ fn find_all_windows(target: &Target) -> Result<Vec<String>, Error> {
         }
 
         for path in paths {
-            let executable = format!(r"{}\python", path);
+            let executable = Path::new(&path).join("python");
             let python_info = Command::new(&executable)
                 .arg("-c")
                 .arg("import sys; print(sys.version)")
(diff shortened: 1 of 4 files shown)


paddyhoran:
Hey @konstin,

Thanks for the feedback.
> What is not clear to me though is if you want to use conda's base interpreter or the one in the currently activated environment

For my use case, using the interpreter in the active env is sufficient, but I can see the value in building for all envs.  For instance, you can easily create a few temp envs in order to build wheels for all the Python versions you need.  I could see this really being useful for someone that maintains a package on pypi (which I do not).

I don't think that using the base interpreter is as useful.  For instance, I install miniconda only to build envs.  I never actually use the base interpreter.

> do we need to filter duplicates

Yes, this is an oversight, I'll fix it when I get a chance along with the clippy issues.

konstin:
I finally got to install conda on windows and play around with it.

It'd like to have the output of `conda info --envs` documented somewhere near the regex parsing it, maybe as an inline comment. I've failed to do that when I added `py -0`, but it will definitely add this eventually (likely after this pull request is merged, to avoid merge conflicts) 

```
>conda info --envs
# conda environments:
#
base                     C:\Users\Konstantin\Anaconda3
foo1                  *  C:\Users\Konstantin\Anaconda3\envs\foo1
```

```
>py -0
Installed Pythons found by py Launcher for Windows
 -3.7-64 *
 -3.6-32
 -2.7-64
```

> Let me know how you feel about building for the active env vs all (I favor all, but obviously your call)?

I like the current solution!

One thing we need is an automated way of testing this, either alongside the other tests or as a standalone script. it should create at least 3 environments, with two having the same version and one a different version, and then check the two versions are picked up correctly, and ideally jsut do nothing if conda is no installed.

src/python_interpreter.rs
```diff
@@ -29,79 +30,155 @@ print(json.dumps({
...
+                }
             }
+        }
 
-            let version = format!("-{}.{}-{}", major, minor, pointer_width);
-
-            let output = Command::new("py")
-                .args(&[&version, "-c", code])
+        for path in paths {
+            let executable = format!(r"{}\python", path);
```
konstin: Nit: Use [Path::join](https://doc.rust-lang.org/nightly/std/path/struct.Path.html#method.join) instead of format

paddyhoran:
Hey @konstin,

I updated to use `Path::join` and I added comments regarding how we find interpreter info, feel free to re-word, etc.

CI was the next step, just wanted to get the functionality in place before I start messing with that :).  Do you want to merge this first or should I keep working on CI on this branch?

konstin:
Thank you, this is one great doc comment!

The Ci for windows, appveyor, only builds the code and creates binaries for releases. The integration tests always failed for mysterious reasons I couldn't reproduce, so I had to [deactivate](https://github.com/PyO3/pyo3-pack/blob/086bb6de7e679d2cbd29f1e3f1cf7de07ba735b0/appveyor.yml#L24-L27) them. That's why CI for windows isn't important as long as the tests can be run locally.

>  Do you want to merge this first or should I keep working on CI on this branch?

I'd like to merge after we have a test.

konstin:
Oh, and the current travis ci failure seems to be because clippy is broken on nightly rust, which should be fixed by a new rust nightly in the next days.

paddyhoran:
Ok, sounds good, thanks very much.

konstin:
I'd recommend solving this without appveyor first. For that you can assume that conda is installed and in path. Then just run conda commands as you've said, check that `cargo run --quiet -- list-python` prints the expected versions. Bonus points for picking one of the test crates, building wheels, installing them in the envs and running `check_installed.py`, but that's not required. 

For the existing interpreters, afaik you can override `py` by some dummy alias (like [here](https://superuser.com/a/560558/586197)) so that the regex wont capture anything. 

> Also, which test should I update or replicate for the conda testing?

Once you have a script, you can try to replicate tests/test_integration.rs if you want.

paddyhoran:
As always, thanks @konstin.

paddyhoran:
Sorry for the lack of activity on this, I plan to get back to it soon.

paddyhoran:
@konstin, I'm just testing but AppVeyor did not trigger even though I made a change to `AppVeyor.yml`?

paddyhoran:
Ok, sounds good, thanks.  Still a WIP but just wanted to how it fairs with CI.

konstin:
The tests look good, but need one addition: We need to check the selected interpreters are actually the conda environments, e.g. by checking that `pyo3-build-env-` is in the interpreter path. Otherwise it could happen that the conda detecting code becomes broken and the test still passes because it chooses the system python versions.  (You can confirm that by running the test in an environment where conda isn't in `PATH` after placing a `dbg!()` around `wheels` in `for (filename, _, python_interpreter) in wheels`.

---

I've tried to get the windows tests working on appveyor in #67, but had only moderate success. I think the best is to merge once we have the additional and leave figuring out appveyor for later.

konstin:
Thank you!

Issue #69, opened 2019-01-23
manylinux2010 / PEP 571


ijl:
pip 19.0 supports the tag: https://pip.pypa.io/en/stable/news/

https://www.python.org/dev/peps/pep-0571/

ijl:
I can implement this.

konstin:
Oh sorry, I though I had already tagged that. I've now released current master as 0.5.0-beta.4. I plan to publish 0.5.0 once https://github.com/eqrion/cbindgen/pull/271 is released, so I can publish on crates.io again.

PR #70, opened 2019-01-23
manylinux2010

ijl:
This enables what was mostly there already, fixing #69. I built a manylinux2010 wheel locally and it installs fine with pip 19.0.

### DIFF
--- a/src/build_context.rs
+++ b/src/build_context.rs
@@ -1,5 +1,5 @@
 #[cfg(feature = "auditwheel")]
-use crate::auditwheel_rs;
+use crate::auditwheel::{auditwheel_rs, MANYLINUX1, MANYLINUX2010};
 use crate::compile;
 use crate::module_writer::WheelWriter;
 use crate::module_writer::{write_bin, write_bindings_module, write_cffi_module};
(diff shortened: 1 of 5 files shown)


src/build_context.rs
```diff
@@ -151,9 +151,21 @@ impl BuildContext {
...
+                        auditwheel_rs(&artifact, MANYLINUX1)
+                            .context("Failed to ensure manylinux compliance")?;
+                    }
+                    Manylinux::Manylinux2010 => {
+                        auditwheel_rs(&artifact, MANYLINUX2010)
+                            .context("Failed to ensure manylinux compliance")?;
+                    }
+                    _ => {}
+                }
+            }
```
konstin: Please move the above into a method

src/build_context.rs
```diff
@@ -204,9 +216,21 @@ impl BuildContext {
...
```
konstin: This should also use the method

konstin:
Thank you! Please add that method, otherwise the PR looks good.

ijl:
Ok.

PR #75, opened 2019-01-30
Fix error message for interpreter not being found

ijl:
```
pyo3-pack build -i none
🔗 Found pyo3 bindings
none doesn't exist
```


konstin:
Thank you!

PR #105, opened 2019-03-26
pypy support

ijl:


src/python_interpreter.rs
```diff
@@ -22,6 +22,10 @@ print(json.dumps({
     "major": sys.version_info.major,
     "minor": sys.version_info.minor,
     "abiflags": sysconfig.get_config_var("ABIFLAGS"),
+    # implementation_lower is only specified by pypy
+    "interpreter": sysconfig.get_config_var("implementation_lower") or "cpython",
+    "ext_suffix": sysconfig.get_config_var("EXT_SUFFIX"),
+    "abi_tag": (sysconfig.get_config_var("SOABI") or "-").split("-")[1] or None,
```
ijl: This can possibly replace the ABI flag logic, but I haven't tried python2.

ijl:
I see this as working fine with pypy3 v7.1 and python3.7.

src/python_interpreter.rs
konstin: I'll drop python2 support anyway, so that's not problem

konstin:
FYI I've added two pypy sysconfigs as reference

omerbenamram:
@konstin both PyPy3.5 6.0.0, and PyPy3.6 7.0.0/7.1.0 aren't fully working.
I haven't added all the stuff that was changed in 3.6 for datetime, and I think there were some symbols that were added only in 7.0.0, and also some additional functionality in the `sysconfig` module which saved me some work when finding `libpypy`.

I'd suggest adding the PyPy3.5 7.0.0 that is currently on the website

konstin:
FWIW at least cargo check did work for me with 3.5 6.0 and 3.6 7.0 did work with the patch I posted on the other thread (I basically disabled a few symbols; I haven't check which of them would actually be supported, but it's easy to add a symbols when the base support is working). But I agree that we should settle on pypy >= 7.0 and I've updated the sysconfig.

src/python_interpreter.rs
```diff
@@ -22,6 +22,10 @@ print(json.dumps({
     "major": sys.version_info.major,
     "minor": sys.version_info.minor,
     "abiflags": sysconfig.get_config_var("ABIFLAGS"),
+    # implementation_lower is only specified by pypy
+    "interpreter": sysconfig.get_config_var("implementation_lower") or "cpython",
```
konstin: I'd prefer `platform.python_implementation()` because that part of the official python docs.

konstin:
Using `EXT_SUFFIX` and `SOABI` unfortunately doesn't work on windows. E.g. for [cpython 3.7](https://github.com/PyO3/pyo3-pack/blob/master/sysconfig/cpython-win-3.7.txt) we have `EXT_SUFFIX = ".pyd"` and no `SOABI`. I'm fine with using `SOABI` for pypy, but we should check that it works on windows and shouldn't break cpython on windows.

ijl:
Ok, regarding EXT_SUFFIX and SOABI, the existing logic is kept for non-pypy builds. I can't test pypy on Windows or mess with AppVeyor right now, though.

src/python_interpreter.rs
```diff
@@ -360,68 +385,71 @@ impl PythonInterpreter {
...
+                format!(
+                    "pp3{abi_tag}-pypy3_{abi_tag}-{platform}",
+                    abi_tag = self.abi_tag,
+                    platform = platform,
+                )
+            }
         }
     }
 
-    /// Generates the correct suffix for shared libraries
```
konstin: Please re-add that doc comment. Probably it's the best to move the cpython logic to its own function and re-add the comment there

konstin:
I've uploaded a sysconfig for windows, but unfortunately I don't have any reference on how a native extension for pypy should look like on windows. I'm however fine to merge this if we document pypy support as linux only and make a separate issue for windows/mac testing

src/python_interpreter.rs
ijl: Ok, I readded it and some on pypy here.

konstin:
Building on your work, I've created a `pypy-ux` branch where I've fixed running the tests on my machine and adapted the user facing messages and docs a bit. I've also written #105 as some sort of personal wishlist for the pypy support. 

I don't know what's the best way to proceed with this pull request; Do you want to rebase onto master and cherry-pick my commit or should I open another pull request? I'm fine with merging `pypy-ux` as it is now.

ijl:
Go ahead with pypy-ux.

PR #130, opened 2019-05-11
Add a flag to pass an already compiled library

cchalmers:
For the build system I'm using (Nix), using Cargo directly is annoying because I need to build everything from scratch each time. This adds a flag so that you can point to an already compiled library to make the wheel.

### DIFF
--- a/src/build_context.rs
+++ b/src/build_context.rs
@@ -69,6 +69,8 @@ pub struct BuildContext {
     pub interpreter: Vec<PythonInterpreter>,
     /// Cargo.toml as resolved by [cargo_metadata]
     pub cargo_metadata: Metadata,
+    /// The path to an already compiled binding
+    pub artifact: Option<PathBuf>,
 }
 
 impl BuildContext {
@@ -100,7 +102,10 @@ impl BuildContext {
     ) -> Result<Vec<(PathBuf, String, Option<PythonInterpreter>)>, Error> {
         let mut wheels = Vec::new();
         for python_version in &self.interpreter {
-            let artifact = self.compile_cdylib(Some(&python_version))?;
+            let artifact = match self.artifact.clone() {
+                Some(artifact) => artifact,
+                None => self.compile_cdylib(None)?,
+            };
 
             let tag = python_version.get_tag(&self.manylinux);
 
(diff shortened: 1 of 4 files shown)


konstin:
Could you say a bit more about why Nix forces you to build from scratch each time? I'd prefer getting the cargo cache to work on NixOS (e.g. by setting `CARGO_TARGET_DIR`) to circumventing cargo if that's possible.

cchalmers:
I'm no expert on the matter but I'll try to explain how I think it works.

There's two main ways to install rust packages on nix:

There's [`rustPlatform.buildRustPackage`](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/rust/default.nix) which uses the `Cargo.lock` file to prefetch all the dependencies, makes a cargo registry with it and then runs cargo in a sandboxed environment. (At least I think this is what happens, I haven't found much [documentation on it](https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/rust.section.md)). I don't know an easy that you could use a cache with this method.

The other way is [`carnix`](https://nest.pijul.com/pmeunier/carnix) which generates a `Cargo.nix` from a `Cargo.toml` and `Cargo.lock` file. This `Cargo.nix` generates a separate derivation for each dependency so you only have to rebuild your package when you make a change. This method doesn't use `cargo` at all when building, everything's build directly with `rustc`. The result of this build is a `.so` file that I can hand to `pyo3-pack` (with this PR). It might be possible to reuse the dependency derivations and run `cargo` at the last stage here but I don't know how.

konstin:
If carnix forces you to recompile on every change, how does does the option to pack an arbitrary library into a wheel solve that problem?

While allowing to package arbitrary libraries is easy, it also makes it really easy to shoot yourself in the foot, e.g. you could package the libraries for the wrong python version, for different bindings or of a different version of the library.

cchalmers:
Sorry, I wasn't very clear. You only have to recompile *your* package when you make a change with carnix, all of your dependencies are precompiled and stored in the nix store. Using `buildRustPackage` without carnix, every change you make to your package would mean starting from scratch and recompiling all of the dependencies too (for instance `pyo3-pack` on nix uses this method and takes over 4 minutes to compile on my machine).

I appreciate it's easy to make mistakes using this method, I should have probably added a warning message to the help for the `--artifact-path` flag. Are there checks that could be added to detect these differences (something similar to #129)? I also considered adding a `--dry-run` or something that prints out the `rustc` flags, environment variables and the target wheel name which should be useful for testing/scripting. Would this be more acceptable?

PR #143, opened 2019-06-11
Fully remove python 2 support to fix #116

### LINKED ISSUE #116: Fully remove python2 support
Python 2 support is removed from the docs and autodiscovery, but it's actually still supported.

 - [ ] https://github.com/PyO3/pyo3-pack/blob/d0d9040846283d162b178f90fdce6025b4119aef/src/python_interpreter.rs#L44-L47
 - [ ] https://github.com/PyO3/pyo3-pack/blob/8924a3ce4f116a82f19aa1e02d5dba242175ff84/src/compile.rs#L137-L152
 - [ ] The conda test
 - [ ] Other code and docs referencing 2.7

konstin:


lilianmoraru:
Could you please make a new release with these changes?
Since PyO3 does not support Python2, currently `pyo3-pack` is basically broken when it finds Python2 unless you explicitly pass the interpreter.
Also, I see that incremental compilation now works on master and this was a big pain point(I have many dependencies) + I see that I don't need to pass `--manylinux off` any more, it just works.

This would be a great release.

konstin:
Sure. I've pushed the tag and published 0.6.0 on crates.io, so ci should publish the binaries to github and pypi soon.

lilianmoraru:
@konstin Thank you!!
The experience with this release is very nice :)

PR #152, opened 2019-06-26
Error messages for bad cargo/rust environment during PEP517 install

ijl:
The error output from pip is otherwise not understandable by
a user who does not know this package needs Rust or what Rust is.

  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... error
    ERROR: Complete output from command ... pip/_vendor/pep517/_in_process.py prepare_metadata_for_build_wheel ...
    ERROR: 💥 pyo3-pack failed
    Cargo metadata failed: No such file or directory (os error 2)
    ...
    subprocess.CalledProcessError: Command '['pyo3-pack', 'pep517', 'write-dist-info', ... ' returned non-zero exit status 1

#2

### DIFF
--- a/pyo3_pack/__init__.py
+++ b/pyo3_pack/__init__.py
@@ -100,6 +100,16 @@ def get_requires_for_build_sdist(config_settings=None):
 
 # noinspection PyUnusedLocal
 def prepare_metadata_for_build_wheel(metadata_directory, config_settings=None):
+    print("Checking for Rust toolchain....")
+    output = subprocess.check_output(["cargo", "--version"]).decode("utf-8", "ignore")
+    if not "cargo" in output:
+        sys.stderr.write(
+            "cargo, the Rust language build tool, is not installed or is not on PATH.\n"
+            "This package requires Rust to compile extensions. Install it through\n"
+            "the system's package manager or via https://rustup.rs/\n"
+        )
+        sys.exit(1)
+
     command = [
         "pyo3-pack",
         "pep517",


ijl:
I don't know whether all bindings require nightly. If so that is something to check as well.

konstin:
Great idea, thank you!

> I don't know whether all bindings require nightly. If so that is something to check as well.

Only pyo3 requires nightly. pyo3 already prints an understandable error message, even though it might be better to check this earlier:

```
process didn't exit successfully: `/home/konsti/pyo3/target/debug/build/pyo3-46d57cafcc76894a/build-script-build` (exit code: 101)
--- stderr
Error: pyo3 requires a nightly or dev version of Rust.
Installed version is: 1.35.0 (2019-05-20). Minimum required: 1.34.0-nightly (2019-02-06).
thread 'main' panicked at 'Aborting compilation due to incompatible compiler.', build.rs:556:17
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

pyo3_pack/__init__.py
```diff
@@ -100,6 +100,16 @@ def get_requires_for_build_sdist(config_settings=None):
 
 # noinspection PyUnusedLocal
 def prepare_metadata_for_build_wheel(metadata_directory, config_settings=None):
+    print("Checking for Rust toolchain....")
+    output = subprocess.check_output(["cargo", "--version"])
+    if not "cargo" in output:
```
konstin: Running this locally in a venv with python 3.6 on linux I get a `TypeError: a bytes-like object is required, not 'str'`

ijl: Oh, yes, it's back to `.decode("utf-8", "ignore")`.

PR #170, opened 2019-07-23
Use existing cbindgen.toml files as base configuration.

0ndorio:
If a custom `cbindgen.toml` file exist use this one as base configuration and just overwrite all settings which are either required or known to be broken with the current version of cffi.

This should help to avoid the requirement for a `target/header.h` which is only required by pyo3-pack file and still allow to introduce custom configuration (e.g. a list of additional crates to include).

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -38,7 +38,9 @@ pyo3 will set the used python interpreter in the environment variable `PYTHON_SY
 
  Cffi wheels are compatible with all python versions, but they need to have `cffi` installed for the python used for building (`pip install cffi`).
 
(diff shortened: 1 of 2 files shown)


konstin:
Thank you for your pull request, this is a great idea!

Could you please add sentence to the readme explaining that pyo3-pack will pick a cbindgen.toml in the project root?

0ndorio:
> Could you please add sentence to the readme explaining that pyo3-pack will pick a cbindgen.toml in the project root?

I amended an update for the readme. Would you please check if the new description fits your expectations?

0ndorio:
Seems like the build failed due to some network problems.

konstin:
Thank you!

PR #172, opened 2019-07-31
Cookiecutter

pattonw:
Add a reference to a mixed layout python/rust package cookiecutter.

Very much still a work in progress.

`pip install .` and deploying to pypi on tagged commit through travis_ci should work out of the box.

Most other features have not been tested.

Contributions welcome

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -98,6 +98,8 @@ my-project
     └── lib.rs
 ```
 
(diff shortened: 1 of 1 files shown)


konstin:
I think that an example/template for the mixed layout is definitely a good idea, however I'm not sure if cookiecutter-pypackage is right base for that. Cookiecutter is centered around setup.py, while pyo3-pack is based on the new PEP 517 with its `pyproject.toml`. Also note that idiomatic testing with tox is currently blocked on a tox bug (https://github.com/tox-dev/tox/issues/1344).

pattonw:
My fork of cookiecutter-pypackage has removed setup.py and tox. It just has a make pattern for testing which uses pyo3-pack develop and pytest. Travis handles making virtual environments for testing different versions.

PR #177, opened 2019-08-12
linux, darwin compatibility for python3.8

ijl:
From https://docs.python.org/3.8/whatsnew/3.8.html:

"Default sys.abiflags became an empty string: the m flag
for pymalloc became useless (builds with and without
pymalloc are ABI compatible) and so has been removed.

...

On Linux, for example, the Python 3.7 suffix
.cpython-37m-x86_64-linux-gnu.so became
.cpython-38-x86_64-linux-gnu.so in Python 3.8."

### DIFF
--- a/src/python_interpreter.rs
+++ b/src/python_interpreter.rs
@@ -327,7 +327,10 @@ fn fun_with_abiflags(
         );
     }
 
-    if target.is_windows() {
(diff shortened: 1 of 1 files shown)


konstin:
Thank you!

PR #184, opened 2019-08-19
Replace hyper with ureq

konstin:


konstin:
The only reason for doing this is getting the dependency count from 267 to 206.

konstin:
Cancelled due to https://github.com/abonander/multipart/issues/122

PR #185, opened 2019-08-22
Switch to curl

konstin:


konstin:
Blocked by https://github.com/alexcrichton/curl-rust/issues/282 and https://github.com/alexcrichton/curl-rust/issues/286

PR #189, opened 2019-08-29
Add support to detect conda virtual environments

ahmed-masud:
This patch allows `pyo3-pack develop` to be
executed under conda environments.


konstin:
I've bumped the nightly used for building in #190. `Option::xor` should now work without any feature flags

PR #203, opened 2019-09-06
Check interpreter platform instead of target for windows abiflags

### LINKED ISSUE #200: platform mismatch for cross compilation
When compiling a cffi crate for a different target through a Cargo environment which has been setup for cross-compilation, maturin fails on a platform check.
I'm trying to build a windows cffi package from a linux build system, but maturin tries to get the windows python.exe, which can easily be overruled by the `-i` flag. But when passed a linux python version, maturin crashes with this output:

```sh
$ maturin build -b cffi --target x86_64-pc-windows-gnu -i python3.7

💥 maturin failed
  Caused by: Failed to find python interpreter for generating cffi bindings
  Caused by: Failed to get information from the python interpreter at python3.7
  Caused by: sys.platform in python, linux, and the rust target, Target { os: Windows, is_64_bit: true }, don't match ಠ_ಠ
```
I can manually replace the native.so file within the built wheel by a built windows dll file and it works fine on windows. So I'm guessing this is just a limitation within maturin currently.
This check doesn't seem necessary for me, but could probably be wrong for other use-cases. Does it seem feasible to change this behavior?

JasperDeSutter:
fixes #200 
This does the trick for cross-compiling windows targets under linux.
I don't really know the implications of this change, but the tests aren't failing because of it, giving me the impression that it works.

### DIFF
--- a/src/python_interpreter.rs
+++ b/src/python_interpreter.rs
@@ -334,7 +334,7 @@ fn fun_with_abiflags(
     if message.interpreter == "pypy" {
         // pypy does not specify abi flags
         Ok("".to_string())
-    } else if target.is_windows() {
+    } else if message.platform == "windows" {
         if message.abiflags.is_some() {
             bail!("A python 3 interpreter on windows does not define abiflags in its sysconfig ಠ_ಠ")
         } else {


konstin:
Thank you!

PR #225, opened 2019-10-23
Specify python_requires >= 3.5

ijl:
This is needed to prevent pip to try to install it on unsupported versions. https://github.com/ijl/orjson/issues/35


konstin:
Thank you

PR #239, opened 2019-11-19
Try caching layers to speed up the docker build

konstin:


konstin:
It's only a minute less, but at least that.

PR #248, opened 2019-12-04
Add support to detect conda virtual environments

robertodr:
Replays commit by @ahmed-masud to let `maturin` work with Conda environments as well.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -441,7 +441,7 @@ fn run() -> Result<(), Error> {
             release,
             strip,
         } => {
-            let venv_dir = match env::var_os("VIRTUAL_ENV") {
+            let venv_dir = match env::var_os("VIRTUAL_ENV").xor(env::var_os("CONDA_PREFIX")) {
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -441,7 +441,7 @@ fn run() -> Result<(), Error> {
             release,
             strip,
         } => {
-            let venv_dir = match env::var_os("VIRTUAL_ENV") {
+            let venv_dir = match env::var_os("VIRTUAL_ENV").xor(env::var_os("CONDA_PREFIX")) {
                 Some(dir) => PathBuf::from(dir),
                 None => {
                     bail!("You need be inside a virtualenv to use develop (VIRTUAL_ENV isn't set)")
```
konstin: Please change this error message to also include conda.

robertodr: Thanks for pointing this out. Comment addressed.

konstin:
Thank you!

PR #249, opened 2019-12-16
fix(python-interpreter) Fix an error message

Hywan:
This patch patch fixes fixes an error error :-).

Thanks for the awesome project!


konstin:
Thank you!

PR #250, opened 2019-12-26
32 bit Python modules use i386, not x86

dae:
The generated .so files on a 32 bit machine weren't being recognised by Python. When renaming to 'i386' instead of 'x86', Python is able to pick them up. Tested on a 32 bit Ubuntu 16.04 machine using Python 3.5.


konstin:
Thank you!

PR #254, opened 2020-01-02
Fix non-interpreter-specific wheels using tag with py2

ijl:
For example the issue is the py2 tag in maturin-0.7.7-py2.py3-none-macosx_10_7_x86_64.whl.


konstin:
Thank you!

PR #271, opened 2020-02-26
Filter Cargo.toml.orig from sdist

ijl:
This path began appearing in the output of
cargo package --list --allow-dirty --manifest-path=Cargo.toml around
nightly-2020-02-22. It doesn't exist prior to the call and isn't created
by it. This causes:

💥 maturin failed
  Caused by: Failed to add file from Cargo.toml.orig to sdist as Cargo.toml.orig
  Caused by: No such file or directory (os error 2) when getting metadata for Cargo.toml.orig

### DIFF
--- a/src/source_distribution.rs
+++ b/src/source_distribution.rs
@@ -69,6 +69,7 @@ pub fn source_distribution(
             let relative_to_cwd = manifest_dir.join(relative_to_manifests);
             (relative_to_manifests.to_path_buf(), relative_to_cwd)
         })
+        .filter(|(target, _)| target != Path::new("Cargo.toml.orig"))
         .collect();
 
     if !target_source


konstin:
Imho this is a cargo bug, caused by https://github.com/rust-lang/cargo/pull/7905. I've filed this as a bug in https://github.com/rust-lang/cargo/issues/7938. 

Thanks for reporting this!

konstin:
(I'm going to leave this open until there is a decision on the bug report. In case they argue that it is correct to include Cargo.toml.orig, I'm going to merge this)

ijl:
It's a fatal error to maturin but not a practical issue for cargo, and there are rust releases with this behavior even if it's fixed later, so it may worth just working around anyway.

oconnor663:
The instructions int the README no longer work due to this issue, for the release version of maturin. When I added `pyproject.toml` to my project, I started seeing this error from several commands. If it's not appropriate to fix this in maturin itself, what workaround would you suggest for users? I'd be happy to submit a PR documenting this workaround in the README.

oconnor663:
For some unfortunate reason, Google does not find this issue when I search "Failed to add file from Cargo.toml.orig to sdist as Cargo.toml.orig". It took me quite some time to get here, and I worry others might be similarly stuck.

konstin:
So it seems we actually need to filter ourselves. 

Thanks again for the pull request!

Issue #237, opened 2019-11-11
ARM Support


tcirstea:
Any plans for supporting ARM? 

On a side note, building maturin works, but then fails to be used since this happens:
```
💥 maturin failed
Caused by: The architecture ARM is not supported
```

konstin:
Do you have any documentation or reference wheel for native wheels on arm? From reading the distutils source code the architecture is `os.uname().machine`, but couldn't find a list of possible values or notes about python support of those architectures. You should also note that everything but x86 also isn't manylinux compatible, so it can't be published as wheel on pypi.

programmerjake:
If you want, I can build some wheels on my armv7a computer.

konstin:
A wheel as reference and a sysconfig (`python3 -m sysconfig`) would definitely be helpful!

programmerjake:
see also: https://github.com/pypa/manylinux/issues/84

programmerjake:
I created a wheel for gmpy2 since that doesn't take long.
[python3.5-wheel-ubuntu-16.04-on-odroid-xu4.tar.gz](https://github.com/PyO3/maturin/files/3837062/python3.5-wheel-ubuntu-16.04-on-odroid-xu4.tar.gz)

konstin:
Thanks for those files! So the apparent differences are:
 * `linux_armv7l` instead of `linux_x86_64` for the wheel name
 * `arm-linux-gnueabihf` instead of `x86_64-linux-gnu` for the so name

Both travis ci and github actions support arm, so we could use either for continuous integration.

programmerjake:
apparently cpython gets the platform name on non-android unix systems from https://github.com/python/cpython/blob/5c0c325453a175350e3c18ebb10cc10c37f9595c/configure.ac#L708-L849
pip gets wheel platform names from https://github.com/pypa/pip/blob/4447cb0fae4b38c12bb1cec2d59bb2a230638145/src/pip/_internal/pep425tags.py#L140-L164

see also [PEP 425](https://www.python.org/dev/peps/pep-0425/) and [PEP 427](https://www.python.org/dev/peps/pep-0427/)

konstin:
I've tinkered a bit with this and while it's not that difficult to implement, the current `Target` struct and `OS` enum should be replaced by `Platform` enum to avoid this becoming to messy:

```rust
pub enum Target {
    Linux { manylinux: Manylinux, arch: Arch},
    Windows {is_64_bit: bool},
    Macos,
[...]

PR #273, opened 2020-02-28
Support arm7l, aarch64 architectures and manylinux2014 tag

ijl:
The architecture, wheel names, and shared object names
match what is specified in CPython and pip.

arm7l and aarch64 are supported for Linux only.

Target::from_target_triple() returns an Err on an
unsupported OS and architecture combination.

This allows building arm7l and aarch64 on manylinux1 and
manylinux2010, which should not work in pip.
Target::from_target_triple() could be expanded to error
on this but that needs more refactoring than seems
appropriate here.

#237


konstin:
Thank you!

Issue #291, opened 2020-04-05
Support for cargo workspaces


g-braeunlich:
In `test-crates` there is an example `lib_with_path_dep` referencing `some_path_dep`. However this example does not provide a `pyproject.toml` and also an own attempt to build such a project with `pip install .` fails for me.
My idea here would be to add support for cargo workspaces:
https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html
Basically this would mean: you have a top level project dir with a cargo.toml only specifying its members in the above example `lib_with_path_dep` and `some_path_dep`).
I would then put the section `[package.metadata.maturin]` into this file as `[workspace.metadata.maturin]` and also introduce a new field `workspace.metadata.maturin.extension(s)` specifying which of the members is a / the crate containing the pyO3 extension.
For mixed projects, the python code would now go to the toplevel directory together with (pyproject.toml / README.md).
What do you think?

konstin:
FYI, the two crates in `test-crates` exist to test that is shown when using local (including workspace) dependencies as a test for https://github.com/PyO3/maturin/issues/238. Also note that `maturin develop` and `maturin build` work fine, only source distributions are affected.

I think the most user friendly solution would be calling `cargo package --list` on each linked or transitively linked package in the workspace and adding them to the source distribution (This information is part of the `cargo metadata` output). If the root package is not directly in the workspace root than the location of the root package needs to be recorded somewhere, e.g. by adding a metadata key the workspace root `Cargo.toml`.

clbarnes:
Somewhat related: my use case is a cargo workspace including a rust crate (for crates.io) and a pyo3-based wrapper using maturin, which is a mixed project and has a local dependency on the other crate. I'm trying to version them independently, but publishing with maturin produces an sdist which has a remote dependency on the published crate, and a wheel which packages the compiled result of the local dependency: potentially two different versions of the crate.

I think that either way (use the local, or use the published) would be fine (or could be worked around) so long as sdist and bdist were consistent.

tari:
I recently had similar problems. I wanted to build a crate with multiple bindings in a workspace, something like:

* Cargo.toml (workspace)
* base/Cargo.toml (Rust core with external dependencies)
* py/Cargo.toml (Python extension module, depends on `base = { path = "../base" }`)
* wasm/Cargo.toml (WASM bindings, depends on base)

[...]

PR #296, opened 2020-04-27
sdist-include option

ijl:
This allows including arbitrary data in source distributions
that will not appear in cargo list, namely vendored packages.

It feels hacky but I think it's needed for a vendored package or the one Python crate, one Rust crate use case in #291.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -542,6 +542,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "glob"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


konstin:
Thank you for the pull request! Clippy complains about one thing, otherwise the code looks good.

ijl:
@konstin ok fixed and a changelog entry added.

konstin:
Thank you!

ijl:
@konstin if you have time could you please tag a version for this?

konstin:
Sure, I've tagged 0.8.1. Should be available as soon as ci passes.

PR #300, opened 2020-04-30
resulting wheels to work on CentOS 7

AvdN:
The results of building with maturin don't work on CentOS 7, as I noted when installing blake3 (which is build using maturin).
This is because of libc-2.5 in manylinux1 (which is CentOS 5 based). If you change to manylinux2010 (which is CentOS 6 based) you'll get libc-2.16 and the wheels from that do install on CentOS 7.
Those wheels would no longer work on CentOS 5, but a) that is no longer maintained and b) your current wheels don't work on that  anyway. Those wheels should work on CentOS 6, but you would have to compile a Python3 for that platform yourself.

By switching to manylinux2010 as the base, you'll gain support for CentOS 7 and linux version of the same "generation", only losing older platforms where things (i.e. rust) will not work anyway.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/pypa/manylinux1_x86_64
+FROM quay.io/pypa/manylinux2010_x86_64
 
 ENV PATH /root/.cargo/bin:$PATH
 # Add all supported python versions


ijl:
orjson builds using maturin on PyPA's CentOS 5 manylinux1 image without issue. Is the problem perhaps specific to the blake3 implementation requiring glibc>2.5?

konstin:
`manylinux1` built wheels do work on CentOS 7, otherwise none of the manylinux1 native wheels would would work on CentOS 7, including numpy and pandas. The problem with blake3 is rather that it is built on ubuntu-latest instead of manylinux1 and is therefore not manylinux1 complaint.

oconnor663:
Apologies for the confusion here. This is a result of me not knowing what I was doing. I run `maturin build` [here](https://github.com/oconnor663/blake3-py/blob/0.1.4/.github/workflows/maturin_build_wheel.py#L12-L13) (via [here](https://github.com/oconnor663/blake3-py/blob/0.1.4/.github/workflows/tag.yml#L36-L38)). That CI machine is ubuntu-latest, so that's the environment it's running in. Reading the docs more closely now, it looks like I should run this instead:

    docker run --rm -v $(pwd):/io konstin2/maturin:master build

I'll try it and see how it goes :)

PR #320, opened 2020-06-04
Fix typo in publish --no-strip description

hombit:


konstin:
Oops, I missed that when inverting the option.

Thanks!

PR #336, opened 2020-07-26
Fix `build_sdist` return value for PEP 517

davidhewitt:
According to [PEP 517] the `build_sdist` hook should return just the `basename` of the output archive.

https://www.python.org/dev/peps/pep-0517/#build-sdist


davidhewitt:
@konstin just wanted to check you agree this implementation is sensible. If you are happy with it, I will also update the README as tox support should work with next version of maturin 😄

konstin:
Thanks for pointing this out, I've fixed this on the rust side in 7808bbf796f2c557ef047939fad0474f9c43be40.

PR #341, opened 2020-07-29
Fix armv7l wheels tag

omerbenamram:
Hi,

It seems that while generating wheels for arm targets, maturin uses the wrong tags.
PEP 0599 defines the list of added tags as:
```
x86_64
i686
aarch64
armv7l
ppc64
ppc64le
s390x
```

and currently the code exports the tag `arm7l`, which pip will refuse to install.
I've verified that with this changeset the wheels produced are working as expected :)


konstin:
Thank you!

Issue #247, opened 2019-12-03
Handle restructured text (rst) readme files


clbarnes:
A lot of python packages use sphinx for documentation, and many of those incorporate the README into the docs page; so many packagers of python projects prefer RST format for the readme.  Indeed, the default readme content type for PyPI is RST. Maturin is currently angled more towards rust packages, and so expects a markdown readme, even though it publishes packages to pypi. As such, the long descriptions on pypi are messed up.

Is there any way around this?

funkyfuture:
Markdown is hard-coded as content-type [here](https://github.com/PyO3/maturin/blob/master/src/metadata.rs#L79-L84). regarding an implementation i wonder whether that field should be mandatory when a description is given, or whether maturin should try to sniff it. i also have doubts whether GFM (or any other) should be the default dialect.

clbarnes:
It would be most familiar if it behaved the same as setuptools (assume rst) but realistically the current behaviour should probably remain the default. Markdown flavour could easily be configurable, again defaulting to current behaviour.

clbarnes:
Closed by #360

PR #360, opened 2020-10-07
Configurable Description-Content-Type

clbarnes:
No validation is performed.
The defaults are unchanged (and inconsistent with setuptools/PyPI): that is, None if there is no readme, GFM if there is a readme and no explicit content type.

Basic test for checking metadata gets through.

See #247

### DIFF
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -76,22 +76,25 @@ impl Metadata21 {
             None
         };
 
-        let description_content_type = if description.is_some() {
-            // I'm not hundred percent sure if that's the best preset
-            Some("text/markdown; charset=UTF-8; variant=GFM".to_owned())
-        } else {
-            None
-        };
-
         let classifier = cargo_toml.classifier();
-
-        let extra_metadata = cargo_toml.remaining_core_metadata();
 
         let author_email = if authors.contains('@') {
             Some(authors.clone())
         } else {
             None
         };
+
+        let extra_metadata = cargo_toml.remaining_core_metadata();
+
+        // See https://packaging.python.org/specifications/core-metadata/#description-content-type
+        let description_content_type = extra_metadata.description_content_type.or_else(|| {
+            if description.is_some() {
+                // I'm not hundred percent sure if that's the best preset
+                Some("text/markdown; charset=UTF-8; variant=GFM".to_owned())
+            } else {
+                None
+            }
+        });
 
         Ok(Metadata21 {
             metadata_version: "2.1".to_owned(),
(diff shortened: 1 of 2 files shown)


src/metadata.rs
```diff
@@ -263,6 +257,39 @@ mod test {
...
+
+        let cargo_toml: CargoToml = toml::from_str(&cargo_toml).unwrap();
+
+        let metadata =
+            Metadata21::from_cargo_toml(&cargo_toml, &readme_md.path().parent().unwrap()).unwrap();
+
+        let actual = metadata.to_file_contents();
+
+        if actual.trim() != expected.trim() {
+            panic!(
```
konstin, line 270: You can pass a message and formatter options as third and following parameters to assert_eq ([example](https://doc.rust-lang.org/nightly/std/macro.assert_eq.html#examples))

src/metadata.rs
```diff
@@ -314,13 +335,61 @@ mod test {
...
+
+        let cargo_toml = indoc!(
+            r#"
+            [package]
+            authors = ["konstin <konstin@mailbox.org>"]
+            name = "info-project"
+            version = "0.1.0"
+            description = "A test project"
+            homepage = "https://example.org"
+            readme = "readme.md"
```
konstin, line 357: Shouldn't this be `readme.rst`?

konstin:
Thank you for the pull request! 

I like adding the `description_content_type` option, but could we try to guess it based on the file extension? So set it markdown for `Readme.md` and to rst for `Readme.rst`?

src/metadata.rs
clbarnes: It could be but it doesn't have to be, as this test just needs to prove that the metadata gets through. I'll fix it for clarity, though.

src/metadata.rs
clbarnes: Good point, that was a leftover from a slightly different implementation, I'll change that.

clbarnes:
> could we try to guess it based on the file extension?

Should be easy enough!

src/metadata.rs
clbarnes: On second look, it doesn't actually matter because the substring is replaced by the path of the temporary file created to write the readme contents into.

clbarnes:
Note https://github.com/PyO3/maturin/pull/360/commits/fef38f82fbc4712b2a7a2b29b3536c69729b61a2 - where no content type is given, and the file path does not have a known extension, content type defaults to plaintext rather than GFM. I think this makes most sense and the results shouldn't be disastrous (as markdown should be largely readable in plain text), but it is technically a breaking change.

There's a little hole in the testing here because the existing tests write to and read from a file without an extension, so the end-to-end test for checking extensions is a bit tricky. But the filename-parsing logic is tested.

clbarnes:
Test failure relates to a missing GLIBC version, which I hope I'm not responsible for :grimacing:

konstin:
> Note fef38f8 - where no content type is given, and the file path does not have a known extension, content type defaults to plaintext rather than GFM. I think this makes most sense and the results shouldn't be disastrous (as markdown should be largely readable in plain text), but it is technically a breaking change.

Thanks, that's definitely better than before!

> Test failure relates to a missing GLIBC version, which I hope I'm not responsible for grimacing

Yep, iirc a recent rustc version dropped support for manylinux1 and I have to move it to manylinux2010.

konstin (approved):
Code looks good, thank you for implementing that!

PR #361, opened 2020-10-08
Remove python 3.5 and manylinux1, add python 3.9 and default to manylinux2010

### LINKED ISSUE #358: Add Python 3.9 to maturin Docker image
Python 3.9 [is released](https://docs.python.org/3/whatsnew/3.9.html), please can you add it to the official `maturin` Docker image?

### LINKED ISSUE #359: Deprecate EOL python 3.5

konstin:
Fixes #359 
Fixes #358


programmerjake:
As mentioned in https://github.com/PyO3/maturin/issues/359#issuecomment-705063893 and in https://github.com/PyO3/pyo3/issues/685#issuecomment-568299334

I don't think we should remove Python 3.5 just yet, since it is the default version on Ubuntu 16.04 which is still supported until April 2021.

konstin:
Sorry, but python3.5 is unsupported independent of whatever ubuntu is doing, especially since both numpy and django have dropped it. Pandas has already removed 3.6, and numpy should follow soon. I recommend [pyenv](https://github.com/pyenv/pyenv) for installing and managing python versions on linux.

PR #365, opened 2020-10-22
Allow dot paths (.conda) to be found

tafia:
My conda installs environment in `C:\Users\xxx\.conda\envs` folders. This PR allows searching such paths to find an interpreter.


konstin:
Thank you!

Issue #366, opened 2020-10-22
PowerPC64LE support


programmerjake:
Wheel name `numpy-1.19.2-cp38-cp38-linux_ppc64le.whl`
`.so` name `lapack_lite.cpython-38-powerpc64le-linux-gnu.so`

[python-sysconfig.txt](https://github.com/PyO3/maturin/files/5424532/python-sysconfig.txt)

I have to install maturin using:
```bash
cargo install maturin --no-default-features --features=auditwheel,log,human-panic
```
since ring doesn't support powerpc64le-unknown-linux-gnu: https://github.com/briansmith/ring/issues/389#issuecomment-711801345

mzpqnxow:
@programmerjake I have the same issue and have been watching the AES-GCM issue since April :/

I only have a dependency on `maturin` because of the inclusion of `orjson` in one of my projects. It's not entirely clear to me how to install the `maturin` Python, though I did install `maturin` rust via `cargo`. Can you give me a pointer as to what needs to be done to get the `maturin` Python package to install after this, so I can get past this dependency? Or maybe I'm misunderstanding and you're still blocked as well?

I'm looking at the `setuptools` `setup.py` and I see the invocation of:

```
["cargo", "rustc", "--bin", "maturin", "--", "-C", "link-arg=-s"]
```

Do I need to replace this?

Thanks for any help you can provide, I have nearly zero experience with rust/cargo as you can probably guess...

mzpqnxow:
I tried replacing the `check_call` arguments in `setup.py` with:

```
                ["cargo", "rustc", "--bin", "maturin", "--", "-C", "link-arg=-s", '--no-default-features', '--features=auditwheel,log,human-panic']
```

But no luck, same error. I also pulled the latest stable rustc, same issue:

```
rustc 1.47.0 (18bf6b4f0 2020-10-07)
```

mzpqnxow:
@programmerjake I think I figured it out.. sorry for bugging you ;)

I ended up using your command to install maturin, then commenting out the `check_call` in `setup.py`, then modifying the path in the file copy function to point at the `maturin` that was already installed and that seems to take care of it..

mzpqnxow:
For anyone else having this issue, a more explicit explanation of the solution from @programmerjake 

```
$ cat > maturin-setuptools.patch << 'EOF'
--- ../setup.py	2020-08-17 14:47:52.000000000 +0000
+++ setup.py	2020-10-24 21:25:06.074357597 +0000
[...]

PR #367, opened 2020-10-25
Add ppc64le/POWERPC64 as a "supported" target (issue #366)

mzpqnxow:
The checks here are (largely) synthetic. There is no real architecture specific code or configuration aside from the platform triplet, which this patch covers. This is associated with issue #366 

Without this patch, it is possible to *build* maturin on ppc64le, but it will refuse to run, erroring out with "The architecture POWERPC64 is not supported"  before starting the build

* This works fine to prevent maturin from refusing to try to build a package on ppc64le
* Whether the package you intend to build with maturin will support ppc64le (at build or run time) is another matter
* With modifications, this was used to successfully build and run a functional orjson (rust => native code Python module) on py3.7+ppc64le - binary wheels available [here](https://github.com/mzpqnxow/orjson-ppc64le-wheel)
* Should not have any breaking affect on other platforms

Please let me know if you prefer this PR to be based from and send to a different branch other than master


mzpqnxow:
There are some caveats here- and because of these I would understand if you don't want to accept this:

1. I don't know how your CI pipeline is going to like this (presumably, it will not- it will require a `ppc64le-linux-musl` toolchain at the very least (which I am happy to provide the config.mak for if you *do* want it)
2. Because of the open issue with ring not supporting ppc64le, there are build-time issues that remain even with this patch

The build-time issues can be resolved by avoiding dependence upon ring using the following when building via `setuptools` as noted by @programmerjake in #366

```
--no-default-features --features=auditwheel,log,human-panic
```

This can be wrapped in an architecture check conditional in `setup.py`, e.g:

```
import platform
...
if platform.machine() == 'ppc64le':
   cargo_args = [..., '--no-default-features', '--features=auditwheel,log,human-panic'] 
else:
   cargo_args = ["cargo", "rustc", "--bin", "maturin", "--", "-C", "link-arg=-s"]
...
```

I realize you may not want to support this which is fine, but if you do let me know and I can clean up the changes to setuptools and include them in this PR. If you're not at all interested feel free to close this PR

Thanks!

programmerjake:
The enumerant should probably be named `PowerPC64LE` since there is a separate `powerpc64` platform which is for big-endian 64-bit Power.

mzpqnxow:
> The enumerant should probably be named `PowerPC64LE` since there is a separate `powerpc64` platform which is for big-endian 64-bit Power.

I agree, as it stands now a big-endian PPC64 build would get a `ppc64le-linux-gnu` build target, which is certainly not correct

I originally actually used `POWERPC64LE` but realized that both [core::arch](https://doc.rust-lang.org/core/arch/) and [platforms::target::Arch][https://docs.rs/platforms/0.2.1/platforms/target/enum.Arch.html] don't have any byte-order specific values. At that point I took a shortcut and just assumed everyone would be LSB :)

From what I can tell, in Rust, it's required to check byte-order separately from `Arch`. In `Arch`, there is, e.g. `POWERPC64`, `MIPS`, `MIPS64, etc. but no `POWERPC64LE`, `MIPSEL`, `MIPS64EL`. It seems kind of odd to me that it's done this way since most interfaces in other languages provide you with the endianness as part of the architecture string, all in a single module

To do this properly in `maturin` then, it seems an additional level of logic needs to be added in `target.rs` that explicitly determines endianness either before or after the it determines the `Arch`. It can then (depending on the architecture) append the `le` (for `ppc`) or `el` (for `mips`) if it determines they're in little-endian mode.

From what I can tell, the only direct way to get byte-order directly in Rust is via `cfg`, like this:

```
#[derive(Copy, Clone, Debug)]
enum Endianness {
    MSB, LSB,
}
 
impl Endianness {
    fn target() -> Self {
        #[cfg(target_endian = "big")]
        {
            Endianness::MSB
        }
        #[cfg(not(target_endian = "big"))]
        {
            Endianness::LSB
        }
    }
}
```

The implementation as an `enum` is probably not completely necessary, but `cfg(target_endian)` is the way to go right?

Because MIPS uses `EL`/`el` and PPC uses `LE`/`le`, there will have to be one little piece of logic that can convert `Endianness::LSB` to either `le` or `el`.

Does this sound about right to you? Or is there a more straightforward way that you're aware of to get `ppc64le` at runtime in rust core?

I'll try to get back to this later on this week as I have time to add it and test it and will update the PR.

Thanks, I really appreciate you reviewing this and providing the info about the original problem I was having with the ring dependency. This weekend was the first time I've worked with `cargo`/`rustc` so you saved me a tremendous amount of time

programmerjake:
You should be able to do:
```rust
let arch = match platform.target_arch {
    platforms::target::Arch::X86_64 => Arch::X86_64,
    platforms::target::Arch::X86 => Arch::X86,
    platforms::target::Arch::ARM => Arch::ARMV7L,
    platforms::target::Arch::AARCH64 => Arch::AARCH64,
    platforms::target::Arch::POWERPC64 if platform.target_triple.starts_with("powerpc64-") => Arch::POWERPC64,
    platforms::target::Arch::POWERPC64 if platform.target_triple.starts_with("powerpc64le-") => Arch::POWERPC64LE,
    unsupported => bail!("The architecture {:?} is not supported", unsupported),
};
```

Using `#[cfg(target_endian = ...)]` doesn't work when using maturin to cross-compile since those `cfg` are set based on what arch maturin is compiled for, not the arch maturin is targeting.

konstin:
Thank you for the pull request! I don't have any knowledge about powerpc64, so I'm happy with whatever solution for endianness you decide upon.

konstin:
> This can be wrapped in an architecture check conditional in `setup.py`, e.g:
> 
> ```
> import platform
> ...
> if platform.machine() == 'ppc64le':
>    cargo_args = [..., '--no-default-features', '--features=auditwheel,log,human-panic'] 
> else:
>    cargo_args = ["cargo", "rustc", "--bin", "maturin", "--", "-C", "link-arg=-s"]
> ...
> ```
> 
> I realize you may not want to support this which is fine, but if you do let me know and I can clean up the changes to setuptools and include them in this PR. If you're not at all interested feel free to close this PR

I'm fine with adding that to setup.py, please feel free to add it to the pull request :)

mzpqnxow:
> You should be able to do:
> 
> ```rust
> let arch = match platform.target_arch {
>     platforms::target::Arch::X86_64 => Arch::X86_64,
>     platforms::target::Arch::X86 => Arch::X86,
>     platforms::target::Arch::ARM => Arch::ARMV7L,
>     platforms::target::Arch::AARCH64 => Arch::AARCH64,
>     platforms::target::Arch::POWERPC64 if platform.target_triple.starts_with("powerpc64-") => Arch::POWERPC64,
>     platforms::target::Arch::POWERPC64 if platform.target_triple.starts_with("powerpc64le-") => Arch::POWERPC64LE,
>     unsupported => bail!("The architecture {:?} is not supported", unsupported),
> };
> ```
> 
> Using `#[cfg(target_endian = ...)]` doesn't work when using maturin to cross-compile since those `cfg` are set based on what arch maturin is compiled for, not the arch maturin is targeting.

Ah, thank you for this. I was wondering where `cfg` came from, whether it was based on build-flags or based on the native system, that answers my question

mzpqnxow:
@konstin I just pushed the last few bits into this so it's ready to merge

I took the suggestion from @programmerjake to cover both ppc64 lsb and msb correctly

To test:

- Created a new user on an Ubuntu 16.04.5 POWER8 (ppc64le) system
- Installed a nighttly `rustc`/`cargo` using `rustup.sh`
- Created a plain Python3.7 virtualenv
- Cloned this branch and used `pip install .`

It installed correctly. I was able to build the `orjson` package (with some hacks required for that, unrelated to `maturin`)

I'll clean up the rustfmt CI complaints

programmerjake:
> EDIT: Actually, I see the CI failure is: `error: the 'rustfmt' component which provides the command 'cargo-fmt' is not available for the 'nightly-powerpc64le-unknown-linux-gnu' toolchain`. I'm not sure I'll be able to resolve that

I don't see that in the CI. I do know that rustfmt didn't build in the last few nightly versions, so you should use a stable version of rustfmt.

What I see is:
```
Diff in /home/runner/work/maturin/maturin/src/target.rs at line 109:
             platforms::target::Arch::X86 => Arch::X86,
             platforms::target::Arch::ARM => Arch::ARMV7L,
             platforms::target::Arch::AARCH64 => Arch::AARCH64,
-            platforms::target::Arch::POWERPC64 if platform.target_triple.starts_with("powerpc64-") => Arch::POWERPC64,
-            platforms::target::Arch::POWERPC64 if platform.target_triple.starts_with("powerpc64le-") => Arch::POWERPC64LE,
+            platforms::target::Arch::POWERPC64
+                if platform.target_triple.starts_with("powerpc64-") =>
+            {
+                Arch::POWERPC64
+            }
+            platforms::target::Arch::POWERPC64
+                if platform.target_triple.starts_with("powerpc64le-") =>
+            {
+                Arch::POWERPC64LE
+            }
             unsupported => bail!("The architecture {:?} is not supported", unsupported),
         };
-        
+
         // bail on any unsupported targets
```

Basically you can either apply the above diff manually or run:
```bash
rustup install stable
rustup component add rustfmt --toolchain stable
cargo +stable fmt
```

That should correctly format the code.

mzpqnxow:
There's one outstanding issue that I neglected to mention because I hacked around it and forgot- it has to do with the way `pep517 build-wheel` names the shared library in the wheel.

By default, `orjson` comes out as follows on my machine:

`orjson.cpython-37m-ppc64-linux-gnu.so`

After installing the wheel (or installing directly via `pip`) I'm not able to import the module:

```
>>> import orjson
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'orjson'
>>> 
```

To make it work I'm currently building a wheel, unzipping it, renaming the shared library to `orjson.so`, updating the name in the wheel metadata, and then zipping it back up again and installing it with `pip`. It then imports fine

I read your blog post about how inconsistent and poorly documented the "standard" for wheel shared library naming is in Python (and how it's different with different major/minor version of Python) so I understand that this can get pretty weird- I expect that the architecture may be throwing another wrench into this

This isn't a blocker since it has an easy workaround, but you may want to look into it. I'm happy to do any testing you need for it to produce the correct name

mzpqnxow:
Thanks @programmerjake, not sure how I misread that. I fixed the formatting and whitespace and the rust fmt check is now green. 

Looks like there's some spurious failure on the AppVeyer job, not related to the code- seems like a network failure on the backend. I won't worry about that for now

programmerjake:
> There's one outstanding issue that I neglected to mention because I hacked around it and forgot- it has to do with the way `pep517 build-wheel` names the shared library in the wheel.
> 
> By default, `orjson` comes out as follows on my machine:
> 
> `orjson.cpython-37m-ppc64-linux-gnu.so`

As mentioned in https://github.com/PyO3/maturin/issues/366#issue-727585760 on the server I'm testing on the name follows this pattern (picked random `.so` out of numpy): `lapack_lite.cpython-38-powerpc64le-linux-gnu.so`

I'd guess that here:
https://github.com/PyO3/maturin/blob/9e3c5b6aadb2a98756540d600240071d8a625730/src/target.rs#L210-L215
The PowerPC lines should probably instead be:
```rust
 (OS::Linux, Arch::POWERPC64) => "powerpc64-linux-gnu", 
 (OS::Linux, Arch::POWERPC64LE) => "powerpc64le-linux-gnu", 
```

mzpqnxow:
> > There's one outstanding issue that I neglected to mention because I hacked around it and forgot- it has to do with the way `pep517 build-wheel` names the shared library in the wheel.
> > By default, `orjson` comes out as follows on my machine:
> > `orjson.cpython-37m-ppc64-linux-gnu.so`
> 
> As mentioned in [#366 (comment)](https://github.com/PyO3/maturin/issues/366#issue-727585760) on the server I'm testing on the name follows this pattern (picked random `.so` out of numpy): `lapack_lite.cpython-38-powerpc64le-linux-gnu.so`
> 
> I'd guess that here:
> https://github.com/PyO3/maturin/blob/9e3c5b6aadb2a98756540d600240071d8a625730/src/target.rs#L210-L215
> 
> 
> The PowerPC lines should probably instead be:
> ```rust
>  (OS::Linux, Arch::POWERPC64) => "powerpc64-linux-gnu", 
>  (OS::Linux, Arch::POWERPC64LE) => "powerpc64le-linux-gnu", 
> ```

That looks like it works as the naming convention, thanks.

However ... it appears that in `target.rs`, the match is happening on `POWERPC64` as opposed to `POWERPC64LE`, so the triplet is coming out as `powerpc64-linux-gnu`, so the import is still failing when I build from this modified logic- which makes sense because before I started this patch, I received a run-time error of `The architecture POWERPC64 is not supported`- as opposed to `... POWERPC64LE is not supported`

After a quick look at https://docs.rs/platforms/0.2.0/platforms/target/enum.Arch.html, it appears that `platforms::target::Arch` does not consider byte-order and does not have a `POWERPC64LE`, so I think there will need to be an extra piece of logic to check the endianness before assigned the triplet. I can probably adapt using endianness stub I pasted towards the top of this discussion though I would happily take your version of that logic if you want to paste one in here, seeing as I'm just winging it when I write rust code :)

programmerjake:
I opened a pull request to `platforms` to fix deducing the default target: https://github.com/RustSec/platforms-crate/pull/29

I tested that this fixes the issue with maturin picking `powerpc64-unknown-linux-gnu` instead of `powerpc64le-unknown-linux-gnu`.

I had to also switch the required version for cbindgen from `0.14.6` to `0.14.3` since those versions were yanked as mentioned in https://github.com/eqrion/cbindgen/issues/594

programmerjake:
The cbindgen dependency was updated on master, so you can use that commit to fix the yanked crate issue.

programmerjake:
[platforms v1.0.3](https://crates.io/crates/platforms/1.0.3) was released, it includes the necessary fixes.

programmerjake:
I got it to work successfully: https://github.com/mzpqnxow/maturin/pull/1

I updated platforms to v1.0.3, fixed the broken code in setup.py (it would use any old version of maturin instead of the version from the source directory), and fixed manylinux auditing to correctly ignore the dynamic linker/loader on powerpc64le where it is named differently.

konstin:
Thanks to you both, this is excellent!

mzpqnxow:
@konstin, @programmerjake thank you for your effort on this- I was expecting resistance and a general lack of interest, not the active help you both provided in getting to a resolution!

I'm used to "Sorry, ppc64le is not supported and will not be any time soon" :)

programmerjake:
I'm currently helping build a Libre-licensed PowerPC64 CPU/GPU as part of [Libre-SOC](https://libre-soc.org/), how could I not help? 😄

PR #373, opened 2020-11-01
support package name override through metadata

houqp:
the python package name I would like to use conflicts with an existing rust crate name, so it would be useful to support name override to workaround the restriction.


konstin:
Thank you!

I'm amazed it's only such a small change in the code.

houqp:
well, kudos to you for laying out a good code structure :)

PR #374, opened 2020-11-01
add support for loading pypi cred from token and pypirc

houqp:
Is pypi token is specified through `MATURIN_PYPI_TOKEN` env var, then it will be used. Next it will look for credentials stored in `~/.pypirc` before prompting for username and password.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -7,6 +7,8 @@ use anyhow::{bail, Context, Result};
 #[cfg(feature = "upload")]
 use bytesize::ByteSize;
 use cargo_metadata::MetadataCommand;
+#[cfg(feature = "upload")]
+use configparser::ini::Ini;
 #[cfg(feature = "human-panic")]
 use human_panic::setup_panic;
 #[cfg(feature = "password-storage")]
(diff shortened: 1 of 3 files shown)


konstin:
Thank you!

src/main.rs
```diff
@@ -70,14 +72,54 @@ fn get_username() -> String {
...
+    if let Some(mut config_path) = dirs::home_dir() {
+        config_path.push(".pypirc");
+        if let Ok(pypirc) = fs::read_to_string(config_path.as_path()) {
+            let mut config = Ini::new();
+            if config.read(pypirc).is_err() {
+                return None;
+            }
+
+            if let (Some(username), Some(password)) = (
+                config.get(package_name, "username"),
```
konstin, line 85: @houqp Sorry for asking so late, but shouldn't this be the registry name rather than the package name?

houqp: this is because pypi tokens can be scoped to packages. I was expecting users to set token config per package. perhaps we should fallback to the registry when per package config is not found?

konstin: Do you have an example file? I've read https://packaging.python.org/specifications/pypirc/ and I only see how you can set username/password per repository, but I haven't found a way to scope this per package.

houqp: Ha, looking at the doc again, maybe i misused the config section name. They way i have it configured locally is to set the section name as package name:

```
[package-name]
repository = https://upload.pypi.org/legacy/
username = __token__
password = API_TOKEN 
```

houqp: i think the correct behavior here should be just picking the first repo from `index-servers` key instead of using package name?

konstin: Hmm, I understood the header in square brackets as the name of the repository rather than name of the package, e.g. you could do `twine upload -r first-repository myproject/dist/*` and `twine upload -r first-repository myproject/dist/*`, given the example from the docs:

```ini
[distutils]
index-servers =
    first-repository
    second-repository

[first-repository]
repository = <first-repository URL>
username = <first-repository username>
password = <first-repository password>

[second-repository]
repository = <second-repository URL>
username = <second-repository username>
password = <second-repository password>
```

houqp: yeah, based on the doc, section names are supposed to be mapped to repo name. although i think this is a design issue with pypi, which makes it hard to use package scoped tokens for security best practices, it's better to stick to the official spec to avoid surprises for now.

PR #378, opened 2020-11-10
build_wheel: return absolute path from pep517 command

davidhewitt:
I hit an edge case while testing https://github.com/PyO3/pyo3/pull/1269 where a package inside a workspace has the target directory not where maturin expects, so copying the wheel to the final output directory fails.

This hasn't been an issue until now because I am testing with the upcoming pip option `--use-feature=in-tree-build`, which preserves the workspace hierarchy. (In out-of-tree build, once the source is copied then the workspace Cargo.toml will no longer affect the build.)

Solution is easy: just return the absolute path to the built wheel from the `maturin pep517` command.

### DIFF
--- a/maturin/__init__.py
+++ b/maturin/__init__.py
@@ -67,8 +67,9 @@ def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
     sys.stdout.buffer.write(output)
     sys.stdout.flush()
     output = output.decode(errors="replace")
-    filename = output.strip().splitlines()[-1]
-    shutil.copy2("target/wheels/" + filename, os.path.join(wheel_directory, filename))
+    wheel_path = output.strip().splitlines()[-1]
+    filename = os.path.basename(wheel_path)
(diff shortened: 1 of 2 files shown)


konstin:
Thank you, but I fear this violates PEP 517 (https://www.python.org/dev/peps/pep-0517/#build-wheel):

> Must build a .whl file, and place it in the specified wheel_directory. It must return the basename (not the full path) of the .whl file it creates, as a unicode string.

Regarding workspaces, this is a missing feature in maturin. Unfortunately it's complex to implement because it requires changing the sdist folder structure and rewriting paths in Cargo.toml files.

davidhewitt:
Sorry, I explained poorly. The maturin `build_wheel` hook continues to be pep 517 compliant.

However, in its implementation, the `maturingpep517` internal command now returns an absolute path, so that the `build_wheel` hook can copy the wheel to `wheel_directory` and return just the filename.

So I think it's fine?

konstin:
Sorry, I misread your changed!

This indeed a good find, thanks for fixing that!

PR #390, opened 2020-12-20
setup.py: install correctly if CARGO_ env vars set

davidhewitt:
On https://github.com/PyO3/pyo3/pull/1269 I currently have an issue where `CARGO_BUILD_TARGET` environment variable causes `pip install maturin` to fail because the `maturin` executable is put in an unexpected location.

This PR changes setup.py to use the `--message-format=json` output from rustc to snoop the built `maturin` binary location for installation.


konstin:
Thank you!

PR #392, opened 2020-12-22
pep517: print output on build failure

davidhewitt:
While debugging https://github.com/PyO3/pyo3/pull/1269 I learned that if the maturin pep517 build fails, no rust compiler output is shown.

This PR reworks those functions so that build output is printed even if the `maturin pep517` invocation was unsuccessful, which makes debugging errors easier!


konstin:
Thank you!

PR #394, opened 2020-12-26
Complete abi3 support with PyO3 0.13

kngwyu:
Built on #383.
- Update PyO3
- Set `PYO3_NO_PYTHON` if the target is not Windows for speeding up the build
  - I think this is OK but please let me know if you have another idea

BTW, I was planning of adding an option that changes the tag name to `cp3-abi3-...`.
But now I'm sad to know that it actually does not work with pip 😢. See https://github.com/pypa/packaging/issues/261 for more.

### DIFF
--- a/src/build_context.rs
+++ b/src/build_context.rs
@@ -27,8 +27,9 @@ pub enum BridgeModel {
     /// providing crate, e.g. pyo3.
     Bindings(String),
     /// `Bindings`, but specifically for pyo3 with feature flags that allow building a single wheel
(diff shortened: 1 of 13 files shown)


kngwyu:
I specified `abi3-py36` in `pyo3-pure` example since `cp36-abi3` wheel is installable in Python >= 3.6.

konstin:
Thank you for this pull request and all the work that went into the pyo3 part! I've added windows support on top in another branch, so I'm closing this in favor of #399.

PR #396, opened 2020-12-28
Add support for Apple Silicon macOS

LER0ever:
This PR adds support for building and developing PyO3 extensions in macOS on Apple Silicon (aarch64-apple-darwin). 
Before this PR, maturin gives out an error while guessing the current architecture on arm64 mac. 

This PR includes:

- Updates `platforms` crate to 1.1.0 to include support for Apple Silicon macOS: https://github.com/RustSec/platforms-crate/pull/32
- Sets Python Wheel filename with binary format string `macosx_11_0_arm64`, ref: https://github.com/pypa/packaging/pull/319
- Leaves the native extension suffix as it is `darwin`, ref: https://docs.python.org/3/whatsnew/3.5.html#build-and-c-api-changes

This PR does NOT include Universal 2 binary format support, and only produces native arm64 binary wheels on M1 machines.

A screenshot of before and after:

Screen Shot 2020-12-29 at 12 01 35 AM

---

<details>

```
The target whl filename will include binary format string
`macosx_11_0_arm64` as indicated in the CPython packaging PR that added
support for Apple Silicon: https://github.com/pypa/packaging/pull/319

The dot so file will have the same extension suffix as x86_64 darwin,
specified at
https://docs.python.org/3/whatsnew/3.5.html#build-and-c-api-changes

Updated `platforms` crates to 1.1.0 to include the PR for apple silicon
support, so that guessing platform will not throw an unsupported error.
Ref: https://github.com/RustSec/platforms-crate/pull/32

Signed-off-by: LER0ever <i+git@rongyi.io>
```

</details>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1239,9 +1239,9 @@ dependencies = [
 
 [[package]]
 name = "platforms"
-version = "1.0.3"
+version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cc77a3fc329982cbf3ea772aa265b742a550998bad65747c630406ee52dac425"
+checksum = "989d43012e2ca1c4a02507c67282691a0a3207f9dc67cec596b43fe925b3d325"
 
 [[package]]
 name = "podio"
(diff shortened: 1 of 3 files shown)


konstin:
Thank you!

PR #398, opened 2021-01-05
Look for `cargo.exe` on windows.

richardkiss:
This fixes a problem with bootstrapping on Windows.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,23 +44,25 @@ class PostInstallCommand(install):
 
     def run(self):
         source_dir = os.path.dirname(os.path.abspath(__file__))
-        executable_name = "maturin.exe" if sys.platform.startswith("win") else "maturin"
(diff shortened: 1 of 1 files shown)


konstin:
Thank you for the pull request! Could you explain when `cargo` and when `cargo.exe` are in PATH? Could there e.g. be circumstances where `cargo` worked, but `cargo.exe` doesn't?

richardkiss:
I'm not a Windows expert (or even a Windows user). I had to make this change to get `pip install maturin` to work on GitHub ci. I saw a similar branch on `maturin` vs `maturin.exe` and generalized that. Another possibility would be to use `which` on a list of potential names like `["cargo", "cargo.exe"]`, and using the first one found rather than switching on `sys.platform` which feels a bit icky.

PR #403, opened 2021-01-16
Add support for macOS universal2 wheel

messense:
Closes #402

PR #404, opened 2021-01-16
Fix building bin bindings on Apple Silicon

messense:
#403 breaks bin bindings on Apple Silicon Mac machines, found out when trying to use maturin to build [py-spy](https://github.com/benfred/py-spy)


messense:
Side note: should bin bindings allow Python 2? Since it doesn't really depends on Python.

konstin:
Thank you!

> Side note: should bin bindings allow Python 2? Since it doesn't really depends on Python.

No, I don't want to support any python 2 anymore

PR #405, opened 2021-01-20
Replace lipo command line with fat-macho crate

messense:
`lipo` is only available on macOS as a command line tool, the [fat-macho](https://github.com/messense/fat-macho-rs) crate implemented a cross platform Mach-O fat binary writer, makes it easier to do cross compilation.

Based on #404

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,8 +35,10 @@ jobs:
         with:
           profile: minimal
           toolchain: stable
-          target: aarch64-apple-darwin
           override: true
+      - name: Install aarch64-apple-darwin Rust target
+        if: matrix.os == 'macos-latest'
+        run: rustup target add aarch64-apple-darwin
       #- name: Cache cargo registry
       #  uses: actions/cache@v2
       #  with:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,6 +35,7 @@ jobs:
         with:
           profile: minimal
           toolchain: stable
+          target: aarch64-apple-darwin
           override: true
       #- name: Cache cargo registry
       #  uses: actions/cache@v2
@@ -51,6 +52,16 @@ jobs:
       #  with:
       #    path: target
       #    key: ${{ runner.os }}-${{ steps.rustup.outputs.rustc_hash }}-target-${{ hashFiles('**/Cargo.lock') }}
+      - name: Setup Xcode env
+        if: matrix.os == 'macos-latest'
+        shell: bash
+        run: |
+          echo "PYO3_CROSS_LIB_DIR=/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/lib" >> "${GITHUB_ENV}"
+          sudo xcode-select -s /Applications/Xcode_12.3.app
+          bindir="$(xcode-select --print-path)/Toolchains/XcodeDefault.xctoolchain/usr/bin"
+          echo "CC=${bindir}/clang" >> "${GITHUB_ENV}"
+          echo "CXX=${bindir}/clang++" >> "${GITHUB_ENV}"
+          echo "SDKROOT=$(xcrun --sdk macosx --show-sdk-path)" >> "${GITHUB_ENV}"
       - name: cargo test
         uses: actions-rs/cargo@v1
         with:
--- a/src/build_context.rs
+++ b/src/build_context.rs
@@ -118,6 +118,8 @@ pub struct BuildContext {
     pub interpreter: Vec<PythonInterpreter>,
     /// Cargo.toml as resolved by [cargo_metadata]
     pub cargo_metadata: Metadata,
+    /// Whether to use universal2 or use the native macOS tag (off)
+    pub universal2: bool,
 }
 
 type BuiltWheelMetadata = (PathBuf, String);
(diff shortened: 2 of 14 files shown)


.github/workflows/test.yml
```diff
@@ -51,6 +52,16 @@ jobs:
...
+      - name: Setup Xcode env
+        if: matrix.os == 'macos-latest'
+        shell: bash
+        run: |
+          echo "PYO3_CROSS_LIB_DIR=/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/lib" >> "${GITHUB_ENV}"
+          sudo xcode-select -s /Applications/Xcode_12.3.app
+          bindir="$(xcode-select --print-path)/Toolchains/XcodeDefault.xctoolchain/usr/bin"
+          echo "CC=${bindir}/clang" >> "${GITHUB_ENV}"
+          echo "CXX=${bindir}/clang++" >> "${GITHUB_ENV}"
+          echo "SDKROOT=$(xcrun --sdk macosx --show-sdk-path)" >> "${GITHUB_ENV}"
```
messense, line 64: Xcode 12.2 and later can compile to both x86_64 and arm64 targets on Intel & Apple Silicon platforms. I am not sure why `xcode-select` alone does not work, had to setup `CC`, `CXX` etc.

src/build_context.rs
```diff
@@ -118,6 +118,8 @@ pub struct BuildContext {
     pub interpreter: Vec<PythonInterpreter>,
     /// Cargo.toml as resolved by [cargo_metadata]
     pub cargo_metadata: Metadata,
+    /// Whether to use universal2 or use the native macOS tag (off)
+    pub universal2: bool,
```
messense, line 122: If we defaults to set universal2 to `true`, I think this needs to be changed to an off switch just like `skip_auditwheel`

.github/workflows/test.yml
```diff
@@ -51,6 +52,16 @@ jobs:
       #  with:
       #    path: target
       #    key: ${{ runner.os }}-${{ steps.rustup.outputs.rustc_hash }}-target-${{ hashFiles('**/Cargo.lock') }}
+      - name: Setup Xcode env
+        if: matrix.os == 'macos-latest'
+        shell: bash
+        run: |
+          echo "PYO3_CROSS_LIB_DIR=/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/lib" >> "${GITHUB_ENV}"
```
messense, line 59: `PYO3_CROSS_LIB_DIR` is kind of problematic for cross compiling from Inte/Apple Silicon to Apple Silicon/Intel, do we really need it in this situation?

.github/workflows/test.yml
```diff
@@ -35,6 +35,7 @@ jobs:
         with:
           profile: minimal
           toolchain: stable
+          target: aarch64-apple-darwin
```
konstin, line 38: I don't think that component should be installed on non-mac targets. Maybe make a separate step that's only run on mac?

.github/workflows/test.yml
konstin: I don't know, and unfortunately I don't have a mac so I unfortunately can't check. That's something from the pyo3 build script, so if that's a problem it should be discussed in pyo3/pyo3

src/build_context.rs
konstin: I think it should default to off

src/build_options.rs
```diff
@@ -68,6 +68,9 @@ pub struct BuildOptions {
     /// Use as `--rustc-extra-args="--my-arg"`
     #[structopt(long = "rustc-extra-args")]
     pub rustc_extra_args: Vec<String>,
+    /// Control whether to build universal2 wheel for macOS or not
```
konstin, line 71: Please add what this does on not-mac platforms

src/compile.rs
```diff
@@ -2,8 +2,10 @@ use crate::build_context::BridgeModel;
 use crate::BuildContext;
 use crate::PythonInterpreter;
 use anyhow::{anyhow, bail, Context, Result};
+use fat_macho::FatWriter;
 use fs_err::File;
 use std::collections::HashMap;
+use std::fs;
```
konstin, line 8:
```suggestion
use fs_err as fs;
```

tests/common/integration.rs
```diff
@@ -80,6 +80,21 @@ pub fn test_integration(package: impl AsRef<Path>, bindings: Option<String>) ->
 
         let python = target.get_venv_python(&venv_dir);
 
+        // Upgrade pip
```
konstin, line 83: I'd like to avoid doing this in tests because this is slow and prone to break things in hard to understand ways. 

Is there a special reason you did this here and would it be possible to replace it by e.g. ensuring that virtualenv uses a sufficiently recent pip version and otherwise error while suggesting `virtualenv --upgrade-embed-wheels`?

src/compile.rs
konstin: In case something goes wrong, this tells the user which file was at fault

src/build_options.rs
konstin: Also please add this to the readme (paste the output of `build --help` in the box and maybe a sentence somewhere that there's a universal2 opt-in for macs)

konstin (commented):
Looks good overall!

konstin:
Thanks again!

.github/workflows/test.yml
messense: Resolved by https://github.com/PyO3/pyo3/pull/1428

PR #406, opened 2021-01-22
Provide modern CMake in the docker image.

cjordan:
Hi!  Thanks for your work on `maturin`.

I have a [Rust package](https://github.com/MWATelescope/mwa_hyperbeam) that depends on the HDF5 library (usable with the [hdf5 crate](https://github.com/aldanor/hdf5-rust)). Because of manylinux compliance, I can't ship the Python wheel that I compile with `maturin`, as it would depend on a specific version of the HDF5 C library. However, the `hdf5` crate allows one to statically compile the HDF5 C library into Rust packages (great!). This would "just work", except that to build `hdf5` this way, it relies on the system having `CMake` > 3.10 available.

Looking at the CentOS image that the `maturin` docker images is based off of (6?), the most recent version `CMake` available in the repos is much older than 3.10.  This commit adds code to download and compile `CMake` 3.19 from source.  It may be convenient to other users of the `maturin` docker image to have this available by default, hence this PR.

I'm happy for the `CMake` version to be backed off a bit, if there are opinions on that. Perhaps more importantly, though, I couldn't get `CMake` to compile without `-DCMAKE_USE_OPENSSL=OFF`.  I'm not sure of all of the ramifications of doing this, but maybe this could be easily fixed?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,14 @@ RUN cargo rustc --bin maturin --manifest-path /maturin/Cargo.toml --release -- -
     && mv /maturin/target/release/maturin /usr/bin/maturin \
     && rm -rf /maturin
 
+ADD https://github.com/Kitware/CMake/releases/download/v3.19.1/cmake-3.19.1.tar.gz .
+RUN tar -xf cmake-3.19.1.tar.gz && \
+    cd cmake-3.19.1 && \
+    ./bootstrap -- -DCMAKE_USE_OPENSSL=OFF && \
+    make -j8 install && \
+    cd .. && \
+    rm -r cmake*
+
 WORKDIR /io
 
 ENTRYPOINT ["/usr/bin/maturin"]


konstin:
Thank you for contributing this, I appreciate it if people care about CI. However this workflow looks more appropriate for a project built with maturin rather than maturin itself - maturin itself only needs one python version for building and rather than the manylinux image and auditwheel it's statically linked with musl.

If you're interested in testing this workflow, I can give you access to the manylinux package on testpipy, but I fear that the current workflow will fail due to colliding py3 wheel.

konstin:
Hi! While I want to assist with getting more complex build systems running with manylinux+maturin, I think it would be better to keep the maturin image itself small and have more complex stuff in downstream images. E.g. for your case I'd image something like:

```dockerfile
FROM konstin2/maturin

ADD https://github.com/Kitware/CMake/releases/download/v3.19.1/cmake-3.19.1.tar.gz .
RUN tar -xf cmake-3.19.1.tar.gz && \
    cd cmake-3.19.1 && \
    ./bootstrap -- -DCMAKE_USE_OPENSSL=OFF && \
    make -j8 install && \
    cd .. && \
    rm -r cmake*

ENTRYPOINT ["/usr/bin/maturin"]
```

Would you be interested in contributing such a Dockerfile with a matching ci setup as an example?

Alternatively, you could also download precompiled cmake (https://github.com/Kitware/CMake/releases/download/v3.19.4/cmake-3.19.4-Linux-aarch64.tar.gz - it's a static binary) and mount it into the container.

konstin:
Update: I've added some more information to the readme (https://github.com/PyO3/maturin/commit/9f95d0af69291ab0811380f92bcb971d0c96cba8#diff-1550ec65ac92f65817fc28928dfef526912b5f52356ff43651369bae92f56031). Could you check if those instructions work for you or if there's anything missing/unclear?

cjordan:
Hey, thanks for getting back to me, and sorry for my delay in replying.

So I already made an [image](https://hub.docker.com/r/chjordan/maturin) extremely similar to what you proposed, although you make a good point on the CMake releases - I shall adopt that!

My main question is how I would contribute this. I'm not familiar with a CI setup and docker images, but if you could point me in the right direction, I'll happily do the work.

konstin:
I've added the dockerfile to https://github.com/konstin/complex-manylinux-maturin-docker/tree/main/custom-docker-image . If you want to add anything besides the dockerfile (e.g. a github actions configuration that uses that image), please add it there.

cjordan:
Thanks again!

PR #412, opened 2021-02-06
Build aarch/arm64 wheels for Python 3.7 - 3.9 using QEMU

hoffmang9:
I've created this workflow to build wheels for ARM64 using the [QEMU action](https://github.com/docker/setup-qemu-action).

I'm happy to change this to specify further Python supported version.

However, It may also make more sense for you to adapt your linux binary wheel strategy to use QEMU.

Open to your feedback, but this solves our use case for https://github.com/Chia-Network/clvm_rs

### DIFF
--- a/.github/workflows/build-aarch64.yml
+++ b/.github/workflows/build-aarch64.yml
@@ -1,0 +1,91 @@
+name: Build ARM64 wheels
+
+on:
+  push:
+    tags: [ 'v*' ]
+
+jobs:
+  build_wheels:
+    name: ARM64 Python Wheels on ubuntu-latest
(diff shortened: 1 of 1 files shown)


konstin:
Thank you for contributing this, I appreciate it if people care about CI. However this workflow looks more appropriate for a project built with maturin rather than maturin itself - maturin itself only needs one python version for building and rather than the manylinux image and auditwheel it's statically linked with musl.

If you're interested in testing this workflow, I can give you access to the manylinux package on testpipy, but I fear that the current workflow will fail due to colliding py3 wheel.

konstin:
This is now done by #422

PR #416, opened 2021-02-13
Use 'classifiers' instead of 'classifier'

### LINKED ISSUE #407: "classifer" should be "classifiers"
[Thanks for creating PyO3 and maturin!]

When I set
```
[package.metadata.maturin]
classifier=[
...
```
nothing appears in [Test.PyPi](https://test.pypi.org/project/bed_reader/0.1.5/). It should look like the lower left side of [this](https://pypi.org/project/bed-reader/0.1.2/). I noticed that maturin is also missing classifers on its [PyPi page](https://pypi.org/project/maturin/0.9.0/).

My guess is that the PyPi expects the keyword "classifers" (plural), not the singular "classifer" that maturin seems to be providing.

On a related note, there are other PyPi tags that I would like to use, but are missing from maturin. For example, "url" and "project_urls". It would be great if maturin could allow any tags through.

Thanks,
Carl

kngwyu:
Fixes #407. 
I confirmed it works by uploading a test crate: https://test.pypi.org/project/pyo3-upload-test/
Do we need some more tests?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ edition = "2018"
 
 [package.metadata.maturin]
 requires-python = ">=3.5"
-classifier = [
+classifiers = [
     "Topic :: Software Development :: Build Tools",
     "Programming Language :: Rust",
     "Programming Language :: Python :: Implementation :: CPython",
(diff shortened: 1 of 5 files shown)


konstin:
Thank you!

BTW I've opened https://github.com/pypa/warehouse/issues/9085 to ask what this discrepancy between core metadata and twine is about.

PR #417, opened 2021-02-14
Also release Windows x86 wheel to pypi

messense:


konstin:
Sorry, I wasn't aware that 32-bit windows is still a thing

PR #418, opened 2021-02-15
Fix build on FreeBSD on powerpc*

pkubaj:
FreeBSD uses powerpc64 and powerpc64le names for 64-bit POWER architecture.

However, they are actually subarchitectures of powerpc architecture, and Python returns powerpc,
even on 64-bits:
>>> platform.machine()
'powerpc'

The more correct way to check for architecture would be:
>>> platform.processor()
'powerpc64le'

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,7 +62,7 @@ class PostInstallCommand(install):
 
             cargo_args = [cargo, "rustc", "--bin", "maturin", "--message-format=json"]
 
-            if platform.machine() in ("ppc64le", "ppc64"):
+            if platform.machine() in ("ppc64le", "ppc64", "powerpc"):
                 cargo_args.extend(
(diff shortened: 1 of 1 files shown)


pkubaj:
This PR proposes adding `powerpc` instead of changing to `platform.processor()` and looking for `powerpc64`, `powerpc64le`, because I'm not sure whether `platform.processor()` and `platform.machine()` return the same strings on Linux (I only have FreeBSD installation).

programmerjake:
well, in TerMux, on aarch64 Android, `platform.processor()` returns `""`. On my Odroid XU4 (32-bit Arm Ubuntu), it returns `"armv7l"`. On my x86_64 desktop (Ubuntu), it returns `"x86_64"`. On a Power9 server (powerpc64le Debian) it returns `""`.

konstin:
Thank you!

PR #421, opened 2021-02-17
 Set executable bit on shared library

messense:
See the setuptools-rust side issue: https://github.com/PyO3/setuptools-rust/issues/109


konstin:
Didn't know about that, thanks!

PR #423, opened 2021-02-17
Fix building for musl libc

### LINKED ISSUE #154: Unable to install pyo3-pack on Alpine Linux
This is due to the lack of support for PEP 517 (See #2).

We should release a source distribution of pyo3-pack.

messense:
Closes #154

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -5,6 +5,7 @@ use anyhow::{anyhow, bail, Context, Result};
 use fat_macho::FatWriter;
 use fs_err::{self as fs, File};
 use std::collections::HashMap;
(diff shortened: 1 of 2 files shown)


messense:
cc @thedrow

konstin:
Thank you!

messense:
Seems that `-C target-feature=-crt-static` breaks https://github.com/PyO3/maturin/runs/1918499503

konstin:
Yeah, I've overlooked that it needs to be guarded on libraries only. Should be fixed in 53dd4c2a2b548e8e9a1a390dccf161f61ee6137e

Issue #376, opened 2020-11-02
Manylinux default version


programmerjake:
I encountered an issue with manylinux support for architectures not included in manylinux2010 (powerpc64le-unknown-linux-gnu for me):
Maturin's argument parsing defaults to manylinux2010, which causes pip to not be able to install the built wheel:
```
ERROR: power_instruction_analyzer-0.2.0-cp38-cp38-manylinux2010_ppc64le.whl is not a supported wheel on this platform.
```

What do you think of refactoring argument parsing to pick the default manylinux version based on the target arch? something like this:
```
if manylinux arg passed then
    use arg
else if i686 or x86_64 then // list from PEP 571
    manylinux2010
else if armv7l or aarch64 or ppc64 or ppc64le or s390x then // list from PEP 599
    manylinux2014
else
    off
end
```

programmerjake:
Related to #367

programmerjake:
An alternative option is to just default to manylinux2014 for supported platforms (otherwise it defaults to off) and people have to explicitly use `--manylinux=2010` when they really need support for such an old OS.

konstin:
Sounds like a good idea! Implementation-wise, the manylinux could have an `auto` value that takes the lowest value currently supported. (It would be even better if we could take the lowest version the built binary supports, but unfortunately I haven't figured out how to port that auditwheel part that checks what glibc version is linked to rust)

messense:
Bitten by this in https://github.com/messense/maturin/runs/1915790218?check_suite_focus=true , it happily builds `maturin-0.9.1-py3-none-manylinux2010_aarch64.whl` but only aarch64 wheel only works on manylinux2014 and later. We should have a better default version.

messense:
Fixed by #424

PR #424, opened 2021-02-18
Change manylinux default version based on target arch

messense:
Ref https://github.com/PyO3/maturin/issues/376

### DIFF
--- a/src/build_options.rs
+++ b/src/build_options.rs
@@ -24,14 +24,16 @@ pub struct BuildOptions {
     /// will be rejected by pypi, unless they are separately validated by
     /// `auditwheel`.
     ///
+    /// The default is the lowest supported value for a target, which is 2010 for x86 and 2014 for
+    /// arm and powerpc.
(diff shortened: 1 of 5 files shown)


kngwyu (commented):
Looks nice 👍🏼  
I guess we also need to edit `Manylinux and auditwheel` section in README?

src/build_options.rs
```diff
@@ -29,9 +29,8 @@ pub struct BuildOptions {
         long,
         possible_values = &["2010", "2014", "off"],
         case_insensitive = true,
-        default_value = "2010"
```
konstin, line 32: Please update the doc comment saying something like "The default is the lowest supported value for a target, which is 2010 for x86 and 2014 for arm and powerpc".

konstin (approved):
Thank you, I really want to have that. Code looks good, just needs updated documentation

konstin:
Thanks!

PR #426, opened 2021-02-18
Use -crt-static if compiling maturin itself under musl libc

thedrow:
This should help us build maturin in Alpine.


messense:
I was under the impression that maturin build fines on Alpine?

thedrow:
Yes, it wasn't obvious I also require gcc.

PR #429, opened 2021-02-19
Refactor target triple handling

messense:
Currently `maturin build --target xxx --universal2` would results in two `--target xxx` passed to `cargo rustc` which errored with:

```
error: specifying multiple `--target` flags requires `-Zmultitarget`
💥 maturin failed
  Caused by: Failed to build a aarch64 library through cargo
  Caused by: Cargo build finished with "exit code: 101": `cargo rustc --message-format json --manifest-path Cargo.toml --target x86_64-apple-darwin --release --target aarch64-apple-darwin --bins --`
```

This patch changes how `target` is passed to `compile_target` function by remove it from `cargo_extra_args` and pass it directly into `compile_target` function.

### DIFF
--- a/src/target.rs
+++ b/src/target.rs
@@ -82,6 +82,7 @@ pub struct Target {
     os: OS,
     arch: Arch,
     env: Option<Env>,
+    triple: &'static str,
 }
 
 impl Target {
@@ -145,6 +146,7 @@ impl Target {
             os,
             arch,
             env: platform.target_env,
+            triple: platform.target_triple,
         })
     }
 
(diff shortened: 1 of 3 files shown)


src/target.rs
```diff
@@ -145,6 +146,7 @@ impl Target {
             os,
             arch,
             env: platform.target_env,
+            triple: platform.target_triple,
```
konstin, line 149: I hope that this is really the same target as cargo uses and we don't start cross-compiling where we wouldn't need to :crossed_fingers:

konstin:
Thanks!

PR #431, opened 2021-02-19
Build i686 musl wheel on GitHub Action

messense:
A lot of time spent on installing maturin to build wheel for 32-bit linux: https://github.com/messense/rjieba-py/runs/1933811414?check_suite_focus=true

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -126,8 +126,9 @@ jobs:
     strategy:
       matrix:
         platform: [
-          { target: "aarch64-unknown-linux-musl", image_tag: "aarch64-musl" },
-          { target: "armv7-unknown-linux-musleabihf", image_tag: "armv7-musleabihf" },
+          { target: "aarch64-unknown-linux-musl", image_tag: "aarch64-musl", manylinux: "2014" },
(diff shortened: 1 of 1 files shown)


konstin:
Afaik we can build this as a regular target on ubuntu-latest, so no need to go through cross, which I'd prefer

messense:
I was originally try to use a regular target, but it now has too many `if`s makes things complicated.

konstin:
Do you think https://github.com/PyO3/maturin/compare/github-actions-i686-unknown-linux-musl would work?

messense:
I think that should work, thanks!

messense:
Closing in favor of https://github.com/PyO3/maturin/commit/64cf56dd1600142d71729df22b075a8ba97bc7ba

konstin:
My version seems to fail: https://github.com/PyO3/maturin/runs/1963814800?check_suite_focus=true

Do you by chance know why?

messense:
Hmm, I am not sure. Looked at [ring's build script](https://github.com/briansmith/ring/blob/930b97b0c95acebf1ae61698c0ed3c171cac00ea/mk/install-build-tools.sh#L65), it uses clang instead of gcc for musl targets.

PR #433, opened 2021-02-19
Set a more reasonable LC_ID_DYLIB entry on macOS

messense:
cc https://github.com/PyO3/setuptools-rust/pull/119

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -133,10 +133,30 @@ fn compile_target(
         }
     }
 
+    let module_name = &context.module_name;
+    let so_filename = match python_interpreter {
+        Some(python_interpreter) => python_interpreter.get_library_name(module_name),
+        // abi3
+        None => {
+            format!("{base}.abi3.so", base = module_name)
+        }
+    };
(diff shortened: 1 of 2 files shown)


src/compile.rs
```diff
@@ -133,10 +133,27 @@ fn compile_target(
...
 
+    let module_name = &context.module_name;
+    let so_filename = match python_interpreter {
+        Some(python_interpreter) => python_interpreter.get_library_name(module_name),
+        // abi3
+        None => {
+            format!("{base}.abi3.so", base = module_name)
+        }
+    };
+    let macos_dylib_install_name = format!("link-args=-Wl,-install_name,@rpath/{}", so_filename);
```
konstin, line 144: Please add a comment what this does / why this is required

konstin (approved):

src/compile.rs
messense: How about a link to https://github.com/PyO3/setuptools-rust/issues/106 ?

konstin: Yep, that should do

PR #434, opened 2021-02-20
Support configure macOS universal2 build with ARCHFLAGS

messense:
See https://github.com/PyO3/setuptools-rust/pull/115#pullrequestreview-593692923

### DIFF
--- a/src/build_options.rs
+++ b/src/build_options.rs
@@ -9,7 +9,8 @@ use crate::Target;
 use anyhow::{bail, format_err, Context, Result};
 use cargo_metadata::{Metadata, MetadataCommand, Node};
 use serde::{Deserialize, Serialize};
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
+use std::env;
 use std::io;
 use std::path::PathBuf;
 use structopt::StructOpt;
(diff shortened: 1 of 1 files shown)


konstin:
Sorry, I don't fully understand what `ARCHFLAGS` does. Is the idea that you can somehow request a target architecture through that and do you have an example on how one would use that?

messense:
> Sorry, I don't fully understand what `ARCHFLAGS` does. Is the idea that you can somehow request a target architecture through that and do you have an example on how one would use that?

That's right. You can use `ARCHFLAGS` to tell C/C++ compiler to build for multiple architectures. For example `ARCHFLAGS="-arch x86_64 -arch arm64"` instruct compiler to build for both `x86_64` & `arm64`.

https://github.com/python/cpython/blob/a5ed2fe0eedefa1649aa93ee74a0bafc8e628a10/Lib/_osx_support.py#L260

messense:
Basically `ARCHFLAGS` make it consistent with C/C++ way of creating multiarch binary, also consistent with setuptools-rust and cibuildwheel. If [rustc/cargo can produce universal macos binary in the future](https://github.com/rust-lang/cargo/issues/8875), I think they might want to use it too to play nice with C dependencies.

konstin:
Ah, that's cool!

PR #435, opened 2021-02-20
Use sys.executable instead of "python" when invoking pep517 command

### LINKED ISSUE #410: Basic pip3 install with pyproject.toml fails
When trying to install a Python package locally with `pip3 install .` and the default `pyproject.toml` file, I get the following error:

```
  Running `maturin pep517 build-wheel -i python`
    File "<string>", line 19, in <module>
  AttributeError: 'module' object has no attribute 'base_prefix'
  💥 maturin failed
    Caused by: The given list of python interpreters is invalid
    Caused by: python is not a valid python interpreter
    Caused by: Trying to get metadata from the python interpreter 'python' failed
  🔗 Found pyo3 bindings
  Error: command ['maturin', 'pep517', 'build-wheel', '-i', 'python'] returned non-zero exit status 1
```

This succeeds if the wheel is built instead manually with `maturin pep517 build-wheel -i python3` (note Python3), so this is hopefully an easy fix!

My project builds correctly with `maturin build` and also `maturin develop`.

 * Your python version: `3.9.1`
 * Your pip version: `pip3 21.0.1`
 * The version of the bindings you're using: `pyo3 0.13.1`
 * Does `cargo build` work? Yes

messense:
Fixes #410

### DIFF
--- a/maturin/__init__.py
+++ b/maturin/__init__.py
@@ -54,8 +54,9 @@ def get_config_options() -> List[str]:
 
 # noinspection PyUnusedLocal
 def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
-    # The PEP 517 build environment guarantees that `python` is the correct python
-    command = ["maturin", "pep517", "build-wheel", "-i", "python"]
+    # PEP 517 specifies that only `sys.executable` points to the correct
+    # python interpreter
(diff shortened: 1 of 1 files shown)


konstin:
Thank you!

PR #436, opened 2021-02-21
Enforce manylinux symbol versions check

### LINKED ISSUE #303: make it harder to do the wrong thing with manylinux?
As part of solving https://github.com/oconnor663/blake3-py/issues/5, I learned that I was using Maturin incorrectly. Previously I thought that `maturin build` automatically took care of all the manylinux details, but now I understand that you need to run the build in a special Docker container to get this right. I think the docs could be clarified around this point (like rewording the remarks about the `--manylinux` flag, and maybe adding some bold/red text highlighting that the Docker command is **really important**), but I also want to suggest an underlying issue: It's easier to do the wrong thing than to do the right thing.

What I mean by this, is that the first thing any new user is going to try (because it's documented prominently) is just `maturin build --release`. Then what happens is they test the result of that build, and they find that it works. It looks like everything's good, and there's no indication that they've produced a broken artifact that's not going to work on other people's systems. The ease of running `maturin build` directly is a footgun on Linux, tempting people to do the wrong thing.

I'm not super familiar with Maturin or manylinux myself, and this project was my first time using both, so I can't tell whether the following is a dumb idea or not, but here's a suggestion: Have `maturin build` (or maybe just `maturin build --release`) return an error by default on Linux. It could say something Clippy style, like "hey, I noticed you're trying to build a Python wheel for Linux, did you know that this will fail on older systems if you don't run the following Docker command instead? Alternatively, if you know what you're doing, please use the `--i-know-what-im-doing` flag." That would make it much less likely that other new users make the same mistake I made.

messense:
Closes #303


messense:
Hmm... 

> Your library is not manylinux compliant because it links the following forbidden libraries: ["libc.so.6 version 2.18,2.14,2.17"]

messense:
Now you can figure out which symbol is not compliant:

> Your library is not manylinux compliant because it links the following forbidden libraries: ["libc.so.6 offending symbols:  clock_gettime@@GLIBC_2.17, __cxa_thread_atexit_impl@@GLIBC_2.18, memcpy@@GLIBC_2.14"]

konstin:
Huge thank you for this, I wanted to have that for a long time!

src/auditwheel/mod.rs
```diff
@@ -0,0 +1,249 @@
...
+            let mut parts = v.splitn(2, '_');
+            let name = parts.next().unwrap();
+            let version = parts.next().unwrap();
+            versions
+                .entry(name.to_string())
+                .or_default()
+                .insert(version.to_string());
+        }
+        let arch = target.target_arch().to_string();
+        let arch_versions = &policy.symbol_versions[&arch];
```
konstin, line 214: Does this panic when using an unsupported? In that case it would be better to use `.get` and error appropriately

src/auditwheel/mod.rs
```diff
@@ -0,0 +1,249 @@
...
+    "Your library is not manylinux compliant because it links the following forbidden libraries: {0:?}",
+    )]
+    ManylinuxValidationError(Vec<String>),
+}
+
+/// Structure of "version needed" entries is documented in
+/// https://refspecs.linuxfoundation.org/LSB_3.0.0/LSB-PDA/LSB-PDA.junk/symversion.html
+#[derive(Clone, Copy, Debug, Pread)]
+#[repr(C)]
+struct GnuVersionNeed {
```
konstin, line 47: Could those two structs (and maybe that relevant reading part) be upstreamed to goblin?

src/auditwheel/mod.rs
```diff
@@ -0,0 +1,249 @@
...
+                let offending_symbol_versions: Vec<String> = offending_versions
+                    .iter()
+                    .map(|v| format!("{}_{}", name, v))
+                    .collect();
+                let offending_symbols = find_incompliant_symbols(&elf, &offending_symbol_versions)?;
+                let offender = format!(
+                    "{} offending symbols:  {}",
+                    library.name,
+                    offending_symbols.join(", ")
+                );
```
konstin, line 231: This will actually make nicer error messages than auditwheel!

src/auditwheel/mod.rs
```diff
@@ -0,0 +1,249 @@
...
+                    .difference(versions_allowed)
+                    .map(|v| v.as_ref())
+                    .collect();
+                let offending_symbol_versions: Vec<String> = offending_versions
+                    .iter()
+                    .map(|v| format!("{}_{}", name, v))
+                    .collect();
+                let offending_symbols = find_incompliant_symbols(&elf, &offending_symbol_versions)?;
+                let offender = format!(
+                    "{} offending symbols:  {}",
```
konstin, line 228:
```suggestion
                    "{} offending symbols: {}",
```

konstin:
eb3c4a5b3e6b9439bd9c6b9e494c1a277a821712 should hopefully fix the tests

src/auditwheel/mod.rs
messense: I'd love to, but I'll do it separately to not block this.

konstin:
Thanks again!

PR #438, opened 2021-02-22
Add a workaround for the GitLab PyPI API implementation

ravenexp:
GitLab includes its own PyPI upload API implementation
that is not based on Warehouse.
It insists that "requires_python" metadata field must
always be sent, even when it's empty.

The package upload to GitLab PyPI fails with:

    400 Bad Request: {"error":"requires_python is missing"}

when it's not included in the request.

The official API reference does not mandate this, though:

    https://warehouse.readthedocs.io/api-reference/legacy.html

Unfortunately, twine always includes these metadata fields
in its requests, even when they are empty, and the GitLab
API implementation happened to depend on this.

Add a workaround that mimics twine behavior and silently
provides "requires_python" field even when the package
developers do not define "Requires-Python".

### DIFF
--- a/src/upload.rs
+++ b/src/upload.rs
@@ -105,6 +105,12 @@ pub fn upload(registry: &Registry, wheel_path: &Path) -> Result<(), UploadError>
     add_option("requires_python", &metadata.requires_python);
     add_option("summary", &metadata.summary);
 
+    if metadata.requires_python.is_none() {
+        // GitLab PyPI repository API implementation requires this metadata field
+        // and twine always includes it in the request, even when it's empty.
+        api_metadata.push(("requires_python", "".to_string()));
+    }
+
     let mut add_vec = |name, values: &[String]| {
         for i in values {
             api_metadata.push((name, i.clone()));


konstin:
Thank you for the pull request, but I think that this should be fixed on gitlab's side: https://gitlab.com/gitlab-org/gitlab/-/issues/322434

I'll leave this open until there is a response from gitlab.

ravenexp:
I agree that GitLab is the root cause of this issue. However, even if it's fixed in the next GitLab software point release and on the gitlab.com public website, there are still thousands of private GitLab software installations used by companies and big OSS projects.
It may take years to update these. I won't be surprised if most of them are still on GitLab 12.x or older.

There's of course a workaround that does not require modifying maturin, such as manually adding `requires-python` field to `[package.metadata.maturin]` section in project's Cargo.toml. But then the users will have to figure it out on their own in every time someone tries to use maturin in place of twine.

messense:
@ravenexp Can we  just detect the `requires_python is missing` error and tell user to add `requires-python` field to `[package.metadata.maturin]` section in project's `Cargo.toml`?

ravenexp:
> @ravenexp Can we just detect the `requires_python is missing` error and tell user to add `requires-python` field to `[package.metadata.maturin]` section in project's `Cargo.toml`?

Sound good to me. It's definitely an UX improvement compared to

    400 Bad Request: {"error":"requires_python is missing"}

I believe the implementation of this suggestion belongs to a separate PR though.
And this one will have to stay as it is, because it's already referenced in the GitLab issue tracker.

ravenexp:
Unfortunately the GitLab issue seems to have stalled, and telling the user to add a field to `[package.metadata.maturin]` is no longer a viable option because `maturin upload` can now upload third-party wheel files.

I rewrote the workaround to make it less invasive and compatible with the upload command code.

ravenexp:
Still no progress on the GitLab issue, unfortunately. I've updated the workaround for the current `upload()` implementation.

messense:
I'm not opposed to merge this, but I'll let @konstin decide.

konstin:
I dislike adding custom workarounds, but I think this is ok to merge given how important gitlab is

PR #439, opened 2021-02-22
Fix `sdist-include` for PEP 517 compliant packages

koehlma:
Hey there!

This PR fixes the following issues:
- As of commit 036fff7adb9ce08fa7a31778aa39be53e25e1641, when building a source distribution the files are placed in a folder `{NAME}-{VERSION}` as required by PEP 517. However, this has not been the case for additional files included via `sdist-include`. With this PR, those additional files are placed inside the `{NAME}-{VERSION}` directory as well.
- The previous usage of `glob` did not search for files relative to `manifest_dir` (at least as far as I can tell).
- It seems like `target` and `source` have previously been been swapped when adding files with `add_file`.


konstin:
Thank you!

PR #440, opened 2021-02-23
Respect CARGO_BUILD_TARGET environment variable

messense:
https://doc.rust-lang.org/cargo/reference/environment-variables.html

> CARGO_BUILD_TARGET — The default target platform

`cargo rustc` will use it automatically if set but maturin isn't aware of it so maturin creates wrong manylinux wheel for the host target not the specified target.

### DIFF
--- a/src/build_options.rs
+++ b/src/build_options.rs
@@ -58,7 +58,7 @@ pub struct BuildOptions {
     #[structopt(long = "skip-auditwheel")]
     pub skip_auditwheel: bool,
     /// The --target option for cargo
-    #[structopt(long, name = "TRIPLE")]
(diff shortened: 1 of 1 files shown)


src/build_options.rs
```diff
@@ -58,7 +58,7 @@ pub struct BuildOptions {
     #[structopt(long = "skip-auditwheel")]
     pub skip_auditwheel: bool,
     /// The --target option for cargo
-    #[structopt(long, name = "TRIPLE")]
+    #[structopt(long, name = "TRIPLE", env = "CARGO_BUILD_TARGET")]
```
konstin, line 61: Didn't know about `env =`, this is really neat

konstin:
Thank you!

PR #443, opened 2021-02-24
maturn develop install dependencies automatically

### LINKED ISSUE #414: maturin develop does not install dependencies
When adding a section
```
 [package.metadata.maturin]
requires-dist = ["flask~=1.1.0", "toml==0.10.0"]
```
in Cargo.toml, the dependencies are not installed in the current virtualenv when running `maturin develop`.
Is this the expected behavior ?

If so, it is a bit annoying: the simplest way I found for installing the python dependencies specified in Cargo.toml is to build the wheel and install it. Is there a better solution ? I believe this could be mentioned in the README.

 * Your python version (`python -V`): 3.7.3
 * Your pip version (`pip -V`): 21.0.1
   * If this isn't the [latest pip version](https://pip.pypa.io/en/stable/news/), have you checked using the latest pip version? This is the latest pip version
 * The version of the bindings you're using, if any (e.g. `pyo3`, `rust-cpython` or `cffi`): pyo3=0.13
 * Does `cargo build` work? yes
 * If on windows, have you checked that you aren't accidentally using unix path (those with the forward slash `/`)? N/A (on linux)

messense:
Closes #414

### DIFF
--- a/src/develop.rs
+++ b/src/develop.rs
@@ -5,9 +5,10 @@ use crate::Manylinux;
 use crate::PythonInterpreter;
 use crate::Target;
 use crate::{write_dist_info, BuildOptions};
-use anyhow::{anyhow, format_err, Context, Result};
+use anyhow::{anyhow, bail, format_err, Context, Result};
 use fs_err as fs;
 use std::path::Path;
+use std::process::Command;
 
 /// Installs a crate by compiling it and copying the shared library to site-packages.
 /// Also adds the dist-info directory to make sure pip and other tools detect the library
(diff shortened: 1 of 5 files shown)


src/develop.rs
```diff
@@ -46,6 +47,28 @@ pub fn develop(
...
+                .iter()
+                .map(|x| x.as_str()),
+        );
+        let mut pip_install = Command::new(&interpreter.executable)
+            .args(&args)
+            .spawn()
+            .context("Failed to run pip install")?;
+        let status = pip_install
+            .wait()
+            .expect("Failed to wait on pip install process");
```
konstin, line 66:
```suggestion
        let status = Command::new(&interpreter.executable)
            .args(&args)
            .status()
            .context("Failed to run pip install")?;
```

konstin (approved):
Thanks, looks good!

PR #451, opened 2021-02-26
Add manylinux_2_24 support

messense:
PEP 600: https://www.python.org/dev/peps/pep-0600/
Tracking issue: https://github.com/pypa/manylinux/issues/542


konstin:
Thanks!

PR #452, opened 2021-02-26
Improve error message when auditwheel failed to find versioned offending symbols

### LINKED ISSUE #447: auditwheel error reporting fails for maturin binary
Running `cargo run -- build -b bin --strip` in the maturin repository on my not manylinux2010 compliant machine produces:

```
💥 maturin failed
  Caused by: Failed to ensure manylinux2010 compliance
  Caused by: Your library is not manylinux compliant because it links the following forbidden libraries: ["libc.so.6 offending symbols: ", "libm.so.6 offending symbols: "]
```

`println!("{:?}", offending_symbol_versions);` however gives me:

```
["GLIBC_2.29", "GLIBC_2.27"]
["GLIBC_2.14", "GLIBC_2.15", "GLIBC_2.18", "GLIBC_2.17"]
```

I think `find_incompliant_symbols` fails here because we there's no @ in the symbol names

messense:
Closes #447

### DIFF
--- a/src/auditwheel/mod.rs
+++ b/src/auditwheel/mod.rs
@@ -232,11 +232,19 @@ pub fn auditwheel_rs(
                     .map(|v| format!("{}_{}", name, v))
                     .collect();
                 let offending_symbols = find_incompliant_symbols(&elf, &offending_symbol_versions)?;
-                let offender = format!(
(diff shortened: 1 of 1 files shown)


konstin:
Thanks!

PR #453, opened 2021-02-26
 Fix rewriting of path dependencies and include only path dependencies of root crate

### LINKED ISSUE #449: Failure to package workspace with non-existinent lockilfes
It seems to be looking for lockfiles that don't necessarily exist here. With the same layout as https://github.com/PyO3/maturin/issues/291#issuecomment-758328862 and lightly anonymized:

```
$ maturin --version
maturin 0.10.0-beta.1
$ maturin build -m py/Cargo.toml
🔗 Found pyo3 bindings
🐍 Found CPython 3.9 at python3.9
💥 maturin failed
  Caused by: Failed to add local dependency PACKAGE at .../base to the source distribution
  Caused by: Failed to add file from .../base/Cargo.lock to sdist as PACKAGE/local_dependencies/PACKAGE/Cargo.lock
  Caused by: No such file or directory (os error 2) when getting metadata for .../base/Cargo.lock
```

I have one Cargo.lock file next to the workspace's Cargo.toml, and created `py/pyproject.toml` so it will create an sdist. If no sdist is built, then things still seem to work (but of course there's no sdist).

_Originally posted by @tari in https://github.com/PyO3/maturin/issues/291#issuecomment-784507757_

koehlma:
Hey there!

This PR contains the following changes:
- ~~A fix for issue #449.~~ It seems like it fixed only a similar problem for me.
- It correctly rewrites the path to path dependencies of path dependencies. Previously these paths have been rewritten relative to the root directory of the source distribution causing Cargo not to find them on installation. 
- It changes the way path dependencies are collected. In particular, the new method only collects path dependencies of the actual root crate (and their path dependencies) but not all path dependencies of the Cargo workspace.

### DIFF
--- a/src/source_distribution.rs
+++ b/src/source_distribution.rs
@@ -21,7 +21,8 @@ const LOCAL_DEPENDENCIES_FOLDER: &str = "local_dependencies";
 /// This method is rather frail, but unfortunately I don't know a better solution.
 fn rewrite_cargo_toml(
     manifest_path: impl AsRef<Path>,
-    known_path_deps: &HashMap<String, String>,
+    known_path_deps: &HashMap<&String, &PathBuf>,
+    is_path_dep: bool,
 ) -> Result<String> {
     let text = fs::read_to_string(&manifest_path).context(format!(
         "Can't read Cargo.toml at {}",
@@ -53,8 +54,11 @@ fn rewrite_cargo_toml(
                     )
                 }
                 // This is the location of the targeted crate in the source distribution
-                table[&dep_name]["path"] =
-                    toml_edit::value(format!("{}/{}", LOCAL_DEPENDENCIES_FOLDER, dep_name));
+                table[&dep_name]["path"] = toml_edit::value(if is_path_dep {
+                    format!("../{}", dep_name)
+                } else {
+                    format!("{}/{}", LOCAL_DEPENDENCIES_FOLDER, dep_name)
+                });
                 if !known_path_deps.contains_key(&dep_name) {
                     bail!(
                         "cargo metadata does not know about the path for {} {} present in {}, which should never happen ಠ_ಠ",
@@ -75,8 +79,9 @@ fn add_crate_to_source_distribution(
     writer: &mut SDistWriter,
     manifest_path: impl AsRef<Path>,
     prefix: impl AsRef<Path>,
-    known_path_deps: &HashMap<String, String>,
+    known_path_deps: &HashMap<&String, &PathBuf>,
     root_crate: bool,
+    is_path_dep: bool,
 ) -> Result<()> {
     let output = Command::new("cargo")
         .args(&["package", "--list", "--allow-dirty", "--manifest-path"])
@@ -123,7 +128,7 @@ fn add_crate_to_source_distribution(
         )
     }
 
-    let rewritten_cargo_toml = rewrite_cargo_toml(&manifest_path, &known_path_deps)?;
+    let rewritten_cargo_toml = rewrite_cargo_toml(&manifest_path, known_path_deps, is_path_dep)?;
 
     writer.add_directory(&prefix)?;
     writer.add_bytes(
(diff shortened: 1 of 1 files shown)


konstin:
Hi! It seems the fix required to make transitive path dependencies work is only a small section of these changes. I've cherry-picked the corresponding section into a new commit together with a test case. If there are still cases missing, please open an issue!

src/source_distribution.rs
```diff
@@ -139,7 +143,38 @@ fn add_crate_to_source_distribution(
...
+        .context("Expected the dependency graph to have a root package")?;
+    // scan the dependency graph for path dependencies
+    let mut path_deps = HashMap::new();
+    let mut stack: Vec<&cargo_metadata::Package> = vec![root];
+    while let Some(top) = stack.pop() {
+        for dependency in &top.dependencies {
+            if let Some(path) = &dependency.path {
+                path_deps.insert(&dependency.name, path);
+                // we search for the respective package by `manifest_path`, there seems
+                // to be no way to query the dependency graph given `dependency`
```
konstin, line 159: The cargo metadata output for the root package contains everything you need to know, no need to run it multiple times (see resolve.nodes)

koehlma:
> If there are still cases missing, please open an issue!

It does not miss any path dependencies but instead includes crates which are not path dependencies of the root crate being packaged (the Python package crate). Iterating over `resolver.nodes` does not trace the dependencies of the actual crate being packaged but instead includes all *path crates* of the Cargo workspace (https://github.com/PyO3/maturin/commit/3f2a9c532e7cc223636f256e91c033dc79e67a12 does not solve this):

https://github.com/PyO3/maturin/blob/929dbcd2dea3debdad1c34bac2e0f21107633c77/src/source_distribution.rs#L171-L177

I might be able to provide a test case for this issue. I have something like this in mind
```
workspace/python-crate
workspace/generic-library-crate
workspace/helper-tool
```
where `python-crate` depends on `generic-library-crate` but not on `helper-tool`. The current approach seems to unnecessarily include `helper-tool` in the source distribution. In my case `helper-tool` is a binary crate while `generic-library-crate` is a pure library crate. Packaging does then trigger #449 because it tries to package `Cargo.lock` for `helper-binary` rooted in the fact that `cargo list` does list `Cargo.lock` for binary crates despite the fact that this file does not exist.

src/source_distribution.rs
koehlma: I am aware of `resolve.nodes`. Hower, the [`Dependency`](https://docs.rs/cargo_metadata/0.12.3/cargo_metadata/struct.Dependency.html) obtained by iterating over `top.dependencies` does not contain a [`PackageId`](https://docs.rs/cargo_metadata/0.12.3/cargo_metadata/struct.PackageId.html) field which would be necessary to locate the respective dependency in `cargo_metadata.packages` or in `resolve.nodes`. Also, the dependency graph is transitivity reduced and, as I argued here https://github.com/PyO3/maturin/pull/453#issuecomment-788652178, just iterating over its nodes does include unnecessary crates into the source distribution. Hence, it seems to me, one needs to actually traverse the graph to find all the actual (transitive) path dependencies of the crate.

Now, traversing `resolve.nodes` has an output-sensitive complexity of *O(nm)* where *n* is the amount of (transient) path dependencies and *m* is the number of nodes of the graph. As `resolve.nodes` is a `Vec` we have to scan it linearly given the package ids obtained via a node's [`Node.dependencies`](https://docs.rs/cargo_metadata/0.12.3/cargo_metadata/struct.Node.html#structfield.dependencies). This is the same complexity as my proposed change. What would help, though, is to first build a data structure supporting more efficient queries for dependencies than linearly scanning. Maybe, I miss something here but I cannot see how `resolve.nodes` helps—it represents the dependency graph but not in a way that enables efficient queries for (transient) path dependencies.

In addition, working with [`Dependency`](https://docs.rs/cargo_metadata/0.12.3/cargo_metadata/struct.Dependency.html) has the benefit that it is clearly documented how to identify path dependencies. In contrast, the `repr` field of [`PackageId`](https://docs.rs/cargo_metadata/0.12.3/cargo_metadata/struct.PackageId.html) is explicitly considered an implementation detail.

PR #454, opened 2021-02-27
Better cross compiling support for PyO3 binding on Unix

### LINKED ISSUE #446: Cross compilation ignores PYO3_CROSS_PYTHON_VERSION
 * Your python version (`python -V`): host Python 3.9, cross target Python 3.6
 * Your pip version (`pip -V`): -
   * If this isn't the [latest pip version](https://pip.pypa.io/en/stable/news/), have you checked using the latest pip version?
 * The version of the bindings you're using, if any (e.g. `pyo3`, `rust-cpython` or `cffi`): pyo3
 * Does `cargo build` work? yes
 * If on windows, have you checked that you aren't accidentally using unix path (those with the forward slash `/`)? -

https://github.com/messense/manylinux2014-cross-arm/runs/1975989661?check_suite_focus=true

Set`PYO3_CROSS_PYTHON_VERSION =3.6` and `PYO3_CROSS_LIB_DIR ` according to https://pyo3.rs/v0.13.2/building_and_distribution.html#cross-compiling , maturin builds fine but produces `cp39-cp39` wheels instead of `cp36-cp36m`.

messense:
Fixes #446 

I've tested it in https://github.com/messense/manylinux2014-cross-arm/actions/runs/605207408 and verified the built wheels with Docker.

### DIFF
--- a/src/build_options.rs
+++ b/src/build_options.rs
@@ -1,5 +1,6 @@
 use crate::auditwheel::Manylinux;
 use crate::build_context::{BridgeModel, ProjectLayout};
+use crate::cross_compile::{find_sysconfigdata, is_cross_compiling, parse_sysconfigdata};
 use crate::python_interpreter::InterpreterKind;
 use crate::BuildContext;
 use crate::CargoToml;
@@ -8,14 +9,12 @@ use crate::PythonInterpreter;
 use crate::Target;
 use anyhow::{bail, format_err, Context, Result};
 use cargo_metadata::{Metadata, MetadataCommand, Node};
-use fs_err::{self as fs, DirEntry};
 use regex::Regex;
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, HashSet};
 use std::env;
 use std::io;
-use std::path::{Path, PathBuf};
-use std::process::{Command, Stdio};
+use std::path::PathBuf;
 use structopt::StructOpt;
 
 /// High level API for building wheels from a crate which is also used for the CLI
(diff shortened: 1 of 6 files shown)


messense:
The `sysconfigdata` parsing code is mostly copied from `pyo3`'s `build.rs` with some minor adjustments. I'm not sure whether it's better to extract a shared crate for this. cc @davidhewitt

src/build_options.rs
```diff
@@ -469,6 +514,39 @@ fn split_extra_args(given_args: &[String]) -> Result<Vec<String>> {
     Ok(splitted_args)
 }
 
+fn is_cross_compiling(target: &Target) -> Result<bool> {
```
messense, line 517: This function diverges a lot from the pyo3 version because it's not used in build script and have no access to some build time environment variables.

davidhewitt:
> I'm not sure whether it's better to extract a shared crate for this.

Interesting question. I'd been looking at refactoring pyo3's configuration into a shared crate previously, though I hadn't quite finished up a design. I can try to push what I have at some point over the next week.

davidhewitt:
Btw I haven't forgotten about this - what I was thinking of above is what I've described on https://github.com/PyO3/pyo3/issues/1548

My branch that I was originally trying out is now a bit outdated because there have been a lot of improvements to the pyO3 build.rs recently. If you think what I propose in that issue is useful here then I'm happy to remake it.

konstin:
Should I review this or should I wait for changes in pyo3?

messense:
> Should I review this or should I wait for changes in pyo3?

I think we can push this to finish line without waiting for changes in pyo3, it just means some duplicated code between maturin and pyo3, we can refactor it later and maybe extract a shared crate out of them.

konstin:
Please move the bulk of the cross compiling logic to a new `cross_compile.rs`, otherwise the code looks good!

konstin (approved):

Issue #448, opened 2021-02-25
auditwheel check should pass on default


konstin:
Currently, `maturin build` will fail on most machines because they are not manylinux2010 compliant. The default should either be changed to `off` or to autodetect the lowest possible manylinux version before the final 0.10.0 release.

messense:
Autodetect lowest possible manylinux version is possible, but on non manylinux2010/2014 compliant machines `maturin build` would still fail. I think the it's ok to fail with `maturin build` because you just shouldn't distribute wheels built on non manylinux compliant machines. For local development, people should prefer `maturin develop` over `maturin build`. But we certainly should improve the error message a little to tell user to use  the `--manylinux off` option.

messense:
Or we can add a `auto` choice for `--manylinuux` and switch default to `off` and autodetect lowest possible manylinux version with `--manylinux auto`

PR #456, opened 2021-02-28
auditwheel choose higher priority tag when possible

messense:
Example:

```
🔗 Found pyo3 bindings
🐍 Found CPython 3.9 at /opt/python/cp39-cp39/bin/python
   Compiling pyo3 v0.13.2
   Compiling pyo3-mixed v2.1.1 (/io/test-crates/pyo3-mixed)
    Finished dev [unoptimized + debuginfo] target(s) in 6.46s
📦 Built wheel for CPython 3.9 to dist/pyo3_mixed-2.1.1-cp39-cp39-manylinux2010_x86_64.whl
📦 Wheel is eligible for a higher priority tag. You requested manylinux2010 but I have found this wheel is eligible for manylinux1
📦 Fixed-up wheel written to dist/pyo3_mixed-2.1.1-cp39-cp39-manylinux1_x86_64.whl
```


src/auditwheel/policy.rs
```diff
@@ -41,4 +79,31 @@ mod test {
...
+    }
+
+    #[test]
+    fn test_policy_from_name() {
+        use crate::Manylinux;
+
+        let tags = &[
+            Manylinux::Manylinux1,
+            Manylinux::Manylinux2010,
+            Manylinux::Manylinux2014,
```
messense, line 101: TODO: add manylinux_2_24

src/auditwheel/audit.rs
```diff
@@ -0,0 +1,286 @@
...
+        [lib] if is_libpython.is_match(lib) => {
+            Err(AuditWheelError::LinksLibPythonError(lib.clone()))
+        }
+        offenders => Err(AuditWheelError::ManylinuxValidationError(
+            offenders.to_vec(),
+        )),
+    }
+}
+
+/// An (incomplete) reimplementation of auditwheel, which checks elf files for
```
konstin, line 244:
```suggestion
/// A reimplementation of auditwheel, which checks elf files for
```

:tada:

src/build_context.rs
```diff
@@ -166,16 +166,45 @@ impl BuildContext {
...
-            .get_platform_tag(&self.manylinux, self.universal2);
+            let policy = auditwheel_rs(&artifact, target, manylinux)
+                .context(format!("Failed to ensure {} compliance", manylinux))?;
+            Ok(policy)
+        } else {
+            Ok(None)
+        }
+    }
+
+    #[cfg(not(feature = "auditwheel"))]
```
konstin, line 189: Now that manylinux has become such an essential feature, we could also drop that feature flag

src/build_context.rs
```diff
@@ -190,23 +219,96 @@ impl BuildContext {
...
+
+        if let Some(policy) = policy {
+            if policy > Policy::from_name(&self.manylinux.to_string()).unwrap() {
+                let manylinux = policy.manylinux_tag();
+                if self
+                    .auditwheel(python_interpreter, &artifact, &manylinux)
+                    .is_ok()
+                {
+                    println!(
+                        "📦 Wheel is eligible for a higher priority tag. You requested {} but I have found this wheel is eligible for {}",
```
konstin, line 265: This is great :+1:

konstin:
I think the ideal solution would be having three categories:

 * `auto` (default): Goes from the highest to the lowest priority, tags the first one that matches. If nothing matches, uses manylinux
 * `2010`/`2014`/`2_24`/etc.: Tag exactly this version. maturin should tell users about a higher priority tag, but tag the one requested.
 * `off`: Tags linux

There might be a case for a fourth option, `manylinux`, which tags the highest manylinux possible and fails if only linux would match.

This PR is already a huge improvement as-is, so I'm happy with merging this now and figuring out the specifics of the cli arguments later

messense:
> This PR is already a huge improvement as-is, so I'm happy with merging this now and figuring out the specifics of the cli arguments later

Sure, let's do that later in #448

messense:
> auto (default): Goes from the highest to the lowest priority, tags the first one that matches. If nothing matches, uses manylinux

Do you mean `If nothing matches, uses linux`?

konstin:
Oops, yes that should have been `linux`

PR #461, opened 2021-03-04
Pick best matching manylinux by default

konstin:
By default, the manylinux option is None and the highest matching policy will be picked. If the user picked a specific version, this one will be checked and tagged. If the wheel is eligible for a higher version tag than requested, the user gets a note.

I've turned off manylinux with develop and PEP517. Afaik every other PEP517 backend is also emitting `linux`, so I think this is the right thing to do.

I'm not sure if the error handling for the no-manylinux-matched case is good yet, e.g. for cases where the platform didn't match. I think the ideal solution would be checking whether the lowest manylinux failed because of the glibc version (ok), the architecture (warn) or something else (error).


src/auditwheel/manylinux.rs
```diff
@@ -0,0 +1,47 @@
+use serde::{Deserialize, Serialize};
+use std::fmt;
+use std::str::FromStr;
+
+/// Decides how to handle manylinux compliance
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq)]
```
messense, line 6: I was think about making this enum `Copy` since it's cheap to do.

konstin: Good idea

messense:
> If the wheel is eligible for a higher version tag than requested, the user gets a note

Python auditwheel will fix up the wheel automatically if it finds that the wheel is eligible for a higher version tag, having the behavior aligned with that would be nice IMO because for example you can also get manylinux1 wheel when you specified manylinux2010. 

It's quite useful since maturin doesn't allow passing `--manylinux 1`. Without this user needs to use the Python auditwheel to get manylinux1 wheel, which makes the reimplementation of auditwheel in maturin less useful.

messense:
I think maturin should failed if the user picked a specific version but architecture isn't supported.

konstin:
> > If the wheel is eligible for a higher version tag than requested, the user gets a note
> 
> Python auditwheel will fix up the wheel automatically if it finds that the wheel is eligible for a higher version tag, having the behavior aligned with that would be nice IMO because for example you can also get manylinux1 wheel when you specified manylinux2010.
> 
> It's quite useful since maturin doesn't allow passing `--manylinux 1`. Without this user needs to use the Python auditwheel to get manylinux1 wheel, which makes the reimplementation of auditwheel in maturin less useful.

I see your point, however I'd like to mark manylinux1 as completely unsupported. manylinux2010 requires glibc 2.12, which was released 11 years ago, and even centos isn't supported that long. It would also be weird to have a platform that you can install a wheel on, but on which you can't build or install a source distribution on.

> I think maturin should failed if the user picked a specific version but architecture isn't supported.

If the user picked a specific manylinux, than definitely error! I meant the case where someone runs plain `maturin build` on an architecture that's not yet considered by any manylinux spec.

PR #464, opened 2021-03-04
Add a test crate that link disallowed library

### LINKED ISSUE #459: Test crate that link disallowed library
maturin will detect and error if you try to build a library that links another library that is not explicitly allowed by manylinux. It would be good to add a library that links such a library and a test that checks that the error message is helpful.

messense:
Closes #459


konstin:
Thank you, that is a great test!

PR #469, opened 2021-03-07
Add basic support for OpenBSD

octeep:
This adds basic support for OpenBSD targets. I was able to compile python-adblock with this. 

However building on OpenBSD with `pip3 install` fails because of some linking error. 
Compiling with `cargo build --manifest-path Cargo.toml --bin maturin` works. 
Would it be okay to patch `setup.py` to use a `cargo build --manifest-path Cargo.toml --bin maturin` instead
when building on OpenBSD?

### DIFF
--- a/src/module_writer.rs
+++ b/src/module_writer.rs
@@ -582,7 +582,7 @@ pub fn write_bindings_module(
         Some(python_interpreter) => python_interpreter.get_library_name(&module_name),
         // abi3
         None => {
-            if target.is_freebsd() || target.is_unix() {
+            if target.is_freebsd() || target.is_unix() || target.is_openbsd() {
                 format!("{base}.abi3.so", base = module_name)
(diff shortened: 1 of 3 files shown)


messense:
https://builds.sr.ht/ provides OpenBSD CI service, maybe we can add it.

konstin:
Thanks for checking @messense, but I don't want to maintain ci for fringe platforms. 

I'll write up a policy on which platforms get tested/binary releases/supported. In general, I don't want to support platforms that have less than std+host tier 2 [support in rustc](https://doc.rust-lang.org/nightly/rustc/platform-support.html) beyond being able to set the right tags.

> Would it be okay to patch `setup.py` to use a `cargo build --manifest-path Cargo.toml --bin maturin` instead
> when building on OpenBSD?

I'd like to avoid that.

Issue #470, opened 2021-03-09
Issue during cross-compilation to aarch64


rmcgibbo:
I'm seeing an issue with `maturin` failing to auto-detect the python interpreter when cross-compiling to aarch64 from x86_64. The exact same code works fine when compiling on x86_64 for x86_64 or when compiling on aarch64 hardware.

I filed this issue previously with Nix, but I think it's a maturin problem, rather than a nix problem: https://github.com/NixOS/nixpkgs/issues/115601

The whole thing works if I pass `--interpreter $(which python)` to the `maturin` executable.

I've built a modified version of `maturin` to debug, and I can see that it's failing here: https://github.com/PyO3/maturin/blob/master/src/python_interpreter.rs#L490

When running with the following patch: https://gist.github.com/rmcgibbo/dcc82d7a72bee4445e4fc783c960b48d, I see

```
🔗 Found pyo3 bindings
"python3.5" -c "import json\nimport platform\nimport sys\nimport sysconfig\n\nmetadata = {\n    \"major\": sys.version_info.major,\n    \"minor\": sys.version_info.minor,\n    \"abiflags\": sysconfig.get_config_var(\"ABIFLAGS\"),\n    \"interpreter\": platform.python_implementation().lower(),\n    \"ext_suffix\": sysconfig.get_config_var(\"EXT_SUFFIX\"),\n    \"abi_tag\": (sysconfig.get_config_var(\"SOABI\") or \"-\").split(\"-\")[1] or None,\n    \"m\": sysconfig.get_config_var(\"WITH_PYMALLOC\") == 1,\n    \"u\": sysconfig.get_config_var(\"Py_UNICODE_SIZE\") == 4,\n    \"d\": sysconfig.get_config_var(\"Py_DEBUG\") == 1,\n    # This one isn\'t technically necessary, but still very useful for sanity checks\n    \"platform\": platform.system().lower(),\n    # We need this one for windows abi3 builds\n    \"base_prefix\": sys.base_prefix,\n}\n\nprint(json.dumps(metadata))\n" = Ok(
    Output {
        status: ExitStatus(
            ExitStatus(
                32512,
            ),
        ),
        stdout: "",
        stderr: "",
    },
)
Bailed here on 491
💥 maturin failed
  Caused by: The given list of python interpreters is invalid
  Caused by: python3.5 is not a valid python interpreter
  Caused by: Trying to get metadata from the python interpreter 'python3.5' failed
 ```
  
[...]

PR #471, opened 2021-03-10
Look for python interpreters in PATH before invoking them

rmcgibbo:
Fixes https://github.com/PyO3/maturin/issues/470.


rmcgibbo:
Here’s some deeper investigation of this problem: https://github.com/NixOS/nixpkgs/issues/115601#issuecomment-795070335

PR #473, opened 2021-03-17
Allow building mixed projects with __init__.pyi file

vinhowe:
This makes it possible to hand author a pyi stub file for the root module until we have something like https://github.com/PyO3/pyo3/issues/510.


vinhowe:
This doesn't seem to work like I hoped. I'll have to figure out what's actually going on.

vinhowe:
To anyone trying to do this, it looks like you need an `__init__.py` file after all. The `.so` file that Maturin puts in the directory next to it is just treated by Python as a submodule, so none of its members are available from the root module, obviously, unless you import them.

So `__init__.py` just looks like this:
```py
from .package_name import *
```

PR #477, opened 2021-03-26
Remove unused interpreter metadata message fields

ravenexp:
`Py_UNICODE_SIZE` is obsolete since CPython 3.3.
"m" and "d" are duplicated in "abi_tag".

No need to derive Serialize for the message struct.


konstin:
Good catch!

Issue #478, opened 2021-03-30
Strange behavior with PyPy in 0.10 prerelease


messense:
Please provide the following information: 
 * Your python version (`python -V`): PyPy 7.3.3 with Python (3.6.12)
 * Your pip version (`pip -V`): 
   * If this isn't the [latest pip version](https://pip.pypa.io/en/stable/news/), have you checked using the latest pip version?
 * The version of the bindings you're using, if any (e.g. `pyo3`, `rust-cpython` or `cffi`): pyo3
 * Does `cargo build` work? I think so
 * If on windows, have you checked that you aren't accidentally using unix path (those with the forward slash `/`)?

Please list the exact steps required to reproduce your error with all command output and if possible with a repository:

maturin-0.10.0b6 , https://github.com/milesgranger/pyrus-cramjam/runs/2218894957?check_suite_focus=true errored with

```bash
maturin build -i /opt/hostedtoolcache/PyPy/3.6.12/x64/bin/pypy --release --out dist
🔗 Found pyo3 bindings
🐍 Found PyPy 3.6 at /opt/hostedtoolcache/PyPy/3.6.12/x64/bin/pypy
📦 Built source distribution to dist/cramjam-2.1.0.tar.gz
warning: version requirement `0.6.0+zstd.1.4.8` for dependency `zstd` includes semver metadata which will be ignored, removing the metadata is recommended to avoid confusion
   Compiling cfg-if v1.0.0
   Compiling scopeguard v1.1.0
error[E0463]: can't find crate for `core`
  |
  = note: the `x86_64-unknown-linux-musl` target may not be installed


error: aborting due to previous error


error[E0463]: can't find crate for `core`
  |
  = note: the `x86_64-unknown-linux-musl` target may not be installed


error: aborting due to previous error


error: could not compile `scopeguard`

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error: build failed
💥 maturin failed
  Caused by: Failed to build a native library through cargo
  Caused by: Cargo build finished with "exit code: 101": `cargo rustc --message-format json --manifest-path Cargo.toml --release --target x86_64-unknown-linux-musl --lib --`
```

It's strange that `--target x86_64-unknown-linux-musl` is passed to `cargo rustc`, it works fine with maturin 0.9.4

messense:
```
Run make pypy-build
printenv
POWERSHELL_DISTRIBUTION_CHANNEL=GitHub-Actions-ubuntu20
[...]

PR #486, opened 2021-04-04
Guessed platform may be incorrect, so do not force compilation with it

al626:
See https://github.com/PyO3/maturin/issues/478

This reverts most of https://github.com/PyO3/maturin/pull/429 while hopefully keeping the issue described in the PR fixed.

Pretty new to rust so feel free to close & fix otherwise if this isn't suitable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ __pycache__/
 dist/
 build
 dist
+tags
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -22,12 +22,7 @@ pub fn compile(
     if context.target.is_macos() && context.universal2 {
         compile_universal2(context, python_interpreter, bindings_crate)
     } else {
-        compile_target(
-            context,
-            python_interpreter,
-            bindings_crate,
-            context.target.target_triple(),
-        )
+        compile_target(context, python_interpreter, bindings_crate, None)
     }
 }
 
@@ -45,7 +40,7 @@ fn compile_universal2(
         context,
         python_interpreter,
         bindings_crate,
-        "aarch64-apple-darwin",
+        Some("aarch64-apple-darwin"),
     )
     .context("Failed to build a aarch64 library through cargo")?
     .get(build_type)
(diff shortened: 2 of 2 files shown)


messense:
Thanks, I think we should remove the usage of `Platform::guess_current()`.

messense (approved):

messense:
Rethinking about this, we should probably avoid using `Platform::guess_current()` to set [`Target::env`](https://github.com/PyO3/maturin/blob/1239c4f7ffdd6c1d769405c6d51e95a173ae72a7/src/target.rs#L113) since the maturin linux wheel is built with musl libc, the `target_env` of `platforms` would always be `musl`?

al626:
This is slightly confusing; there are two separate concepts:

1. Building maturin
2. Using maturin to build something else.

My understanding of the situation is for (1), it's important to have `Platform::guess_current()` as otherwise tools like `cargo build` and `cargo test` would not work; a target would need to be explicitly specified (e.g. `cargo build --target x86_64-unknown-linux-gnu` ). Here, `Platform::guess_current()` works as `Platform` is compiled from source in the maturin build.

For (2), a precompiled version of maturin is used, so `Platform::guess_current()` will not return the correct `env`, but does return the correct `arch` and `os` (as pip wheels are compiled for a particular `os` and `arch` but not `env`). `os` info is necessary (e.g. to differentiate between windows and linux).

My opinion is that if we were doing things properly, we should strip out uses of `Target` (which wraps `Platform::guess_current()`) for everything involved in using maturin to build other things (e.g. `develop.rs`), and replace with something a bit more specific that only stores and uses `os` info (e.g. `std::env::consts::OS`). However, this is a fairly large change and may be overkill.

We should also write a test that ensures that a maturin compiled for `musl` can build something for `gnu` without the `musl` toolchain installed.

What do you think about the above? I can do this, but will be slow (time limitations + new to rust).

Thanks.

messense:
Merged in #487, thanks!

PR #487, opened 2021-04-05
Read host target triple from `rustc -vV`

### LINKED ISSUE #478: Strange behavior with PyPy in 0.10 prerelease

messense:
Fixes #478, closes #486

cc @al626

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,6 +43,9 @@ jobs:
       - name: Install x86_64-unknown-linux-musl Rust target
         if: matrix.os == 'ubuntu-latest'
         run: rustup target add x86_64-unknown-linux-musl
+      - name: Install musl tools
+        if: matrix.os == 'ubuntu-latest'
+        run: sudo apt-get install -y musl-tools
       #- name: Cache cargo registry
       #  uses: actions/cache@v2
       #  with:
@@ -72,6 +75,12 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
+      - name: cargo test with musl
+        if: matrix.os == 'ubuntu-latest'
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --target x86_64-unknown-linux-musl
 
   test-auditwheel:
     name: Test Auditwheel
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -76,7 +76,7 @@ jobs:
 
       - name: Install musl tools
         if: matrix.os == 'ubuntu-latest'
-        run: sudo apt install -y musl-tools
+        run: sudo apt-get install -y musl-tools
 
       # Those two will also create target/${{ matrix.target }}/maturin
       - name: Publish to pypi (with sdist)
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ __pycache__/
 dist/
 build
 dist
+tags
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -22,12 +22,7 @@ pub fn compile(
     if context.target.is_macos() && context.universal2 {
         compile_universal2(context, python_interpreter, bindings_crate)
     } else {
-        compile_target(
-            context,
-            python_interpreter,
-            bindings_crate,
-            context.target.target_triple(),
-        )
+        compile_target(context, python_interpreter, bindings_crate, None)
     }
 }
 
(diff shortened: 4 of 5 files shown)


konstin:
I think the best solution for both `maturin build --target xxx --universal2` from #429 and the musl handling from #478 would be reverting #429 and manually failing on the `--target xxx --universal2` case. I.e. I'd prefer #486 and add an extra error path. This also has the advantage that we save a subprocess call.

If there are other reasons I've missing and we still need to go for reading the target triple, I'd do [what cargo does](https://github.com/rust-lang/cargo/blob/1dae5acb7d27568ca30fe2da49b4a447d7b78e90/src/cargo/util/rustc.rs#L50-L68) and read `rustc -vV`:

```rust
use std::process::Command;

use anyhow::{format_err, Context, Result};
use std::str;

fn get_target() -> Result<String> {
    let output = Command::new("rustc")
        .arg("-vV")
        .output()
        .context("Failed to run rustc to get the host target")?;
    let output = str::from_utf8(&output.stdout).context("`rustc -vV` didn't return utf8 output")?;

    let field = "host: ";
    let host = output
        .lines()
        .find(|l| l.starts_with(field))
        .map(|l| &l[field.len()..])
        .ok_or_else(|| {
            format_err!(
                "`rustc -vV` didn't have a line for `{}`, got:\n{}",
                field.trim(),
                output
            )
        })?
        .to_string();
    Ok(host)
}

fn main() -> Result<()> {
    let host = get_target()?;
    println!("target triple: {}", host);
    Ok(())
}
```

messense:
There is also https://github.com/PyO3/maturin/pull/423 affected by target_env

messense:
Anyway I think it's good to get rid of `Platform::guess_current()`.

.github/workflows/test.yml
```diff
@@ -72,6 +75,12 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
+      - name: cargo test with musl
+        if: matrix.os == 'ubuntu-latest'
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --target x86_64-unknown-linux-musl
```
konstin, line 83: Since we're only shipping musl anyway, wouldn't it make more sense to only run the tests for musl?

konstin (approved):
Looks good, thank you!

.github/workflows/test.yml
messense: Just in case someone needs to compile maturin from source instead of using prebuilt wheels, for example on some BSD systems?

PR #491, opened 2021-04-08
Add platform support draft

konstin:
I tried to strike a balance between supporting reasonably much and not being burdened by supporting fringe platforms

CC @messense

### DIFF
--- a/Platform Support.md
+++ b/Platform Support.md
@@ -6,8 +6,8 @@ Being built on cargo and rustc, maturin is limited by [rust's platform support](
 
 **Releases**: The following targets are built into wheels and downloadable binaries:
 
- * windows: 32-bit and 64-bit x86 as well armv7 and aarch64
- * linux 32-bit and 64-bit x86 (musl)
+ * windows: 32-bit and 64-bit x86
+ * linux 32-bit and 64-bit x86 as well as armv7 and aarch64 (musl)
  * mac: 64-bit and aarch64
 
 **Other Operating Systems**: It should be possible to build maturin and for maturin to build wheels on other platforms supported by rust. To add a new os, add it in target.rs and, if it doesn't behave like the other unixes, in `PythonInterpreter::get_tag`. Please also submit the output of `python -m sysconfig` as a file in the `sysconfig` folder. It's ok to edit setup.py to deactivate default features so `pip install` works, but new platforms should not require complex workaround in compile.rs.
--- a/Platform Support.md
+++ b/Platform Support.md
@@ -1,0 +1,17 @@
+# Platform Support
+
+Being built on cargo and rustc, maturin is limited by [rust's platform support](https://doc.rust-lang.org/nightly/rustc/platform-support.html).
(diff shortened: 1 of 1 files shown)


Platform Support.md
```diff
@@ -0,0 +1,17 @@
+# Platform Support
+
+Being built on cargo and rustc, maturin is limited by [rust's platform support](https://doc.rust-lang.org/nightly/rustc/platform-support.html).
+
+**Automated tests**: On github actions, windows, mac and linux are tested, all on 64-bit x86. FreeBSD is also tested though circle ci, but might get removed at some point. Since CI is very time intensive to maintain, I'd like to stick to github action and these three platforms.
```
messense, line 5: Nit:

```suggestion
**Automated tests**: On github actions, windows, macOS and linux are tested, all on 64-bit x86. FreeBSD is also tested though circle ci, but might get removed at some point. Since CI is very time intensive to maintain, I'd like to stick to github action and these three platforms.
```

Platform Support.md
```diff
@@ -0,0 +1,17 @@
...
+
+**Releases**: The following targets are built into wheels and downloadable binaries:
+
+ * windows: 32-bit and 64-bit x86 as well armv7 and aarch64
```
messense, line 9: I don't think we have armv7 and aarch64 binaries for windows? They are for Linux.

Platform Support.md
```diff
@@ -0,0 +1,17 @@
...
+ * linux 32-bit and 64-bit x86 (musl)
+ * mac: 64-bit and aarch64
```
messense, line 11:
```suggestion
 * macOS: 64-bit and aarch64
```

Platform Support.md
```diff
@@ -0,0 +1,17 @@
...
+
+**Other Operating Systems**: It should be possible to build maturin and for maturin to build wheels on other platforms supported by rust. To add a new os, add it in target.rs and, if it doesn't behave like the other unixes, in `PythonInterpreter::get_tag`. Please also submit the output of `python -m sysconfig` as a file in the `sysconfig` folder. It's ok to edit setup.py to deactivate default features so `pip install` works, but new platforms should not require complex workaround in compile.rs.
```
messense, line 13:
```suggestion
**Other Operating Systems**: It should be possible to build maturin and for maturin to build wheels on other platforms supported by rust. To add a new os, add it in target.rs and, if it doesn't behave like the other unixes, in `PythonInterpreter::get_tag`. Please also submit the output of `python -m sysconfig` as a file in the `sysconfig` folder. It's ok to edit setup.py to deactivate default features so `pip install` works, but new platforms should not require complex workaround in `compile.rs`.
```

Platform Support.md
konstin: Yeah that was a typeo

messense: This still says `armv7 and aarch64` for windows.

messense (approved):

PR #493, opened 2021-04-09
Try to workaround bsd tar bug in github actions on mac

konstin:


messense (approved):

.github/workflows/test.yml
```diff
@@ -57,19 +57,26 @@ jobs:
           echo "SDKROOT=$(xcrun --sdk macosx --show-sdk-path)" >> "${GITHUB_ENV}"
 
       # Caching
+      # Install gnu-tar because BSD tar is buggy
+      # https://github.com/actions/cache/issues/403
+      - name: Install GNU tar (Macos)
+        if: matrix.build == 'macos'
```
messense, line 63: Oops, I think this should be `if: matrix.os == 'macos-latest'`

konstin: Ah thanks

PR #495, opened 2021-04-09
Consider requires-python when searching for interpreters

### LINKED ISSUE #195: Feature request: Do not build wheels for python versions not supported by package
My computer has more versions of python installed on it than my pyo3-based library supports. `maturin build` should not build wheels for unsupported python versions. Of course, it's possible to specify exactly which versions to build on every call to `maturin build`, but that's a relatively static item of metadata which is a good fit for a config file (probably the `Cargo.toml`, although possibly `pyproject.toml`).

One option would be to parse ~~the trove classifiers specified in `package.metadata.maturin.classifier`~~ `package.metadata.maturin.requires-python`, and build wheels for the intersection of what maturin supports and what is specified there.

konstin:
Fixes #195

Not sure if there's any better way to test this.

### DIFF
--- a/src/build_options.rs
+++ b/src/build_options.rs
@@ -8,6 +8,7 @@ use crate::PythonInterpreter;
 use crate::Target;
 use anyhow::{bail, format_err, Context, Result};
 use cargo_metadata::{Metadata, MetadataCommand, Node};
+use regex::Regex;
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, HashSet};
 use std::env;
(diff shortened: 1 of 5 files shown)


src/build_options.rs
```diff
@@ -176,9 +177,9 @@ impl BuildOptions {
             // Only build a source distribution
             Some(ref interpreter) if interpreter.is_empty() => vec![],
             // User given list of interpreters
-            Some(interpreter) => find_interpreter(&bridge, &interpreter, &target)?,
+            Some(interpreter) => find_interpreter(&bridge, &interpreter, &target, None)?,
```
messense, line 180: Should we also filter the user given list of interpreters based if `requires-python` specified?

messense (approved):

src/build_options.rs
konstin: No, in that case I'd trust the users list more than the PEP 440 heuristic

messense:
The error message isn't good when maturin can't find any interpreters with `requries-python` set.

```bash
🔗 Found pyo3 bindings
💥 maturin failed
  Caused by: Couldn't find any python interpreters. Please specify at least one with -i
```

It's kinda confusing if user already has some other Python versions installed. It would be much nicer if it says `Couldn't find any python interpreters version >= 3.7. Please specify at least one with -i`.

PR #497, opened 2021-04-11
Only run pyo3_no_extension_module test on glibc Linux

messense:
This test case relies on auditwheel which is not supported on musl libc based Linux distributions for example alpine.

cc @omnivagant


messense:
Closing per https://github.com/PyO3/maturin/commit/12ac30f19358c2429aa60e1507c987c4292c54fa

Issue #490, opened 2021-04-07
Importing mixed Rust/Python project fails when structured as recommended


ArniDagur:
The README of this project says the following:
> 
> ## Mixed rust/python projects
> 
> To create a mixed rust/python project, create a folder with your module name (i.e. `lib.name` in Cargo.toml) next to your Cargo.toml and add your python sources there:
> 
> ```
> my-project
> ├── Cargo.toml
> ├── my_project
> │   ├── __init__.py
> │   └── bar.py
> ├── Readme.md
> └── src
>     └── lib.rs
> ```
> 
> maturin will add the native extension as a module in your python folder. When using develop, maturin will copy the native library and for cffi also the glue code to your python folder. You should add those files to your gitignore.
> 
> With cffi you can do `from .my_project import lib` and then use `lib.my_native_function`, with pyo3/rust-cpython you can directly `from .my_project import my_native_function`.
> 
> Example layout with pyo3 after `maturin develop`:
> 
> ```
> my-project
> ├── Cargo.toml
> ├── my_project
> │   ├── __init__.py
> │   ├── bar.py
> │   └── my_project.cpython-36m-x86_64-linux-gnu.so
> ├── Readme.md
> └── src
>     └── lib.rs
> ```
> 
This is the structure I'm using for [`python-adblock`](https://github.com/ArniDagur/python-adblock). The project is written entirely in Rust, but I'm using a mixed Rust/Python project in order to include Python type stubs. I have an `__init__.py` which imports the neccesary symbols from the native module.

However, some of my users have reported that the imports inside `__init__.py` fail as follows:
```
>>> import adblock
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/app/adblock/__init__.py", line 1, in <module>
    from .adblock import __version__, Engine, FilterSet, BlockerResult, UrlSpecificResources
ModuleNotFoundError: No module named 'adblock.adblock'
```
or
```
>>> import adblock
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/app/adblock/__init__.py", line 1, in <module>
    from adblock.adblock import __version__, Engine, FilterSet, BlockerResult, UrlSpecificResources
ModuleNotFoundError: No module named 'adblock.adblock'
```
[...]

PR #498, opened 2021-04-12
Always use EXT_SUFFIX from sysconfig

messense:
Fixes some issues on alpine linux reported in #490


konstin:
I thought this wouldn't possible due to abi3, this actually really smart since with abi3 we also know the extension is `.abi3.so` :+1:

Thanks!

messense:
I just realized that this actually made cross compiling support worser. 😅 I need to finish https://github.com/PyO3/maturin/pull/454

PR #507, opened 2021-04-22
Compile deps in separate docker layer for caching

konstin:


messense:
I wonder why we  don't use GitHub Actions cache? https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#github-cache

konstin:
No strong reason, it's just that the v2 of docker/build-push-action is cumbersome to use with git tags and docker cache should be good enough

Issue #481, opened 2021-03-31
Cannot "pip install maturin" on openbsd 6.8


toyg:
Please provide the following information: 
 * Your python version (`python -V`): 
 tested with 3.8.6 (system) and 3.9.2 (built by pyenv)
 * Your pip version (`pip -V`):
 21.0.1
 * Does `cargo build` work?
 `cargo install maturin` works fine
 
Please list the exact steps required to reproduce your error with all command output and if possible with a repository:

- install OpenBSD 6.8 (I used the Vagrant image generic/openbsd6 )
- install the minimal required packages python3 py3-pip libffi sqlite3 (with `doas pkg_add ...` or `sudo pkg_add ...`)
- create virtualenv: `python3 -m venv venv && source ./venv/bin/activate && pip install --upgrade pip`
- `pip install maturin`

The output is available here: https://pastebin.com/MhPm0ktH

It looks like all dependencies get compiled correctly, but then it fails to compile maturin. However, a generic `cargo install maturin` on the same machine works fine, so the issue must be a difference between the way it's compiled there and in the wheel. I tried using virtualenvs for 3.9 and 3.8 as described above, and on system too (`sudo pip3.8 install maturin` )

I can't seem to find a way to save the actual compilation log (pip deletes the temporary directory on failure), if you know it just let me know and I'll produce the tmp files.

For the record, I am not going to use maturin directly; it's a dependency of something else that I need, so I may not be familiar with all the gotchas of maturin.

messense:
Try `pip install -vv maturin` to get a verbose error output.

toyg:
```
Compiling maturin v0.10.0 (/tmp/pip-install-z5xe1g4r/maturin)
      Finished dev [unoptimized + debuginfo] target(s) in 7m 02s
[...]

PR #514, opened 2021-04-25
Show `cargo rustc` output when `pip install maturin` from sdist failed

messense:
It's hard to debug #481 without rustc's output.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -69,7 +69,12 @@ class PostInstallCommand(install):
 
             cargo_args.extend(["--", "-C", "link-arg=-s"])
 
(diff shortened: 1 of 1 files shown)


konstin:
Good idea!

messense:
I just noticed that maturin [isn't compiled with `--release` in `setup.py`](https://github.com/PyO3/maturin/blob/9f5afe0f5e259903a8b70b1428297593331d5e45/setup.py#L63), why not? @konstin

konstin:
I totally missed that, feel free to add `--release`

Issue #505, opened 2021-04-16
Building maturin fails on FreeBSD 12.2


flz:
I'm going to assume this is a clang-specific issue, compile fails because of link-arg="-s". Using link-arg= completes successfully:

```bash
# cargo rustc --bin maturin --message-format=json -- -C link-arg=-s
[...] /usr/bin/ld+0x2b9e7f0)\ncc: error: unable to execute command: Segmentation fault\ncc: error: linker command failed due to signal (use -v to see invocation)\n","rendered":null,"spans":[]}],"code":null,"level":"error","message":"linking with `cc` failed: exit code: 254","spans":[]}}
{"reason":"compiler-message","package_id":"maturin 0.10.3 (path+file:///tmp/pip-install-agt07kqt/maturin_71e6430055d84773a6851e18d5a2a46a)","target":{"kind":["bin"],"crate_types":["bin"],"name":"maturin","src_path":"/tmp/pip-install-agt07kqt/maturin_71e6430055d84773a6851e18d5a2a46a/src/main.rs","edition":"2018","doc":true,"doctest":false,"test":true},"message":{"rendered":"error: aborting due to previous error\n\n","children":[],"code":null,"level":"error","message":"aborting due to previous error","spans":[]}}
{"reason":"build-finished","success":false}
error: could not compile `maturin`

To learn more, run the command again with --verbose.
```

Some system information:

```bash
# clang -v
FreeBSD clang version 10.0.1 (git@github.com:llvm/llvm-project.git llvmorg-10.0.1-0-gef32c611aa2)
Target: x86_64-unknown-freebsd12.2
Thread model: posix
InstalledDir: /usr/bin
# python3.7 -V
Python 3.7.10
# pip3.7 -V
pip 21.0.1 from /usr/local/lib/python3.7/site-packages/pip (python 3.7)
# rustc -V
rustc 1.51.0
```

konstin:
If the linker segfaults this is a bug in the toolchain and I'm afraid there's nothing maturin can do about this. Note that this would be different if the linker complained about the flag being unsupported rather than crashing. Building maturin without stripping is also not really feasible because the binaries become huge (167MB vs. 14.6MB)

messense:
@flz Can you try `pip install https://github.com/messense/maturin/archive/refs/heads/setup-no-strip.zip` ?

PR #516, opened 2021-04-26
Remove `link-arg=-s` from `setup.py`

messense:
Fixes https://github.com/PyO3/maturin/issues/505 , see https://github.com/PyO3/maturin/issues/481#issuecomment-826876101


konstin:
Not overly happy with that, but I guess it's ok for now

PR #517, opened 2021-04-27
Try to fix FreeBSD CI

### LINKED ISSUE #512: FreeBSD error when running pip
From the Cirrus CI run https://github.com/PyO3/maturin/runs/2427736854: 

```
ld-elf.so.1: /usr/local/lib/libpython3.7m.so.1.0: Undefined symbol "close_range@FBSD_1.6"
```
It seams that this is only a warning but it looks fishy. Might be related to https://github.com/python-poetry/poetry/issues/3669

messense:
See https://github.com/duckinator/bork/issues/249#issuecomment-778841570 , fixes #512

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,5 +1,5 @@
 freebsd_instance:
-  image_family: freebsd-12-1
+  image_family: freebsd-12-2
 
 env:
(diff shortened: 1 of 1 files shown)


konstin:
Great find, thanks!

PR #522, opened 2021-04-29
Make it clearer what the command does

itamarst:
I just copied this without thinking too hard, and ended up with debug-compiled wheels, so try to clarify what's going on.


messense (approved):
Thanks

PR #524, opened 2021-05-05
Add support for Windows arm64 target

messense:
[`aarch64-pc-windows-msvc`](https://rust-lang.github.io/rustup-components-history/aarch64-pc-windows-msvc.html) is a tier 2 Rust target.


konstin:
Thanks!

Issue #523, opened 2021-05-05
auditwheel: generate single wheel for multiple eligible manylinux tags


messense:
1. https://github.com/pypa/auditwheel/pull/289
2. https://github.com/joerick/cibuildwheel/pull/633#issue-608399812

This will make it possible to support `manylinux1` wheel with `--manylinux 2010` and later even if Rust doesn't run on CentOS 5.

konstin:
What's the advantage of tagging multiple manylinux versions instead of just tagging the oldest one possible?

messense:
> What's the advantage of tagging multiple manylinux versions instead of just tagging the oldest one possible?

I guess if we just tagging the oldest one there is no advantage. But if `--manylinux $VERSION` is passed, currently maturin only generate the specified tag which probably isn't the oldest tag. It's nice to have maturin's behavior aligned with the official auditwheel package.

konstin:
TBH I wrote maturin because I disagree with the behaviour of the official tools. They are often counterintuitive and lead to projects badly broken in subtle ways, which regularly causes problems in across different projects. I wanted maturin to be something that might support less things, but does them correctly.

For `--manylinux $VERSION`, I believe that we should only tag `$VERSION`. Ultimately, maintainers should control what the lowest supported version is and maturin should not arbitrarily overrule that. For those who just want the defaults, I recommend the manylinux 2_17 image with `--manylinux 2_17` due to arm support.

messense:
I understand your concern. It's just not clear to me what kind of problem will this feature cause in practice, can you give some examples?

konstin:
I don't have any concrete examples, but from my experience software trying to be smarter than what the user explicitly requested always leads to problems eventually. With default options maturin is already tagging the lowest possible version and we also have the warning saying when a lower version would be possible, so imho this would only be a small gain.

Sorry for rejecting this, I know this seems like an easy win but I think we're better off with just tagging the requested version.

PR #525, opened 2021-05-06
Always output PEP 600 platform tags

messense:
Part of #523 

https://www.python.org/dev/peps/pep-0425/#compressed-tag-sets

### DIFF
--- a/src/target.rs
+++ b/src/target.rs
@@ -158,7 +158,13 @@ impl Target {
                 let release = info.release().replace(".", "_").replace("-", "_");
                 format!("openbsd_{}_amd64", release)
             }
-            (Os::Linux, _) => format!("{}_{}", manylinux, self.arch),
+            (Os::Linux, _) => {
+                let mut tags = vec![format!("{}_{}", manylinux, self.arch)];
+                for alias in manylinux.aliases() {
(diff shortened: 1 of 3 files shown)


src/target.rs
```diff
@@ -158,17 +158,23 @@ impl Target {
...
+                let mut tags = vec![format!("{}_{}", manylinux, self.arch)];
+                for alias in manylinux.aliases() {
+                    tags.push(format!("{}_{}", alias, self.arch));
+                }
+                tags.join(".")
+            }
             (Os::Macos, Arch::X86_64) => {
                 if universal2 {
-                    "macosx_10_9_universal2".to_string()
+                    "macosx_10_9_x86_64.macosx_11_0_arm64.macosx_10_9_universal2".to_string()
```
konstin, line 170: Same as https://github.com/PyO3/maturin/issues/523#issuecomment-833517235: What extra information do we get through this that we don't get through `macosx_10_9_universal2`?

messense: Old version of `pip` which doesn't know about `universal2` can still pick up this wheel.

konstin: I'd prefer not to support old versions of pip and instead tell people to update their pip, given that the pip team doesn't support old versions and they print a yellow warning after every command run from an outdated version

PR #528, opened 2021-05-06
Fix missing `PyInit_<module_name>` warning with Rust submodule

messense:
Fixes an oversight of #489 , it doesn't cause any runtime issue but the warning is annoying.

```
⚠  Warning: Couldn't find the symbol `PyInit_pyo3_mixed_submodule` in the native library. Python will fail to import this module. If you're using pyo3, check that `#[pymodule]` uses `pyo3_mixed_submodule` as module name
📦 Built wheel for CPython 3.8 to /io/target/wheels/pyo3_mixed_submodule-2.1.3-cp38-cp38-manylinux_2_5_x86_64.manylinux1_x86_64.whl
```

https://github.com/PyO3/maturin/runs/2518736555


src/compile.rs
```diff
@@ -305,6 +305,24 @@ pub fn warn_missing_py_init(artifact: &Path, module_name: &str) -> Result<()> {
...
+                }
+                goblin::mach::Mach::Fat(_) => {
+                    // Ignore fat macho,
+                    // we only generate them by combining thin binaries which is handled above
+                    found = true
+                }
+            }
+        }
         _ => {
             // Currently, only linux is implemented
```
konstin, line 327:
```suggestion
            // Currently, only linux and mac are implemented
```

konstin: Also `/// Currently the check is only run on linux` -> `/// Currently checks linux and mac` up where I can't place suggestions

konstin (approved):
Thank you!

PR #533, opened 2021-05-08
Publish maturin wheel for s390x

messense:
[ring doesn't build on s390x yet](https://github.com/briansmith/ring/issues/986), so I've disabled the `rustls` feature.

Test run: https://github.com/messense/maturin/runs/2532708951?check_suite_focus=true

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -167,3 +167,41 @@ jobs:
           file: target/release/maturin.tar.gz
           tag: ${{ github.ref }}
           prerelease: ${{ contains(github.ref, 'alpha') || contains(github.ref, 'beta') }}
+
+  release-gnu-pypi:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        platform: [
+          { target: "s390x-unknown-linux-gnu", image: "messense/manylinux2014-cross:s390x", manylinux: "2014" },
+        ]
+    container:
+      image: docker://${{ matrix.platform.image }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+          target: ${{ matrix.platform.target }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -167,3 +167,40 @@ jobs:
           file: target/release/maturin.tar.gz
           tag: ${{ github.ref }}
           prerelease: ${{ contains(github.ref, 'alpha') || contains(github.ref, 'beta') }}
+
+  release-gnu-pypi:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        platform: [
+          { target: "s390x-unknown-linux-gnu", image: "messense/manylinux2014-cross:s390x", manylinux: "2014" },
```
messense, line 176: Rust doesn't currently distribute `s390x-unknown-linux-musl` target stdlib, so here we use GNU target instead.

.github/workflows/release.yml
```diff
@@ -167,3 +167,40 @@ jobs:
...
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+          target: ${{ matrix.platform.target }}
+      - name: Build and publish wheel
+        env:
+          MATURIN_PASSWORD: ${{ secrets.MATURIN_PASSWORD }}
+        run: |
+          sudo python3 -m pip install maturin==0.10.5_beta.4
```
messense, line 192: `s390x` support in maturin is only usable starting at `v0.10.5_beta.4`: #530 , #532

.github/workflows/release.yml
```diff
@@ -167,3 +167,41 @@ jobs:
...
+          maturin publish -u __token__ -b bin --no-sdist -o dist \
```
konstin, line 193: Do we need the `-o dist` here?

konstin (approved):
Thanks!

.github/workflows/release.yml
messense: No, it's a copypasta problem.

PR #536, opened 2021-05-09
Only use extension name from extra metadata if it contains dot

messense:
Previously the following example would work because maturin will use `lib.name` instead of `package.metadata.maturin.name`

```toml
[package]
name = "py-abc-def"

[lib]
name = "abc_def"
crate-type = ["cdylib"]

[package.metadata.maturin]
name = "abc-def"
```

With #489 it's broken ([example](https://github.com/messense/fat-macho-rs/actions/runs/824473631)) and it will use `package.metadata.maturin.name` for the extension filename which results in `ImportError` at runtime.

This PR restore the previous behavior.

### DIFF
--- a/src/build_options.rs
+++ b/src/build_options.rs
@@ -138,7 +138,13 @@ impl BuildOptions {
             .and_then(|lib| lib.name.as_ref())
             .unwrap_or(&crate_name)
             .to_owned();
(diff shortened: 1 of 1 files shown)


konstin:
LGTM

PR #537, opened 2021-05-09
Merge target specific RUSTFLAGS

messense:
Useful for supporting target like `powerpc64le-unknown-linux-musl` which doesn't have std library binary distribution yet but can be built with `cargo build -Zbuild-std` on nightly.

See also https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags

Before this patch build failed cause of missing some `-L` flags which are set by `CARGO_TARGET_XXX_RUSTFLAGS` env var: https://github.com/messense/maturin/runs/2537804583?check_suite_focus=true
After: https://github.com/messense/maturin/runs/2537949957?check_suite_focus=true

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -119,6 +119,17 @@ fn compile_target(
         .collect();
 
     let mut rust_flags = env::var_os("RUSTFLAGS").unwrap_or_default();
+    // Merge target specific RUSTFLAGS
+    let target_uppercase = context
+        .target
+        .target_triple()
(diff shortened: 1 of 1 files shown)


messense:
I guess this could be work around with build the x86_64-unknown-linux-gnu maturin first, then set `RUSTFLAGS` to build for other targets.

And after #538 gets released, the `build the x86_64-unknown-linux-gnu maturin first` step is not needed. So closing.

PR #538, opened 2021-05-09
Replace 'platforms' crate with `target-lexicon` crate

messense:
[target-lexicon](https://github.com/bytecodealliance/target-lexicon) is much more flexible and doesn't rely on hard-coded target triple list.


messense:
@omnivagant I think you can drop the target patch on alpine now.

PR #539, opened 2021-05-12
Publish maturin wheel for ppc64le

messense:
Test run: https://github.com/messense/maturin/runs/2563195788?check_suite_focus=true


konstin:
Thanks!

PR #543, opened 2021-05-19
Add PEP 656 musllinux support

### LINKED ISSUE #542: Add PEP 656 support: Platform Tag for Linux Distributions Using Musl
[PEP 656](https://www.python.org/dev/peps/pep-0656/) supports building wheels for musl. maturin should be extended to detect the  musl version instead of the glibc version and write wheels with the appropriate tag.

messense:
Closes #542

### DIFF
--- a/src/auditwheel/audit.rs
+++ b/src/auditwheel/audit.rs
@@ -260,6 +260,7 @@ pub fn auditwheel_rs(
     target: &Target,
     manylinux: Option<Manylinux>,
 ) -> Result<Policy, AuditWheelError> {
+    // TODO: add support for musllinux: https://github.com/pypa/auditwheel/issues/305
     if !target.is_linux() || manylinux == Some(Manylinux::Off) {
         return Ok(Policy::default());
     }
--- a/src/auditwheel/audit.rs
+++ b/src/auditwheel/audit.rs
@@ -257,9 +257,20 @@ pub fn auditwheel_rs(
     target: &Target,
     platform_tag: Option<PlatformTag>,
 ) -> Result<Policy, AuditWheelError> {
-    // TODO: add support for musllinux: https://github.com/pypa/auditwheel/issues/305
     if !target.is_linux() || platform_tag == Some(PlatformTag::Linux) {
         return Ok(Policy::default());
+    }
+    if let Some(musl_tag @ PlatformTag::Musllinux { .. }) = platform_tag {
+        // TODO: add support for musllinux: https://github.com/pypa/auditwheel/issues/305
+        eprintln!("⚠  Warning: no auditwheel support for musllinux yet");
+        // HACK: fake a musllinux policy
+        return Ok(Policy {
+            name: musl_tag.to_string(),
+            aliases: Vec::new(),
+            priority: 0,
+            symbol_versions: Default::default(),
+            lib_whitelist: Default::default(),
+        });
     }
     let arch = target.target_arch().to_string();
     let mut file = File::open(path).map_err(AuditWheelError::IoError)?;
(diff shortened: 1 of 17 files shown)


src/auditwheel/manylinux.rs
```diff
@@ -13,15 +13,18 @@ pub enum Manylinux {
...
     },
+    /// Use the musllinux_x_y tag
+    Musllinux {
+        /// musl libc version major
+        x: u16,
+        /// musl libc version minor
+        y: u16,
+    },
     /// Use the native linux tag
     Off,
```
messense, line 24: And rename this to `Linux`

src/auditwheel/manylinux.rs
```diff
@@ -3,7 +3,7 @@ use serde::{Deserialize, Serialize};
 use std::fmt;
 use std::str::FromStr;
 
-/// Decides how to handle manylinux compliance
+/// Decides how to handle manylinux and musllinux compliance
 #[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq, Copy)]
 pub enum Manylinux {
```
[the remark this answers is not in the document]
messense: The `--manylinux` command line option also needs some tweaking.

messense:
FreeBSD rustup issue: https://github.com/rust-lang/rustup/issues/2629

src/auditwheel/manylinux.rs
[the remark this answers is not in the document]
konstin: Difficult question. I worry that having both `--target` and `--platform` will be confusing to people, esp. since platform gets. Maybe something like `--compatibility`? `--tag` would be another possibility, but "tag" does sound more like metadata (e.g. docker tag) than like libc flavor and version.

src/auditwheel/manylinux.rs
konstin: :+1:

konstin (approved):
Already looks good!

src/build_options.rs
```diff
@@ -31,11 +31,12 @@ pub struct BuildOptions {
     ///
     /// This option is ignored on all non-linux platforms
     #[structopt(
-        long,
-        possible_values = &["2010", "2014", "2_12", "2_17", "2_24", "2_27", "off"],
-        case_insensitive = true,
+        name = "compatibility",
+        long = "compatibility",
+        alias = "manylinux",
```
messense, line 36: Let's keep the `--manylinux` as an alias to `--compatibility` for a while. It'd be nice to somehow emit a warning if `--manylinux` is used instead of `--compatibility`.

src/auditwheel/audit.rs
```diff
@@ -238,29 +238,30 @@ fn policy_is_satisfied(
...
-/// Does nothing for manylinux set to off or non-linux platforms.  
+/// Does nothing for `platform_tag` set to `Off`/`Linux` or non-linux platforms.
 pub fn auditwheel_rs(
     path: &Path,
     target: &Target,
-    manylinux: Option<Manylinux>,
+    platform_tag: Option<PlatformTag>,
 ) -> Result<Policy, AuditWheelError> {
-    if !target.is_linux() || manylinux == Some(Manylinux::Off) {
+    // TODO: add support for musllinux: https://github.com/pypa/auditwheel/issues/305
```
messense, line 263: I think it's better to wait for upstream auditwheel to have a clear policy for musllinux tags before merging.

messense: With https://github.com/PyO3/maturin/pull/543/commits/4ed37993359b6f9dc4e9c96a0915e1c60da189c2 , I think it's ok to merge this without waiting for upstream.

PR #558, opened 2021-05-31
Use a module folder for the pure rust layout

konstin:
Old layout:

```
site-packages
├── my_project.cpython-36m-x86_64-linux-gnu.so
└── my_project-2.1.2.dist-info
    └── ...
```

New layout:

```
site-packages
├── my_project
│   ├── __init__.py
│   └── my_project.cpython-36m-x86_64-linux-gnu.so
└── my_project-2.1.2.dist-info
    └── ...
```

This layout is much more like normal python installs (the module is always in `site-packages/my_project` like all pure python module instead of `site-packages/my_project.some-complex-tag-that-sometimes-changes.so`), but more importantly it should allow shipping pyi files for #553


messense (approved):
Looks reasonable!

PR #560, opened 2021-06-04
Update readme for PEP 621 support

messense:
[Rendered](https://github.com/PyO3/maturin/blob/5b350649a03b10f9a97779f813f3eb0dd2c3ad1e/Readme.md)


Readme.md
```diff
@@ -110,46 +123,47 @@ my-project
...
 
-Note that `package.metadata.maturin.classifier` is also supported for backwards compatibility.
-
-You can use other fields from the [python core metadata](https://packaging.python.org/specifications/core-metadata/) in the `[package.metadata.maturin]` section, specifically ` maintainer`, `maintainer-email` and `requires-python` (string fields), as well as `requires-external` and `provides-extra` (lists of strings) and `project-url` (dictionary string to string)
-
-## pyproject.toml
+## Source distribution
 
-maturin supports building through pyproject.toml. To use it, create a `pyproject.toml` next to your `Cargo.toml` with the following content:
+maturin supports building source distributions (sdist) through `pyproject.toml`. To use it, create a `pyproject.toml` next to your `Cargo.toml` with the following content:
```
konstin, line 152: PEP 517 is more than just source distributions! [build](https://pypi.org/project/build/) e.g. can build a wheel directly by invoking `build_wheel` with minimal setup and afaik pip has a no isolation mode where it also builds wheels directly (I think https://github.com/pypa/pip/pull/9091 is related). For maturin users this is relevant as they might want to do `pip install .` or `pip wheel .`, which only works with a pyproject.toml .

konstin (approved):
Thanks for the docs! Please avoid hard-wrapping lines without changes. (I'm fine with wrapping long lines in markdown files, but then let's do this consistently and with the length limit as `max_line_length` in an editorconfig)

Readme.md
messense: Right. I was thinking the paragraph was only talking about sdist and `pip install`/`pip wheel` usually works with sdist published on PyPI, doesn't work with maturin binary release on GitHub.

PR #561, opened 2021-06-04
Update test crates to use PEP 621

messense:


test-crates/pyo3-mixed-submodule/Cargo.toml
```diff
@@ -8,10 +8,6 @@ edition = "2018"
 
 [package.metadata.maturin]
 name = "pyo3_mixed_submodule.rust_module.rust"
```
messense, line 10: Needs to double check this later.

konstin (approved):
Thanks

PR #562, opened 2021-06-05
Replace `read_distribution` mod with `python-pkginfo` crate

messense:
https://github.com/PyO3/python-pkginfo-rs


src/upload.rs
```diff
@@ -32,6 +33,9 @@ pub enum UploadError {
     /// File already exists
     #[error("File already exists: {0}")]
     FileExistsError(String),
+    /// Read package metadata error
+    #[error("Read package metadata error")]
```
konstin, line 37: How about "Could not read the metadata from the package at {0}" with the PathBuf as field in the error? Or simply "Could not read the metadata from the package" if the path is already included in the backtrace.

Cargo.lock
```diff
@@ -188,6 +188,16 @@ dependencies = [
  "libc",
 ]
 
+[[package]]
+name = "bzip2"
```
konstin, line 192: Are there bzip'd python packages? Never came across any

konstin (commented):
Looks good so far

Cargo.toml
```diff
@@ -58,6 +58,7 @@ once_cell = "1.7.2"
 scroll = "0.10.2"
 target-lexicon = "0.12.0"
 pyproject-toml = "0.1.0"
+python-pkginfo = "0.3.2"
```
konstin, line 61: I think python-pkginfo would also be a good candidate to move to the pyo3 organization.

---

You can remove the mailparse dependency from maturin since read_distribution.rs was the only place using it.

Cargo.lock
messense: https://www.python.org/dev/peps/pep-0527/#file-extensions

> Currently sdist supports a wide variety of file extensions like .tar.gz, .tar, .tar.bz2, .tar.xz, .zip, .tar.Z, .tgz, and .tbz. However, of those the only extensions which get anything more than negligible usage is .tar.gz with 444,338 sdists currently, .zip with 58,774 sdists currently, and .tar.bz2 with 3,265 sdists currently.

It's deprecated, I'm going to move deprecated formats behind a off-by-default feature in `python-pkginfo` crate later.

Cargo.toml
messense: Moved to https://github.com/PyO3/python-pkginfo-rs

PR #564, opened 2021-06-07
Metadata-Version 2.2

### LINKED ISSUE #563: Remove Cargo requirement from prepare_metadata_for_build_wheel?
Currently `prepare_metadata_for_build_wheel` would check Cargo’s existence before generating metadata, and persuambly also needs Cargo during the process, but I’m unsure why that is the case—Admittedly I did not read the source and maybe there is a valid reason. But from my naive understanding, the Python package’s metadata can be retrieved entirely by parsing `pyproject.toml` and `Cargo.toml` manually, and should not require Cargo (since anything that’s too complicated to parse without Cargo are Rust-specific from what I can tell). What am I missing here?

Removing the dependency on Cargo during metadata-generation would be beneficial for tools that perform dependency resolution (locking) and package installation separately, like Pipenv and Poetry. Currently a lock file must be generated on a platform with compatible wheels published, or have Cargo installed. This means that if a project does not publish (say) wheels for Mac, a user must install Cargo for the locking phase, even if they never intend to build the project locally (since the installation happens elsewhere, e.g. on the server or in a Docker container running Linux).

konstin:
Since PEP 643 is accepted we can upgrade to Metadata-Version 2.2

Closes #563


messense:
https://github.com/pypa/warehouse/blob/c7925740d088affbc9e2f8ce98afe23284a6ea59/warehouse/forklift/legacy.py#L426-L439

Looks like warehouse haven't support version 2.2 yet?

cclauss:
This branch has conflicts that must be resolved.

konstin:
I still want this, this is still blocked on pypi, but it makes more sense to redo this when pypi is ready or it'll just gather more and more merge conflicts

PR #567, opened 2021-06-12
Add support for packaging type stubs in pure Rust project layout

messense:
Closes #553

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Add PEP 656 musllinux support in [#543](https://github.com/PyO3/maturin/pull/543)
 * `--manylinux` is now called `--compatibility` and supports musllinux
 * The pure rust install layout changed from just the shared library to a python module that reexports the shared library. This should have now observable consequences for users of the created wheel expect that `my_project.my_project` is now also importable (and equal to just `my_project`)
+* Add support for packaging type stubs in pure Rust project layout in [#567](https://github.com/PyO3/maturin/pull/567)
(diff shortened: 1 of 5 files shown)


ravenexp:
PEP 561 requires `py.typed` marker file to be added to packages containing `.pyi` stubs:

https://www.python.org/dev/peps/pep-0561/#packaging-type-information

messense:
> PEP 561 requires `py.typed` marker file to be added to packages containing `.pyi` stubs:
> 
> https://www.python.org/dev/peps/pep-0561/#packaging-type-information

Thanks! Totally forgot about that...

ravenexp:
I've tried to build my Rust-only extension using this patch, and the embedded type information stub was discovered by both mypy and PyCharm. Everything works as expected.

Maybe an info message for the user could be added, signalling that the type information stub was discovered and included in the distribution. Currently, when I add a magically-named file to the project root directory, I can only hope that I did everything correctly, and there is no way to verify the result except for manually inspecting the built wheel contents.

messense:
> Maybe an info message for the user could be added, signalling that the type information stub was discovered and included in the distribution. Currently, when I add a magically-named file to the project root directory, I can only hope that I did everything correctly, and there is no way to verify the result except for manually inspecting the built wheel contents.

Good idea, added.

```bash
🔗 Found pyo3 bindings with abi3 support for Python ≥ 3.6
🐍 Not using a specific python interpreter (With abi3, an interpreter is only required on windows)
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
📖 Found type stub file at pyo3_pure.pyi
📦 Built wheel for abi3 Python ≥ 3.6 to dist/pyo3_pure-2.1.2-cp36-abi3-macosx_11_0_arm64.whl
```

PR #568, opened 2021-06-15
Support i386 on OpenBSD

dkarama:
Compiling py-adblock on OpenBSD-i386, rust 1.52.1, python 3.8.8 breaks up with the following error 

```
i686 is not supported on OpenBSD
```

The given patch allows building with the above configuration. Currently building py-adblock has been the extent of testing maturin has undergone on i386. Let me know if there is anything more to add.


konstin:
Please add an entry in the changelog, otherwise this looks good

messense (approved):
LGTM

PR #574, opened 2021-06-21
Fix typo in "requires_python" metadata field name

ravenexp:
Upload to GitLab was broken because of https://github.com/PyO3/maturin/pull/438


messense (approved):
Thank you for catching this!

PR #575, opened 2021-06-21
Refuse to include the sdist output tarball into itself

ravenexp:
Add a sanity check preventing unbounded sdist tarball growth
when the user forgets to add the sdist output directory to .gitignore.

Accidentally discovered by running tox on a project where .tox directory
is not in .gitignore. tox builds its sdists inside .tox/dist/.


konstin:
Thanks!

PR #576, opened 2021-06-22
Better error message when users try to run maturin without rust installed

konstin:
This is the error someone sees who tries to use `maturin develop` without rust installed (or who missed the PATH step)

Tested with `PATH= target/debug/maturin develop`.

Before:

```
💥 maturin failed
  Caused by: Failed to run rustc to get the host target
```

After: 

```
💥 maturin failed
  Caused by: rustc, the rust compiler is not installed or not in PATH. This package requires Rust and Cargo to compile extensions. Install it through the system's package manager or via https://rustup.rs/.
```


messense (approved):
Looks good!

PR #577, opened 2021-06-23
Replace `toml_edit` with `toml` crate

### LINKED ISSUE #541: Maturin's toml parser errors out on dotted keys in `Cargo.toml` when building sdists
It looks like something in the sdist build process for Maturin doesn't understand dotted keys in toml files.

Apologies, but I'm on-call this week so I won't be able to dig through Maturin to see what/where for a little while, hence making a bug report issue instead of a fix PR.

-----

### Issue Template Q&A

 * Your python version (`python -V`): *N/A* (abi3, PYO3_NO_PYTHON)
 * Your pip version (`pip -V`): *N/A* (same as above)
   * If this isn't the [latest pip version](https://pip.pypa.io/en/stable/news/), have you checked using the latest pip version? *N/A*
 * The version of the bindings you're using, if any (e.g. `pyo3`, `rust-cpython` or `cffi`): Default
 * Does `cargo build` work? *Yes.*
 * If on windows, have you checked that you aren't accidentally using unix path (those with the forward slash `/`)? Repro'd on macos and linux.

* Create a new repository with two library crates in it:

```
demo/
  lib1/
    Cargo.toml
    src/
      lib.rs
  lib2/
    Cargo.toml
    src/
      lib.rs
```

* Make the `lib1` crate refer to the other by path using [dotted keys](https://toml.io/en/v1.0.0#keys), like so:

```
[dependencies]
lib2.path = "../lib2"
```

* Maturin will fail with an error like the following (library name changed in this copypaste to match example above / redact internal repo names from my dayjob):

```
12:02:12  🔗 Found pyo3 bindings with abi3 support for Python ≥ 3.6
12:02:12  🐍 Not using a specific python interpreter (With abi3, an interpreter is only required on windows)
12:02:13  💥 maturin failed
12:02:13    Caused by: Failed to build source distribution
12:02:13    Caused by: Failed to add local dependency lib2 at /build/lib2 to the source distribution
12:02:13    Caused by: Failed to parse Cargo.toml at /build/lib1/Cargo.toml
12:02:13    Caused by: TOML parse error at line 23, column 5
12:02:13     |
12:02:13  23 | lib2.path      = "../lib2"
```

messense:
Should hopefully fixes #541

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -268,19 +268,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "chrono"
-version = "0.4.19"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "670ad68c9088c2a963aaa298cb369688cf3f9465ce5e2d4ca10e6e0098a1ce73"
-dependencies = [
- "libc",
- "num-integer",
- "num-traits",
- "time",
- "winapi",
-]
-
-[[package]]
 name = "cipher"
 version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


messense:
@bearcage Can you try this patch?

konstin (approved):
I'm a bit sad that we lose debugability, but yeah it's better to use the standard-compliant crate instead

PR #578, opened 2021-06-24
Fail build on invalid combination of --skip-auditwheel and --manylinux

ijl:
If --skip-auditwheel is specified maturin currently ignores the
--manylinux argument and builds a platform-specific wheel
equivalent to --manylinux=off.

It will be clearer and easier for the user to fix if maturin
explicitly fails where the error was specified.

% maturin build --skip-auditwheel --manylinux 2014
💥 maturin failed
  Caused by: --skip-auditwheel can only be specified with --manylinux=off


messense:
I thought `maturin build --skip-auditwheel --manylinux 2014` is signalling **trust me I'm building a manylinux2014 compatible wheel so don't enforce auditwheel check**?

Although that sounds dangerous to do so.

konstin:
Yep, `--skip-auditwheel` as it's now implemented was meant as an escape hatch for when maturin has some false positive. Maybe we should just a big warning that using it is dangerous

ijl:
You're right. The real issue in terms of UX is that when I specified --skip-auditwheel, I expected maturin to error 1 if the glibc version failed validation, not rename it to the equivalent of --manylinux off.

PR #586, opened 2021-07-09
Add a command to generate shell completions

messense:


messense:
Utilized in https://github.com/Homebrew/homebrew-core/pull/80933/files#diff-44f6ebda79b35b72b548674fd4f8e769ef704c412496b62c9a63805807cc1d7fR23-R28

PR #594, opened 2021-07-14
Fold long header fields in Python metadata

### LINKED ISSUE #592: Invalid wheel generated when using PEP 621 metadata
 * Your python version (`python -V`): Python 3.9.5
 * Your pip version (`pip -V`): 21.1.1
   * If this isn't the [latest pip version](https://pip.pypa.io/en/stable/news/), have you checked using the latest pip version? N/A
 * The version of the bindings you're using, if any (e.g. `pyo3`, `rust-cpython` or `cffi`): pyo3 (doesn't really matter)
 * Does `cargo build` work? N/A
 * If on windows, have you checked that you aren't accidentally using unix path (those with the forward slash `/`)? N/A

Generate a wheel with the following in the `pyproject.toml`:

```
[project]
...
readme = "README.md"
requires-python = ">=3.7,<4"
license { file = "LICENSE" }
...
```

The above will inline the text of LICENSE in the METADATA of the generated wheel. However, it seems that maturin doesn't properly fold the headers (i.e. put a space on each new line) per RFC822 https://datatracker.ietf.org/doc/html/rfc822#section-3.1.1

This causes pkginfo to parse the metadata incorrectly, and causes `twine check` to fail.

An example repo where this happened: https://github.com/ethanhs/ryaml/ (specifically commit 1808286db1379d04b6069626cd149cf3cb02149d). I have worked around this by manually folding my LICENSE file...

### LINKED ISSUE #593: Clarify interaction of Cargo.toml and pyproject.toml metadata
I was reading through the readme about PEP 621 support in the [Python metadata section](https://github.com/PyO3/maturin#python-metadata) and I realized that it isn't clear how it interacts with metadata listed in `Cargo.toml`. Which values are sourced from where and what if they disagree?

It appears that they are merged, but it would be nice to have this written down somewhere so it is clear.

EDIT: it seems that these values are merged, and that pyproject.toml takes precedence

messense:
Fixes #592 , closes #593

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,7 +26,7 @@ jobs:
           python-version: 3.8
       - uses: actions/setup-python@v2
         with:
-          python-version: 3.9
+          python-version: 3.9.5
       - name: Install cffi and virtualenv
         run: pip install cffi virtualenv
       - uses: actions-rs/toolchain@v1
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -285,7 +285,7 @@ dependencies = [
  "atty",
  "bitflags",
  "strsim",
- "textwrap",
+ "textwrap 0.11.0",
  "unicode-width",
  "vec_map",
 ]
@@ -872,6 +872,7 @@ dependencies = [
  "tar",
  "target-lexicon",
  "tempfile",
+ "textwrap 0.14.2",
  "thiserror",
  "toml",
  "walkdir",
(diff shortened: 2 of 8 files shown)


messense:
Something is wrong with Python 3.6.6 maybe cache? https://github.com/PyO3/maturin/runs/3067824402?check_suite_focus=true , it's also failed on main branch.

messense:
Will investigate Python 3.9.6 failure in another PR.

PR #605, opened 2021-08-06
Use `[tool.maturin]` options from `pyproject.toml` in build command

messense:
`maturin build` should respect these settings even when not building using PEP 517

https://github.com/PyO3/maturin/blob/a37c450cfbbca50003f6ee3e33204c7bbe416741/maturin/__init__.py#L26-L34

The primary use case of this feature is that when a crate have an off-by-default optional feature say `pyo3` to enable Python extension module, user can specify `cargo-extra-args = "--features pyo3"` in `pyproject.toml` to automatically "pass" them to `maturin build`/`maturin publish`

### DIFF
--- a/src/build_options.rs
+++ b/src/build_options.rs
@@ -162,11 +162,13 @@ impl BuildOptions {
             extra_metadata.python_source.as_deref(),
         )?;
 
+        let mut args_from_pyproject = Vec::new();
         let mut cargo_extra_args = self.cargo_extra_args.clone();
         if cargo_extra_args.is_empty() {
             // if not supplied on command line, try pyproject.toml
             if let Some(args) = pyproject.and_then(|x| x.cargo_extra_args()) {
                 cargo_extra_args.push(args.to_string());
+                args_from_pyproject.push("cargo-extra-args");
             }
         }
         cargo_extra_args = split_extra_args(&cargo_extra_args)?;
@@ -196,9 +198,14 @@ impl BuildOptions {
 
         let bridge = find_bridge(
             &cargo_metadata,
-            self.bindings
-                .as_deref()
-                .or_else(|| pyproject.and_then(|x| x.bindings())),
+            self.bindings.as_deref().or_else(|| {
+                pyproject.and_then(|x| {
+                    if x.bindings().is_some() {
+                        args_from_pyproject.push("bindings");
+                    }
+                    x.bindings()
+                })
+            }),
         )?;
 
         if bridge != BridgeModel::Bin && module_name.contains('-') {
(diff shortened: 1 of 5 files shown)


davidhewitt:
(going to try to review this later tonight)

src/build_options.rs
```diff
@@ -158,7 +162,14 @@ impl BuildOptions {
...
         )?;
 
-        let mut cargo_extra_args = split_extra_args(&self.cargo_extra_args)?;
+        let mut cargo_extra_args = self.cargo_extra_args.clone();
+        if cargo_extra_args.is_empty() {
+            // if not supplied on command line, try pyproject.toml
+            if let Some(args) = pyproject.and_then(|x| x.cargo_extra_args()) {
+                cargo_extra_args.push(args.to_string());
+            }
+        }
```
konstin, line 171: I wonder how to communicate to the user when some setting was loaded from `tool.maturin.some-field` and when the cli overwrote `tool.maturin.some-field`. Maybe it a generic "loading configuration from `[tool.maturin]`" and some message for which fields the cli overwrote the `[tool.maturin]` values. Otherwise it could be rather confusing if you e.g. `git pull` a project, run `maturin build` and then get an error that only occurs because `[tool.maturin]` overwrote some logic when the defaults would have worked.

messense: Thanks for the review. We certainly should inform user that we used configuration from `[tool.maturin]` section.

messense:
I think we can also drop the [cli arguments passing on the Python PEP 517  support side](https://github.com/PyO3/maturin/blob/a37c450cfbbca50003f6ee3e33204c7bbe416741/maturin/__init__.py#L26-L34) now that maturin reads it on the Rust side, the `toml` Python dependency could be dropped then.

konstin:
I think `toml` is still required for `get_requires_for_build_wheel`, but we can delete the `[tool.maturin]` part

davidhewitt:
(It sounds like you two have got this so I'm going to assume I'm not needed any more. Ping me if you do need me to review later!)

konstin (approved):

src/build_options.rs
```diff
@@ -228,6 +259,27 @@ impl BuildOptions {
...
+            })
+        });
+        if platform_tag == Some(PlatformTag::manylinux1()) {
+            eprintln!("⚠  Warning: manylinux1 is unsupported by the Rust compiler.");
+        }
+
+        if !args_from_pyproject.is_empty() {
+            eprintln!(
+                "📡 Using build options {} from pyproject.toml",
+                args_from_pyproject.join(",")
```
konstin, line 280:
```suggestion
                args_from_pyproject.join(", ")
```

src/build_options.rs
```diff
@@ -228,6 +259,27 @@ impl BuildOptions {
...
+                x.compatibility()
+            })
+        });
+        if platform_tag == Some(PlatformTag::manylinux1()) {
+            eprintln!("⚠  Warning: manylinux1 is unsupported by the Rust compiler.");
+        }
+
+        if !args_from_pyproject.is_empty() {
+            eprintln!(
+                "📡 Using build options {} from pyproject.toml",
```
konstin, line 279: :100: emoji choice

anthrotype:
why can the `--manifest-path` option not be set from pyproject.toml, unlike the other maturin options? was it an oversight or a deliberate decision? Thanks for considering

PR #608, opened 2021-08-14
Warn missing `build-backend` in pyproject.toml

### LINKED ISSUE #607: `pyproject.toml` missing `build-backend` crashes maturin
I ran `maturin develop` in PyO3's `word-count` example, and got this crash:

```
$ maturin develop
💥 maturin failed
  Caused by: Failed to parse Cargo.toml into python metadata
  Caused by: pyproject.toml is invalid
  Caused by: pyproject.toml is not PEP 517 compliant: missing field `build-backend` at line 1 column 1
```

This is incorrect. Quoting PEP 517:

> If the pyproject.toml file is absent, or the build-backend key is missing, the source tree is not using this specification, and tools should revert to the legacy behaviour of running setup.py (either directly, or by implicitly invoking the `setuptools.build_meta:__legacy__` backend).

I guess there's an open question what to do in projects that have a `setup.py` file. `maturin` could raise a different error which is more correct, or it could give a warning that it's ignoring the missing `build-backend` key and building anyway?

messense:
Closes #607

### DIFF
--- a/src/auditwheel/audit.rs
+++ b/src/auditwheel/audit.rs
@@ -315,7 +315,7 @@ pub fn auditwheel_rs(
     } else {
         println!(
             "⚠️  Warning: No compatible platform tag found, using the linux tag instead. \
-            You won't be able to upload those wheels to pypi."
+            You won't be able to upload those wheels to PyPI."
         );
 
         // Fallback to linux
--- a/src/auditwheel/audit.rs
+++ b/src/auditwheel/audit.rs
@@ -255,7 +255,7 @@ pub fn auditwheel_rs(
     }
     if let Some(musl_tag @ PlatformTag::Musllinux { .. }) = platform_tag {
         // TODO: add support for musllinux: https://github.com/pypa/auditwheel/issues/305
-        eprintln!("⚠  Warning: no auditwheel support for musllinux yet");
+        eprintln!("⚠️  Warning: no auditwheel support for musllinux yet");
         // HACK: fake a musllinux policy
         return Ok(Policy {
             name: musl_tag.to_string(),
@@ -314,7 +314,7 @@ pub fn auditwheel_rs(
         Ok(policy)
     } else {
         println!(
-            "⚠  Warning: No compatible platform tag found, using the linux tag instead. \
+            "⚠️  Warning: No compatible platform tag found, using the linux tag instead. \
             You won't be able to upload those wheels to pypi."
         );
 
--- a/src/build_options.rs
+++ b/src/build_options.rs
@@ -275,7 +275,7 @@ impl BuildOptions {
             })
         });
         if platform_tag == Some(PlatformTag::manylinux1()) {
-            eprintln!("⚠  Warning: manylinux1 is unsupported by the Rust compiler.");
+            eprintln!("⚠️  Warning: manylinux1 is unsupported by the Rust compiler.");
         }
 
         if !args_from_pyproject.is_empty() {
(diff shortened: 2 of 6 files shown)


messense:
https://github.com/PyO3/maturin/pull/608/commits/5ddbf7ca8ec3e0bc3c12f1590823de7c4ce72de0 looks like this on my computer

CleanShot 2021-08-14 at 20 31 57@2x

src/auditwheel/audit.rs
```diff
@@ -314,7 +314,7 @@ pub fn auditwheel_rs(
         Ok(policy)
     } else {
         println!(
-            "⚠  Warning: No compatible platform tag found, using the linux tag instead. \
+            "⚠️  Warning: No compatible platform tag found, using the linux tag instead. \
             You won't be able to upload those wheels to pypi."
```
davidhewitt, line 318:
```suggestion
            You won't be able to upload those wheels to PyPI."
```

src/build_options.rs
```diff
@@ -430,7 +434,7 @@ pub fn find_bridge(cargo_metadata: &Metadata, bridge: Option<&str>) -> Result<Br
         if !pyo3_node.features.contains(&"extension-module".to_string()) {
             let version = cargo_metadata[&pyo3_node.id].version.to_string();
             println!(
```
davidhewitt, line 436: Should this be `eprintln!` ? Maybe there's value in a `warn!` macro:

```rust
macro_rules! warn {
    ($msg:expr, $($args:tt)+) => {
        eprintln!(concat!("⚠️ Warning: ", $msg), $($args)+)
    };
}
```

src/build_options.rs
```diff
@@ -129,7 +129,11 @@ impl BuildOptions {
         let cargo_toml = CargoToml::from_path(&manifest_file)?;
         let manifest_dir = manifest_file.parent().unwrap();
         let pyproject: Option<PyProjectToml> = if manifest_dir.join("pyproject.toml").is_file() {
-            Some(PyProjectToml::new(manifest_dir).context("pyproject.toml is invalid")?)
+            let pyproject =
+                PyProjectToml::new(manifest_dir).context("pyproject.toml is invalid")?;
+            pyproject.warn_missing_maturin_version();
+            pyproject.warn_missing_build_backend();
```
davidhewitt, line 135: Personally I'd just make these two functions as private functions in `build_options.rs` rather than methods on `PyProjectToml`. I don't think the type actually needs to care about these fields missing?

However that's an extreme nitpick so ignore me if you prefer the current structure.

src/pyproject_toml.rs
```diff
@@ -150,6 +149,27 @@ impl PyProjectToml {
...
+                eprintln!(
+                    "⚠️  Warning: `build-backend` in pyproject.toml isn't '{maturin}', \
+                    source distribution may not build properly.",
+                    maturin = maturin
+                );
+                return false;
+            }
+            return true;
+        }
+        eprintln!("⚠️  Warning: missing `build-backend` in pyproject.toml, source distribution may not build properly.");
```
davidhewitt, line 170: These two messages are so similar, maybe it's better to merge them into something like this?

```
eprintln!("⚠️ Warning: `build-backend` in pyproject.toml is not set to `maturin`, source distribution may not build properly.");
```

Also, "source distribution may not build property" seems a bit vague for users to know what will go wrong. I wonder if we can find something more specific to say, maybe "packaging tools such as pip will not use maturin to build this project"?

davidhewitt (commented):
Thank you, looks good! (And nice emoji fixups!)

I had a couple of ideas for the warning which you may or may not agree with...

src/build_options.rs
messense: Further cleanups can be done separately.

Issue #115, opened 2019-04-17
Full PyPy support 


konstin:
pyo3-pack gets basic pypy support from #105, while some things are still missing.

 - [x] Add some automated testing, ideally on ci as pyo3 does.
 - [x] Figure out the best way to deal with manylinux.
   - [x] [pypy doesn't support maynlinux1](https://github.com/antocuni/pypy-wheels#why-not-manylinux1-wheels). manylinux2010 was [mentioned to be on the agenda of a pypy sprint meeting](https://morepypy.blogspot.com/2018/12/pypy-winter-sprint-feb-4-9-in-dusseldorf.html), however I couldn't find anything about the results.
   - [x] If we need to keep different defaults for pypy and cpython, the `Manylinux` enum should get a default variant so we can disambiguate whether a user requested `manylinux1` explicitly (i.e. we need to fail/skip pypy).
 - [x] Add windows support
   - [x] Find a reference how a native pypy wheel on windows should look like: `python_Levenshtein-0.10.2-pp38-pypy38_pp73-win_amd64.whl` contains `Levenshtein.pypy38-pp73-win_amd64.pyd`: https://paste.rs/ZAJ
 - [x] Test on mac (done by @messense, it works)
 - [x] ~~Skip pypy wheels when uploading~~
 - [x] Autodetect pypy binaries, but only for pyo3
 - [x] Look for a better way than using `SOABI` to determine the native extensions filename since that value isn't define on windows for both cpython and ~~pypy~~. `SOABI` is defined on windows for at least pypy 3.7 and later.
 - [x] #460

mattip:
I got here from a [failing build of pywinpty](https://github.com/conda-forge/pywinpty-feedstock/pull/38) when rebuilding for PyPy3.7 and windows. The failure is [`ERROR: pywinpty-1.1.3-pp3pp73-pypy3_pp73-win_amd64.whl is not a supported wheel on this platform.`](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=370519&view=logs&j=5be07ae1-d8ba-5406-47b6-8e3a3a12f825&t=0bf03e01-0bec-5b85-5316-b1633322e895&l=606) (link may dissapear in a week or so). The project uses `maturin build --release -i pypy3.7.exe`. 

[...]

PR #617, opened 2021-09-02
Autodetect PyPy executables

messense:
Part of #115

### DIFF
--- a/src/python_interpreter.rs
+++ b/src/python_interpreter.rs
@@ -16,6 +16,7 @@ const GET_INTERPRETER_METADATA: &str = include_str!("get_interpreter_metadata.py
 const MINIMUM_PYTHON_MINOR: usize = 6;
 /// Be liberal here to include preview versions
 const MAXIMUM_PYTHON_MINOR: usize = 12;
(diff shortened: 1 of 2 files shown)


src/python_interpreter.rs
```diff
@@ -497,9 +498,16 @@ impl PythonInterpreter {
         let executables = if target.is_windows() {
             find_all_windows(target, min_python_minor)?
         } else {
-            (min_python_minor..MAXIMUM_PYTHON_MINOR)
+            let mut executables: Vec<String> = (min_python_minor..MAXIMUM_PYTHON_MINOR)
                 .map(|minor| format!("python3.{}", minor))
-                .collect()
+                .collect();
+            // Also try to find PyPy for cffi and pyo3 bindings
```
davidhewitt, line 504: It might be less confusing for the user if maturin also finds them for `cpython` bindings and then prints a warning / skips.

messense: I believe rust-cpython does not support PyPy. See https://github.com/dgrunwald/rust-cpython/issues/143#issuecomment-582137522

PR #628, opened 2021-09-19
Cache Docker layers with `actions/cache`

messense:
Let's see how it performs.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -162,13 +162,28 @@ jobs:
         with:
           python-version: 3.8
       - uses: docker/setup-buildx-action@v1
+      - name: Cache Docker layers
+        uses: actions/cache@v2
+        with:
+          path: /tmp/.buildx-cache
+          key: maturin-test-docker-${{ runner.os }}-buildx-${{ github.sha }}
+          restore-keys: |
+            maturin-test-docker-${{ runner.os }}-buildx-
       - name: Build
         uses: docker/build-push-action@v2
         with:
           push: false
           tags: maturin
           load: true
-          cache-from: type=registry,ref=konstin2/maturin:buildcache
+          cache-from: type=local,src=/tmp/.buildx-cache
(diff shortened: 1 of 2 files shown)


messense:
Some overall numbers:

1. clean build without cache: 15m 0s
2. build with cache: 4m 7s

I think it's pretty good.

.github/workflows/test.yml
```diff
@@ -162,13 +162,28 @@ jobs:
...
+            maturin-test-docker-${{ runner.os }}-buildx-
       - name: Build
         uses: docker/build-push-action@v2
         with:
           push: false
           tags: maturin
           load: true
-          cache-from: type=registry,ref=konstin2/maturin:buildcache
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache-new
```
konstin, line 179: Does this still need `mode=max` to still work after code changes?

konstin (approved):
Nice, this is even faster than the docker cache

konstin:
If we do this we should also change it in the docker publish job

.github/workflows/test.yml
messense: I am not sure, though I've never encountered any issue in my past experience with this approach.

messense: I'll push a minor code change to test it.

konstin: It now rebuilds all dependencies, with `mode=max` it should only rebuild maturin itself

messense:
Looks like now it's worse than registry cache. So closing.

konstin:
This might also only be compiling maturin vs. a completly cached be, I'll trigger a new build just to check

konstin:
Fully cached the new solution is a minute faster (https://github.com/PyO3/maturin/runs/3642983166 vs. https://github.com/PyO3/maturin/runs/3644100792), so I'd still merge this.

konstin:
Merged in 3aee10d71aeeee80f288634173835a4a7452bb9b

PR #633, opened 2021-09-21
Install entrypoint scripts in develop command on Unix

### LINKED ISSUE #213: Installation of scripts
I have tried out the Python/Rust example in the `pyo3_mixed` folder and I noticed that `maturin develop` does not install the script declared in `Cargo.toml`. Is this the intended behavior?

messense:
Part of #213


messense:
Windows support is much more complicated so I'm going to defer it until I understand more about it.

PR #634, opened 2021-09-21
Install entrypoint scripts in develop command on Windows

### LINKED ISSUE #213: Installation of scripts

messense:
Closes #213 

Per https://pythonhosted.org/distlib/tutorial.html#using-the-scripts-api:
> On Windows, on systems where the PEP 397 launcher isn’t installed, it is not easy to ensure that the correct Python interpreter is used for a script. You may wish to install native Windows executable launchers which run the correct interpreter, based on a shebang line in the script.

~~This is not as portable as embedding the script into the binary launcher, but better than nothing.~~


messense:
The test failure looks like https://github.com/rust-lang/rust/issues/42869

messense:
Even with PEP 397 launcher installed, `.py` and`.pyw` file extensions have to be properly associated with `py.exe` and `pyw.exe` to make scripts work. So I think this is somewhat useless, we have to go with the `embedding the script into the binary launcher` method to make it always work.

davidhewitt:
Do you need me to test this on my machine? (IIRC you have a mac?)

messense:
> Do you need me to test this on my machine? (IIRC you have a mac?)

I've done some tests on a Windows VM, with file association correctly setup invoking the script in PowerShell works, but the test case somehow fails, I suspect it has something to do with Windows UNC path.

messense:
@davidhewitt Please give it a try, looks like it's working on CI.

src/develop.rs
```diff
@@ -9,12 +9,24 @@ use anyhow::{anyhow, bail, format_err, Context, Result};
...
 use std::fs::OpenOptions;
+#[cfg(target_os = "windows")]
+use std::io::Cursor;
 use std::io::Write;
 #[cfg(not(target_os = "windows"))]
 use std::os::unix::fs::OpenOptionsExt;
 use std::path::Path;
 use std::process::Command;
 
+// Windows launcher comes from https://bitbucket.org/vinay.sajip/simple_launcher/
```
messense, line 20: Binaries comes from https://github.com/alvistack/pypa-distlib/tree/master/distlib

messense:
Tested again in Windows VM and both entrypoint scripts and `cargo test` works. Merging.

davidhewitt:
Sorry I didn't get around to testing this in time - I've had a busy recent few days!

PR #644, opened 2021-10-08
Fix Win32 Conda Python Interpreters

formlogic-robert:
Trying to use maturin with a conda environment on Windows that has my `python.exe` in the root with the `.exe` file extension. I keep getting the error message because the Rust Command is sending it to `ShellExecute` and not the in-process command interpreter.

### DIFF
--- a/src/python_interpreter.rs
+++ b/src/python_interpreter.rs
@@ -167,7 +167,7 @@ fn find_all_windows(target: &Target, min_python_minor: usize) -> Result<Vec<Stri
 
         for path in paths {
             let executable_win = Path::new(&path).join("python.exe");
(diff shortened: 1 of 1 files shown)


messense:
> Trying to use maturin with a conda environment on Windows that has my python.exe in the root with the .exe file extension. I keep getting the error message

What error message did you get?

RobertColton:
I got something like this:
`Error getting Python version info from conda env at C:\Users\<user-name>\Anaconda3\envs\foo1`

I go there and I definitely have `python.exe` in that env folder, which is the same as the folder returned in `conda info -e`. I copy it to my terminal inside Visual Studio Code where conda is activated and add the parameters that maturin does and I can correctly obtain the version info.

PR #648, opened 2021-10-13
Add PEP 660 support

### LINKED ISSUE #624: Add support for PEP 660 - Editable installs for pyproject.toml based builds
PEP: https://www.python.org/dev/peps/pep-0660/

Blocked on https://github.com/pypa/pip/pull/8212

messense:
Closes #624

### DIFF
--- a/src/build_context.rs
+++ b/src/build_context.rs
@@ -273,6 +273,7 @@ impl BuildContext {
             None,
             &self.target,
             false,
+            self.editable,
         )
         .context("Failed to add the files to the wheel")?;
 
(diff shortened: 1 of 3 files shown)


messense:
@davidhewitt Can you give this another try? I've tested the pyo3-mixed example and it works for me.

davidhewitt:
👍 it's now working nicely for me too, awesome work!

messense:
> > I think maybe we can also make `maturin develop` work just like editable wheels by placing dist-info and a `.pth` file in `site-packages`? cc @konstin What do you think?
> 
> Yes please! The current develop behavior is mostly a hack and `.pth` would be much better

Great, I'll do it later in follow-up PRs.

PR #658, opened 2021-11-02
Add mdbook based user guide

### LINKED ISSUE #371: Sphinx/ readthedocs integration guidelines
I'm going to be taking a look at this myself soon, just wanted to make sure nobody else had any tips before I got started!

For its auto-doc stuff, I believe sphinx needs to be able to import the library, which means it needs to be able to build the rust component. Readthedocs, therefore, needs to be able to compile the rust. Simplest way to do this seems to be to use the readthedocs config file to set a docker image with both rust and python (probably `konstin2/maturin` as a first pass) and go from there.

Readthedocs seems worth doing (rather than maintaining in-tree HTML files for gh-pages) because you can easily host multiple branches and reduce git churn.

EDIT: to clarify, this issue is to address the situation where users have a python library with sphinx/RtD documentation, then add a bit of rust to it and find RtD throws a fit.

messense:
This PR starts to draft the layout of maturin user guide, it's very much WIP. Feedback welcome.

Closes #371


guide/src/project_layout.md
```diff
@@ -0,0 +1,130 @@
...
+```
+
+rather than:
+
+```python
+from my_project import my_project
+```
+
+N.B.: there is currently no way to tell maturin to include extra data (e.g.
+type stubs or `package_data` in setuptools) for a pure Rust project. Instead,
```
messense, line 42: @aganders3 This is only partially true, maturin will include type stubs for pure rust project layout if a `$module_name.pyi` file exists. For example: https://github.com/PyO3/maturin/blob/main/test-crates/pyo3-pure/pyo3_pure.pyi

aganders3: Ooh thanks for pointing this out! I played around a bit but just missed it. I will make an update to fix this note.

guide/src/bindings.md
```diff
@@ -0,0 +1,92 @@
...
+
+Cffi wheels are compatible with all python versions including pypy. If `cffi`
+isn't installed and python is running inside a virtualenv, maturin will install
+it, otherwise you have to install it yourself (`pip install cffi`).
+
+Maturin uses cbindgen to generate a header file for [supported Rust
+types](https://github.com/eqrion/cbindgen/blob/master/docs.md#supported-types).
+The header file can be customized by configuring cbindgen through a
+`cbindgen.toml` file inside your project root. Aternatively you can use a build
+script that writes a header file to `$PROJECT_ROOT/target/header.h, like so:
```
messense, line 48:
```suggestion
script that writes a header file to `$PROJECT_ROOT/target/header.h`, like so:
```

netlify.toml
```diff
@@ -0,0 +1,4 @@
+[build]
+  base = "guide"
+  publish = "book/"
+  command = "curl -L https://github.com/rust-lang/mdBook/releases/download/v0.4.12/mdbook-v0.4.12-x86_64-unknown-linux-gnu.tar.gz | tar xvz && ./mdbook build"
```
messense, line 4:
```suggestion
  command = "curl -L https://github.com/rust-lang/mdBook/releases/download/v0.4.13/mdbook-v0.4.13-x86_64-unknown-linux-gnu.tar.gz | tar xvz && ./mdbook build"
```

messense:
I'm gonna merge this now and iterate on main branch.

PR #670, opened 2021-11-08
Add Sphinx integration guideline

### LINKED ISSUE #371: Sphinx/ readthedocs integration guidelines

messense:
Closes #371 

cc @clbarnes

### DIFF
--- a/guide/src/SUMMARY.md
+++ b/guide/src/SUMMARY.md
@@ -11,3 +11,4 @@
 - [Python Metadata](./metadata.md)
 - [Local Development](./develop.md)
 - [Distribution](./distribution.md)
+- [Sphinx Integration](./sphinx.md)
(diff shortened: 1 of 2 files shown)


clbarnes:
Amazing, thank you so much!

PR #674, opened 2021-11-10
[win-arm64] Build without rustls package

nsait-linaro:
Build without rustls package as ring dependency is missing win/arm64 support.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -70,7 +70,10 @@ class PostInstallCommand(install):
                 "--message-format=json",
             ]
 
-            if platform.machine() in ("ppc64le", "ppc64", "powerpc") or get_platform() == "win-arm64":
+            if (
+                platform.machine() in ("ppc64le", "ppc64", "powerpc")
+                or get_platform() == "win-arm64"
+            ):
                 cargo_args.extend(
                     ["--no-default-features", "--features=upload,log,human-panic"]
                 )
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,7 @@ import sys
 import toml
 from setuptools import setup
 from setuptools.command.install import install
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -69,7 +70,10 @@ def run(self):
                 "--message-format=json",
             ]
 
-            if platform.machine() in ("ppc64le", "ppc64", "powerpc"):
+            if (
+                platform.machine() in ("ppc64le", "ppc64", "powerpc")
+                or get_platform() == "win-arm64"
```
messense, line 75: Can we use `platform.platform()` instead of setuptools's private API?

messense (commented):
Thanks, I think ring will have Windows arm64 support in the next release and rustls maybe able to support it too. But for now it's ok to disable rustls for it.

* https://github.com/briansmith/ring/pull/1339

setup.py
nsait-linaro: I've used`distutils/get_platform` to be able to do cross-compilation if required in the future. IIRC platform module would return host machines configuration so if I try to build arm64 binary wheel from x64 windows then it wouldn't work.

Also `platform.platform()` returns `Windows-10-10.0.19041-SP0` on my surface pro which is probably not what we want.

I think either we can use platform.machine() and sys.platform() to check on win/arm64 or use get_platform from distutils.

messense: > I think either we can use platform.machine() and sys.platform() to check on win/arm64

This is fine. I just don't want it use private/deprecated APIs.

nsait-linaro: Yes replaced get_platform with platform.machine() and sys.platform

messense (approved):

PR #675, opened 2021-11-10
Publish win/arm64 wheels to pypi

nsait-linaro:
win_arm64 wheels are permitted in the PyPI repository now so we can start publishing them.

See example win/arm64 binary wheel on PyPI test server: https://test.pypi.org/project/test-maturin/


nsait-linaro:
@messense If you can review and merge this it would be great 👍

messense (approved):
Thanks!

PR #684, opened 2021-11-16
Use environments for deployment on GitHub Actions

messense:
https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment


messense:
image

Looking good.

PR #695, opened 2021-11-22
Add support for excluding files from wheels by `.gitignore`

messense:
Closes #694


dariocurr:
What if I want to include a gitignored file?

PR #696, opened 2021-11-23
Refactor release workflow

messense:
Release PyPI packages together in a dedicated job.


.github/workflows/release.yml
```diff
@@ -8,6 +8,7 @@ jobs:
   release-crates-io:
     name: Release crates.io
     runs-on: ubuntu-latest
+    if: "startsWith(github.ref, 'refs/tags/')"
```
messense, line 11: This isn't necessary but it makes it simpler to test this workflow by simply adding a `on.branches` condition in L4.

PR #702, opened 2021-11-24
Add `cargo-generate` templates for quickly bootstrapping new projects

messense:
#615 

TODO:

- [x] Add a default `.gitignore` file
- [ ] Add a default GitHub Actions configuration, can be done later.
- [x] Let user choose bindings?

PR #705, opened 2021-11-25
Add a `maturin new` command for bootstrapping new projects

### LINKED ISSUE #615: idea: integrate `cargo-generate` to provide a `maturin new` command
I think it'd be really cool if `maturin` (or PyO3) provided a few template projects which users could use to get started with their projects.

`cargo-generate` already exists, so it's of course possible to make a few template repos and leave it there.

However, I think that users coming from Python to Rust for the first time would probably prefer to learn fewer tools, so wrapping up `cargo-generate` inside a `maturin new` command would be a nice experience for them.

It looks like it shouldn't be too hard to call [`cargo_generate::generate`](https://docs.rs/cargo-generate/0.9.0/cargo_generate/fn.generate.html) directly. I think the hard work would be creating and maintaining template repos.

What do you think?

messense:
I feel that this command is simple enough to just write one instead of using `cargo-generate`.

Closes #615 
Closes #702

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -346,6 +346,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "06821ea598337a8412cf47c5b71c3bc694a7f0aed188ac28b836fab164a2c202"
 
 [[package]]
+name = "console"
+version = "0.15.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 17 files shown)


davidhewitt:
Yes this is great, sorry I didn't get around to testing it earlier! I'll probably update the PyO3 readme to suggest using this. 🎉

PR #709, opened 2021-11-27
Use platform tag from `sysconfig.platform` on non-portable Linux

messense:
Stop hard-code linux platform tag for non-portable Linux. This allows build `linux_armv8l` wheels on 64-bit Linux running in 32-bit mode, for example `linux32 python3 -m sysconfig`:

```
Platform: "linux-armv8l"
Python version: "3.8"
Current installation scheme: "posix_prefix"
```

This removes the need for patch on Alpine Linux armhf and armv7 CI runners, [it was painful](https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/25616).

I've tested on a Raspberry Pi running a 64-bit Ubuntu.

See also https://github.com/pypa/packaging/issues/476

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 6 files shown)


messense:
Still having issue with abi3 since it doesn't use a Python interpreter on Linux.

PR #710, opened 2021-11-28
Revert "Upgrade to Rust 2021 edition"

### LINKED ISSUE #708: [Apparently] Needless Rust edition bump breaks build on Void Linux
### Bug Description

Is there a purpose for the edition bump in https://github.com/PyO3/maturin/commit/4d51112a18a331248bb36c0cb5b4c60c6fe0091a? Void Linux is still working toward bumping rust and this makes it impossible to build maturin without reverting the edition change. Maturin compiles fine after reverting only the Cargo.toml edition spec.

Void has rust 1.53.0, which is less than six months old. The 2021 edition was declared stable in rust 1.56.0, released in late October. Unless it solves a real problem with maturin, making a one-month-stable edition a hard requirement seems undesirable.

N/A

N/A

_No response_

### Does `cargo build` work?

- [ ] Yes, it works

- [ ] Yes

### Steps to Reproduce

1. Try to build maturin `0.12.*` with rust 1.53.0.
2. Cry to the heavens about the insanity of the rust ecosystem.

messense:
This reverts commit 55a5fefdfa9c00abfd99b4df7302439b3a3d9461.

We're not using any of new features Rust 2021 edition, will upgrade when we do.

Closes #708

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -68,9 +68,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.48"
+version = "1.0.49"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "62e1f47f7dc0422027a4e370dd4548d4d66b26782e513e98dca1e689e058a80e"
+checksum = "0a03e93e97a28fbc9f42fbc5ba0886a3c67eb637b476dbee711f80a6ffe8223d"
 
 [[package]]
 name = "async-io"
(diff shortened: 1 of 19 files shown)


ahesford:
Thanks! If there are compelling features for maturin  in the 2021 edition, legacy support certainly shouldn't stand in the way. In the meantime, this facilitates packaging of current versions of maturin in Void even as work on updating the rust toolchain continues.

PR #713, opened 2021-11-29
new: include function example for PyO3 bindings

davidhewitt:
Thanks for adding `maturin new`, it's a neat solution!

This is just a tiny suggestion to add a function example to the generated template for PyO3 bindings. I think it helps users figure out what to do next as well as removes an "unused variable" warning for `m: &PyModule`.


messense (approved):
Thanks!

PR #718, opened 2021-11-30
Get cross compiling from pyo3-build-config

aganders3:
This is just a proof-of-concept for now to test https://github.com/PyO3/pyo3/pull/1996


messense:
Thanks for taking care of this.

FYI, you can also use my [manylinux-cross](https://github.com/messense/manylinux-cross) docker images to test it locally.

aganders3:
> Thanks for taking care of this.

You're welcome!

> FYI, you can also use my [manylinux-cross](https://github.com/messense/manylinux-cross) docker images to test it locally.

Thanks, this will help a lot.

.github/workflows/test.yml
```diff
@@ -214,7 +214,7 @@ jobs:
           # CPython
           - target: aarch64-unknown-linux-gnu
             arch: aarch64
-            abi: cp36-cp36m
+            abi: cp37-cp37m
```
messense, line 217: This is unfortunate. I don't think maturin is ready to drop cp36 abi3 support just yet since users can depend on old pyo3 versions.

@davidhewitt what do you think?

messense:
```
🔗 Found pyo3 bindings
cargo:rerun-if-env-changed=PYO3_CROSS
cargo:rerun-if-env-changed=PYO3_CROSS_LIB_DIR
cargo:rerun-if-env-changed=PYO3_CROSS_PYTHON_VERSION
⚠️ Cross-compiling is poorly supported
🐍 Using host CPython 3.9 at python3.9 for cross-compiling preparation
cargo:rerun-if-env-changed=_PYTHON_SYSCONFIGDATA_NAME
cargo:rerun-if-env-changed=PYO3_PYTHON
🐍 Found cross compiling target CPython 3.7m
```

These `cargo:XXX` prints should be avoid I think?

.github/workflows/test.yml
aganders3: Yeah I agree but was just trying to get it to pass the tests here for now. The pyo3-build-config dep could be optional but at that point I'm not sure it has the same benefit. Can the pyo3-build-config version be different from the pyo3 version used for the bindings?

messense: I'd make pyo3-build-config only does cross compiling detection for maturin and leave other stuffs to user's pyo3 dependency to decide.

messense: I couldn't find a link to the failed workflow run for cp36, what error does it gives when cross compiling for python3.6?

aganders3: It got stuck on this assert but I didn't chase down where/why that was being called.

https://github.com/aganders3/pyo3/blob/edb59c3be2793efc01aade80d1772bfe1217272b/pyo3-build-config/src/impl_.rs#L133

messense: I think it https://github.com/PyO3/pyo3/blob/5e8d43c8bd93ab6d0fe599307cc7508a368d7777/build.rs#L139

`build.rs` shouldn't run for maturin, I think we need a way to make it a no-op.

aganders3: Maybe it could toggle based on `CARGO_PRIMARY_PACKAGE`?

messense: Didn't know about `CARGO_PRIMARY_PACKAGE`, TIL! I'm not sure if it works, we can give it a try.

aganders3: > For example add a on-by-default feature to pyo3-build-config and disable it in maturin.

Actually I think this already exists with the `resolve-config` feature so I will try to disable it.

aganders3: Okay I pushed a change that uses an empty feature to disable this. I also found this in the docs on [Overriding Build Scripts](https://doc.rust-lang.org/cargo/reference/build-scripts.html#overriding-build-scripts) which seems like another (possibly more correct?) way to do this that would not require changes to pyo3 or changing the project Cargo.toml.

aganders3: Ah they way I did it also seems to break the non-cross-compiled case. I'll keep trying!

aganders3:
Okay I think I understand this now. The problem is pyo3 expecting `PYO3_CROSS_LIB_DIR` to include `lib` at the end, which was changed in maturin to support pypy cross-compilation (since the pypy `_sysconfigdata*.py` is not under `lib`).

I pointed the test crate (`pyo3-mixed`) to my pyo3 branch where I made a similar change, but that also introduced the Python 3.6 min version which is now on main in pyo3. The 3.6 cross-compilation works if I use `pyo3-build-config` from my branch (for maturin) and the released `pyo3` 0.15 (for the test crate) but I have to reintroduce the `lib` on `PYO3_CROSS_LIB_DIR` because internally pyo3 uses it during build configuration.

test-crates/pyo3-mixed/Cargo.toml
```diff
@@ -7,7 +7,9 @@ readme = "Readme.md"
 edition = "2018"
 
 [dependencies]
-pyo3 = { version = "0.15.1", features = ["extension-module"] }
+# pyo3 = { version = "0.15.1", features = ["extension-module"] }
+pyo3 = { git = "https://github.com/aganders3/pyo3", branch = "expose-cross-compiling", features = ["extension-module", "skip-build-config"] }
```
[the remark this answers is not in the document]
aganders3: Yeah I'm going to revert that. It was an attempt based on @messense's suggestion above that:

> build.rs shouldn't run for maturin, I think we need a way to make it a no-op. For example add a on-by-default feature to pyo3-build-config and disable it in maturin.

My comment above (https://github.com/PyO3/maturin/pull/718#issuecomment-984253175) gets to the root of the problem as I understand it. I don't have a solution at the moment to maintain both Python 3.6 (i.e. test crates able to use pyo3 < 0.16) and PyPy cross-compilation compatibility.

messense:
Somehow https://github.com/PyO3/maturin/pull/728 stopped working just now. I've reverted it in main.

aganders3:
> Somehow #728 stopped working just now. I've reverted it in main.

Yeah that was weird - I will rebase.

aganders3:
Okay sorry for all the CI spam - I'm still getting a grasp on GitHub Actions.

The current state of this PR is basically:
* maturin would require pyo3-build-config >= 0.16 ("main")
* cross-compiling PyPy would require pyo3 >= 0.16 for the target crate
* Python 3.6 support would require pyo3 < 0.16 for the target crate

What I did now is modify the cross-compiling tests to use two crates: `pyo3-mixed` (depends on pyo3 main) `pyo3-mixed-legacy` (same content, depends on pyo3 0.15.1). Then the cross-compiling build matrix skips Python 3.6 for pyo3-mixed and skips pypy for pyo3-mixed-legacy.

Let me know if this seems like a reasonable compromise/path forward.

messense:
> What I did now is modify the cross-compiling tests to use two crates: pyo3-mixed (depends on pyo3 main) pyo3-mixed-legacy (same content, depends on pyo3 0.15.1). Then the cross-compiling build matrix skips Python 3.6 for pyo3-mixed and skips pypy for pyo3-mixed-legacy.

It's fine.

aganders3:
Thanks - this seems ready for review then, with the only changes needed being those to update deps once https://github.com/PyO3/pyo3/pull/1996 is merged/releases.

aganders3:
I'm closing this as there have been quite some changes in both maturin and pyo3 since it was started. I spent a few minutes trying to rebase but think it's probably better to start over if this is wanted.

There are several new public APIs in pyo3-build-config that somewhat overlap maturin. Are you open to reusing more pieces from there? If so I will start to play around and/or open a discussion to see what that might look like.

messense:
> Are you open to reusing more pieces from there? If so I will start to play around and/or open a discussion to see what that might look like.

I do want to at least reuse the sysconfigdata finding and parsing code but it's not urgent. And there is a problem with `pyo3-build-config` in that it may drop support for old Python versions in new pyo3 releases while maturin might still want to support them for longer time.

PR #719, opened 2021-11-30
init: new command similar to `cargo init`

davidhewitt:
I was writing an updated PyO3 README and found I wanted to create the virtualenv inside the project directory . In particular, this is the sequence of commands I wanted to run:

```bash
$ mkdir string_sum && cd string_sum
$ python -m venv .env
$ source .env/bin/activate
$ pip install maturin
$ maturin init
$ maturin develop
```

... but this needed a hypothetical `maturin init` command instead of `maturin new`, because the directory already existed.

So this PR adds it :smile:

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -14,9 +14,11 @@ use fs_err as fs;
 use human_panic::setup_panic;
 #[cfg(feature = "password-storage")]
 use keyring::{Keyring, KeyringError};
+use maturin::GenerateProjectOptions;
 use maturin::{
-    develop, new_project, source_distribution, write_dist_info, BridgeModel, BuildOptions,
-    CargoToml, Metadata21, PathWriter, PlatformTag, PyProjectToml, PythonInterpreter, Target,
+    develop, init_project, new_project, source_distribution, write_dist_info, BridgeModel,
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -270,6 +270,21 @@ enum Opt {
...
+        path: Option<String>,
+        /// Set the resulting package name, defaults to the directory name
+        #[structopt(long)]
+        name: Option<String>,
+        /// Use mixed Rust/Python project layout
+        #[structopt(long)]
+        mixed: bool,
+        /// Which kind of bindings to use
+        #[structopt(short, long, possible_values = &["pyo3", "rust-cpython", "cffi", "bin"])]
+        bindings: Option<String>,
```
messense, line 286: I think we can extract `name`, `mixed` and `bindings` to a seperated struct and use `#[structopt(flatten)]` to reuse them. For example 
https://github.com/PyO3/maturin/blob/fc039bd671430c180589b7d9d8ed7d548cf94798/src/main.rs#L197-L199

messense (commented):
Good idea.

src/main.rs
davidhewitt: 👍 will do this and fix formatting tonight

davidhewitt:
👍 have updated this PR with the suggestion and formatting fix!

messense (approved):

PR #720, opened 2021-12-01
Don't package non-path-dep crates in sdist for workspaces

### LINKED ISSUE #462: Additional non-path-dep crates are being packaged in workspaces, causing file not found errors
> If there are still cases missing, please open an issue!

It does not miss any path dependencies but instead includes crates which are not path dependencies of the root crate being packaged (the Python package crate). Iterating over `resolver.nodes` does not trace the dependencies of the actual crate being packaged but instead includes all *path crates* of the Cargo workspace (https://github.com/PyO3/maturin/commit/3f2a9c532e7cc223636f256e91c033dc79e67a12 does not solve this):

https://github.com/PyO3/maturin/blob/929dbcd2dea3debdad1c34bac2e0f21107633c77/src/source_distribution.rs#L171-L177

I might be able to provide a test case for this issue. I have something like this in mind
```
workspace/python-crate
workspace/generic-library-crate
workspace/helper-tool
```
where `python-crate` depends on `generic-library-crate` but not on `helper-tool`. The current approach seems to unnecessarily include `helper-tool` in the source distribution. In my case `helper-tool` is a binary crate while `generic-library-crate` is a pure library crate. Packaging does then trigger #449 because it tries to package `Cargo.lock` for `helper-binary` rooted in the fact that `cargo package --list` does list `Cargo.lock` for binary crates despite the fact that this file does not exist.

_Originally posted by @koehlma in https://github.com/PyO3/maturin/issues/453#issuecomment-788652178_

messense:
Fixes #462

For the included workspace setup, before:

```bash
❯ tar -ztf dist/workspace_with_path_dep-0.1.0.tar.gz
workspace_with_path_dep-0.1.0/local_dependencies/generic_lib/Cargo.toml
workspace_with_path_dep-0.1.0/local_dependencies/generic_lib/src/lib.rs
workspace_with_path_dep-0.1.0/local_dependencies/dont_include_in_sdist/Cargo.toml
workspace_with_path_dep-0.1.0/local_dependencies/dont_include_in_sdist/src/main.rs
workspace_with_path_dep-0.1.0/Cargo.toml
workspace_with_path_dep-0.1.0/pyproject.toml
workspace_with_path_dep-0.1.0/src/lib.rs
workspace_with_path_dep-0.1.0/PKG-INFO
```

after:

```bash
❯ tar -ztf dist/workspace_with_path_dep-0.1.0.tar.gz
workspace_with_path_dep-0.1.0/local_dependencies/generic_lib/Cargo.toml
workspace_with_path_dep-0.1.0/local_dependencies/generic_lib/src/lib.rs
workspace_with_path_dep-0.1.0/Cargo.toml
workspace_with_path_dep-0.1.0/pyproject.toml
workspace_with_path_dep-0.1.0/src/lib.rs
workspace_with_path_dep-0.1.0/PKG-INFO
```

TODO:

- [x] Fix transitive path dependency.


messense:
This one is a little tricky, I'd love to have more eyes on it.

cc @koehlma

davidhewitt:
👍 happy to review however I don't think I'll manage to get to this before the weekend FYI. My FIFO queue is backed up a bit this week 😅

koehlma:
Hey, thanks for addressing this.

It seems to me that this approach misses transitive path dependencies. Imagine I have the dependency graph `A → B → C` and these are all path dependencies. Now, `A.deps` does contain `B` but not `C` because the dependency graph provided by `cargo_metadata` is transitivity reduced as far as I know. So, after discovering that `B` is a dependency of `A` the dependencies of `B` should be scanned and so on. If I may, I would also suggest factoring out the scanning for path dependencies into its own function.

I once implemented this as follows: https://github.com/koehlma/maturin/blob/master/src/source_distribution.rs#L146 This solution also does not depend on the `repr` of a package identifier.

messense:
@koehlma Thanks for the review!

I'm a little worried that this  `top.dependencies` 
https://github.com/koehlma/maturin/blob/dee86e60b1771055210d04fff5d8509d80a0fe04/src/source_distribution.rs#L155 doesn’t support renamed dependencies. Not sure if it breaks anything, needs to give it a test.

image

koehlma:
@messense My pleasure!

The part of the documentation you quoted is for [`Node`](https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.Node.html#structfield.dependencies) but my code actually uses [`Package`](https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.Package.html#structfield.dependencies). In case a dependency is renamed, one can access the new name via the [`rename`](https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.Dependency.html) field of the dependency. This raises an interesting question: What happens if you use a local variant of a crate via renaming and another variant from crates.io? I am not quite sure which names should end up in the `HashMap` in particular, this may be inconsistent accross transitive dependencies.

konstin (approved):
LGTM

A test that compares the archive file list would be useful

PR #721, opened 2021-12-01
Update keyring to 1.0.0

messense:
Changelog: https://github.com/hwchen/keyring-rs/releases/tag/v1.0.0

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -37,7 +37,7 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: -- -D warnings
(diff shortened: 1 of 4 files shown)


konstin:
This allows making password-storage a default features (at least on linux) since this version doesn't depend on libdbus anymore so `cargo build --target x86_64-unknown-linux-musl --features password-storage` is still statically linked

messense:
> This allows making password-storage a default features (at least on linux) since this version doesn't depend on libdbus anymore so `cargo build --target x86_64-unknown-linux-musl --features password-storage` is still statically linked

Yeah, I thought about it when doing the upgrade. I'll give it a try on CI later.

PR #738, opened 2021-12-10
Replace structopt with Clap 3.0

messense:
[Clap 3.0](https://github.com/clap-rs/clap/issues/2869) should have a stable release soon.


messense:
The FreeBSD CI failure is probably caused by https://github.com/tokio-rs/mio/issues/1539

messense:
[Fig](https://fig.io) integration is cool!

PR #741, opened 2021-12-10
Fix `upload::canonicalize_name()` regex subst

ravenexp:
Python `Pattern.sub()` corresponds to Rust `Regex::replace_all()`,
not `Regex::replace()`.

This fixes a bug where the same Python package `a_b_c` is uploaded
as both `a-b-c` (wheel) and `a-b_c` (sdist).


messense:
Thanks for fix this, could you add a changelog entry please?

PR #742, opened 2021-12-12
Implement auditwheel repair with patchelf

### LINKED ISSUE #649: No automatic musllinux compliance
 * Your python version (`python -V`): Python 3.9.5
 * Your pip version (`pip -V`): Pip 21.3
 * The version of the bindings you're using, if any: PyO3 0.14.5

I'm running the latest version of maturin (0.11.5) and saw that recently maturin should be able to auto-detect musl (#642). However, when I try to build [my project](https://github.com/tmtenbrink/rustfrc) using maturin:

```
poetry run maturin build --release --compatibility musllinux_1_2
```

I still seem to get the same issue as in #642 even though I think the fix is supposed to be in the most recent version.

```
💥 maturin failed
  Caused by: Error ensuring musllinux_1_2 compliance
  Caused by: Your library is not musllinux_1_2 compliant because it links the following forbidden libraries: ["libgcc_s.so.1"]
```

I am running the rust:alpine docker image, with default apk python3, etc.

I am able to fix the problem using auditwheel manually (below the output of auditwheel show, auditwheel repair then fixes it):
```
rustfrc-1.0.2-cp39-cp39-linux_x86_64.whl is consistent with the
following platform tag: "linux_x86_64".

The wheel references external versioned symbols in these
system-provided shared libraries: libgcc_s.so.1 with versions
{'GCC_4.2.0', 'GCC_3.0', 'GCC_3.3'}

The following external shared libraries are required by the wheel:
{
    "libc.musl-x86_64.so.1": "/lib/ld-musl-x86_64.so.1",
    "libgcc_s.so.1": "/usr/lib/libgcc_s.so.1"
}

In order to achieve the tag platform tag "musllinux_1_2_x86_64" the
following shared library dependencies will need to be eliminated:

libgcc_s.so.1
```

messense:
Linux dylib dependency resolution built with [lddtree-rs](https://github.com/messense/lddtree-rs), patch dylib's `soname` and `rpath` is done with [patchelf](https://github.com/NixOS/patchelf) which means user needs to install it beforehand. Would be great if there is a pure Rust version but unfortunately I don't have the expertise to do that.

The code isn't well organized yet, but hey it works™.

Closes #649

### DIFF
--- a/src/auditwheel/audit.rs
+++ b/src/auditwheel/audit.rs
@@ -255,6 +255,7 @@ pub fn auditwheel_rs(
     if !target.is_linux() || platform_tag == Some(PlatformTag::Linux) {
         return Ok((Policy::default(), false));
     }
+    let cross_compiling = target.cross_compiling();
     let arch = target.target_arch().to_string();
     let mut file = File::open(path).map_err(AuditWheelError::IoError)?;
     let mut buffer = Vec::new();
@@ -298,10 +299,15 @@ pub fn auditwheel_rs(
                 should_repair = false;
                 break;
             }
-            Err(AuditWheelError::LinksForbiddenLibrariesError(..)) => {
-                highest_policy = Some(policy.clone());
-                should_repair = true;
-                break;
+            Err(err @ AuditWheelError::LinksForbiddenLibrariesError(..)) => {
+                // TODO: support repair for cross compiled wheels
+                if !cross_compiling {
+                    highest_policy = Some(policy.clone());
+                    should_repair = true;
+                    break;
+                } else {
+                    return Err(err);
+                }
             }
             Err(AuditWheelError::VersionedSymbolTooNewError(..))
             | Err(AuditWheelError::BlackListedSymbolsError(..))
@@ -333,9 +339,14 @@ pub fn auditwheel_rs(
                 should_repair = false;
                 Ok(policy)
             }
-            Err(AuditWheelError::LinksForbiddenLibrariesError(..)) => {
-                should_repair = true;
-                Ok(policy)
+            Err(err @ AuditWheelError::LinksForbiddenLibrariesError(..)) => {
+                // TODO: support repair for cross compiled wheels
+                if !cross_compiling {
+                    should_repair = true;
+                    Ok(policy)
+                } else {
+                    Err(err)
+                }
             }
             Err(err) => Err(err),
         }
--- a/src/auditwheel/audit.rs
+++ b/src/auditwheel/audit.rs
@@ -37,7 +37,7 @@ pub enum AuditWheelError {
     /// The elf file isn't manylinux/musllinux compatible. Contains the list of offending
     /// libraries.
     #[error(
-    "Your library is not {0} compliant because of the presence of too-recent versioned symbols: {1:?}",
+    "Your library is not {0} compliant because of the presence of too-recent versioned symbols: {1:?}. Consider building in a manylinux docker container",
     )]
     VersionedSymbolTooNewError(Policy, Vec<String>),
     /// The elf file isn't manylinux/musllinux compatible. Contains the list of offending
@@ -240,8 +240,9 @@ fn get_default_platform_policies() -> Vec<Policy> {
 /// An reimplementation of auditwheel, which checks elf files for
 /// manylinux/musllinux compliance.
 ///
-/// If `platform_tag`, is None, it returns the the highest matching manylinux/musllinux policy, or `linux`
-/// if nothing else matches. It will error for bogus cases, e.g. if libpython is linked.
+/// If `platform_tag`, is None, it returns the the highest matching manylinux/musllinux policy
+/// and whether we need to repair with patchelf,, or `linux` if nothing else matches.
+/// It will error for bogus cases, e.g. if libpython is linked.
 ///
 /// If a specific manylinux/musllinux version is given, compliance is checked and a warning printed if
 /// a higher version would be possible.
--- a/src/auditwheel/audit.rs
+++ b/src/auditwheel/audit.rs
@@ -33,8 +33,18 @@ pub enum AuditWheelError {
     #[error(
     "Your library is not {0} compliant because it links the following forbidden libraries: {1:?}",
     )]
-    PlatformTagValidationError(Policy, Vec<String>),
-    /// The elf file isn't manylinux/musllinux compatible. Contains unsupported architecture
+    LinksForbiddenLibrariesError(Policy, Vec<String>),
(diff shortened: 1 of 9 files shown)


src/auditwheel/audit.rs
```diff
@@ -33,8 +33,18 @@ pub enum AuditWheelError {
     #[error(
     "Your library is not {0} compliant because it links the following forbidden libraries: {1:?}",
     )]
-    PlatformTagValidationError(Policy, Vec<String>),
-    /// The elf file isn't manylinux/musllinux compatible. Contains unsupported architecture
+    LinksForbiddenLibrariesError(Policy, Vec<String>),
+    /// The elf file isn't manylinux/musllinux compatible. Contains the list of offending
+    /// libraries.
+    #[error(
+    "Your library is not {0} compliant because of the presence of too-recent versioned symbols: {1:?}",
```
konstin, line 40:
```suggestion
    "Your library is not {0} compliant because of the presence of too-recent versioned symbols: {1:?}. Consider building in a manylinux docker container",
```

Or something along the lines, otherwise they'll have a hard time finding out how to fix that

src/auditwheel/audit.rs
```diff
@@ -33,8 +33,18 @@ pub enum AuditWheelError {
...
+    )]
+    VersionedSymbolTooNewError(Policy, Vec<String>),
+    /// The elf file isn't manylinux/musllinux compatible. Contains the list of offending
+    /// libraries with blacked-list symbols.
+    #[error("Your library is not {0} compliant because it depends on black-listed symbols: {1:?}")]
+    BlackListedSymbolsError(Policy, Vec<String>),
```
konstin, line 46: Do you know why those symbols are blacklisted and what the fix would be? I couldn't find anything about them when searching in the manylinux PEPs

src/auditwheel/audit.rs
```diff
@@ -229,10 +251,11 @@ pub fn auditwheel_rs(
     path: &Path,
     target: &Target,
     platform_tag: Option<PlatformTag>,
-) -> Result<Policy, AuditWheelError> {
+) -> Result<(Policy, bool), AuditWheelError> {
```
konstin, line 254: > If `platform_tag`, is None, it returns the the highest matching manylinux/musllinux policy,

to

> If `platform_tag` is None, it returns the the highest matching manylinux/musllinux policy and whether we need to repair with patchelf,

(I can't add a comment up there)

src/build_context.rs
```diff
@@ -251,23 +249,101 @@ impl BuildContext {
...
-                format!("Error ensuring {} compliance", platform_tag)
+        let (policy, should_repair) =
+            auditwheel_rs(artifact, target, platform_tag).with_context(|| {
+                if let Some(platform_tag) = platform_tag {
+                    format!("Error ensuring {} compliance", platform_tag)
+                } else {
+                    "Error checking for manylinux/musllinux compliance".to_string()
+                }
+            })?;
+        let external_libs = if should_repair && !self.editable {
```
konstin, line 269: Is this even executed for editable installs?

konstin (approved):
This is fancy!

> The code isn't well organized yet, but hey it works™.

I couldn't even implement this, I'm happy that you're doing that :D 

One other thing I've noticed is https://github.com/messense/lddtree-rs/blob/aa3ba6af1c7e00b1d084596fecc82dc3886a1a34/src/lib.rs#L138. Would this mean that a bogus lib could silently lead to a broken wheel?

BTW have you seen https://twitter.com/andy_kelley/status/1470338795266457600? It's not ready yet, but if it's completed we could drop the docker container and just compile for manylinux2010 directly. Also [zig cc is on pypi](https://github.com/ziglang/zig-pypi) so we could just list it as a dependency.

messense:
> One other thing I've noticed is https://github.com/messense/lddtree-rs/blob/aa3ba6af1c7e00b1d084596fecc82dc3886a1a34/src/lib.rs#L138. Would this mean that a bogus lib could silently lead to a broken wheel?

Thanks for reporting. I forgot to remove that in https://github.com/messense/lddtree-rs/commit/28d4c27558b7247d47167c8ee6f8987533a4d82a#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759 now that it uses a placeholder `Library` for reporting library not found errors later.

src/build_context.rs
messense: I shouldn't, I was been defensive there.

src/auditwheel/audit.rs
messense: No PEPs about them, it was done to allow `libz` linkage, see https://github.com/pypa/auditwheel/pull/334

messense:
> BTW have you seen https://twitter.com/andy_kelley/status/1470338795266457600? It's not ready yet, but if it's completed we could drop the docker container and just compile for manylinux2010 directly. Also zig cc is on pypi so we could just list it as a dependency.

I have seen that. Hopefully when it's ready it can be used when C++ is not involved!

messense:
Hmm, looks like Python 3.10.1 is broken.

src/build_context.rs
```diff
@@ -251,23 +249,101 @@ impl BuildContext {
...
+                    patchelf::replace_needed(path, n, &soname_map[n].0)?;
+                }
+            }
+            writer.add_file_with_permissions(libs_dir.join(new_soname), path, 0o755)?;
+        }
+
+        // Currently artifact .so file always resides at ${module_name}/${module_name}.so
+        let artifact_dir = Path::new(&self.module_name);
+        let new_rpath = Path::new("$ORIGIN").join(relpath(&libs_dir, artifact_dir));
+        // FIXME: preserving existing rpath entries
```
messense, line 343: Note to self: I need to address this before merging.

polarathene:
> patch dylib's `soname` and `rpath` is done with [patchelf](https://github.com/NixOS/patchelf) which means user needs to install it beforehand.
> 
> Would be great if there is a pure Rust version but unfortunately I don't have the expertise to do that.

Just to let you know that appears to be supported by the `elb` crate now: https://github.com/igankevich/elb/issues/6#issuecomment-3508914461

---

**EDIT:** Nevermind. I [raised an issue to track this](https://github.com/PyO3/maturin/issues/2952#issuecomment-3808734444), but realized there's [a PR opened recently](https://github.com/PyO3/maturin/pull/2932) which contributes a change to use the `arwen` crate instead.

konstin:
The real difficulty is to scale the testing to ensure our patchelf replacement is similarly reliable, I've been finding panics and other bugs.

PR #750, opened 2021-12-19
Support pyo3 abi3-py310 feature

danjujan:
When trying to build a pyo3 project with the `abi3-py310` feature enabled, maturin fails with the following error message.
```
💥 maturin failed
  Caused by: You have selected the `abi3` feature but not a minimum version (e.g. the `abi3-py36` feature). maturin needs a minimum version feature to build abi3 wheels.
```
Fix this by allowing more than one character as minor version.


messense (approved):
Thank you!

PR #751, opened 2021-12-20
Fix reddit problems

konstin:
Some [wrote on r/rust](https://www.reddit.com/r/rust/comments/rk12bg/writing_rust_libraries_for_the_python_scientific/) about some confusions with maturin. These changes fix the two I found actionable.


src/metadata.rs
```diff
@@ -282,22 +282,37 @@ impl Metadata21 {
...
 
             description_content_type = extra_metadata
                 .description_content_type
                 .or_else(|| Some(path_to_content_type(&readme_path)));
         } else {
-            description = None;
-            description_content_type = None;
+            // > If no value is specified for this field, and a file named
+            // > README.md, README.txt or README exists in the package root
+            for readme_guess in ["README.md", "README.txt", "README"] {
```
messense, line 304: Add `README.rst`?

messense:
BTW I'd like to cleanup `Readme.md` and point users to the read the guide instead later.

src/metadata.rs
konstin: Done. Cargo itself doesn't pick up .rst since that can't be parsed in the rust world (esp. on crates.io), but it makes sense for python projects

konstin:
Sounds good, just please make sure that the manylinux/auditwheel story is highlighted in a way that says "you have to take care of  this before publishing to pypi"

tests/common/other.rs
```diff
@@ -1,6 +1,6 @@
 use anyhow::{Context, Result};
 use flate2::read::GzDecoder;
-use maturin::BuildOptions;
+use maturin::{BuildOptions, Metadata21};
```
messense, line 3:
```suggestion
use maturin::BuildOptions;
```

I believe `Metadata21` is unused.

messense (approved):

PR #763, opened 2022-01-02
Add a subcommand wrapper for `zig cc` and `zig c++`

### LINKED ISSUE #761: Windows support for zig based cross compiling
https://github.com/PyO3/maturin/pull/756#discussion_r776128009

messense:
Closes #761

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -8,7 +8,7 @@ use clap::{ArgEnum, IntoApp, Parser};
 use clap_complete::Generator;
 use maturin::{
     develop, init_project, new_project, write_dist_info, BridgeModel, BuildOptions,
-    GenerateProjectOptions, PathWriter, PlatformTag, PythonInterpreter, Target,
+    GenerateProjectOptions, PathWriter, PlatformTag, PythonInterpreter, Target, Zig,
 };
 #[cfg(feature = "upload")]
 use maturin::{upload_ui, PublishOpt};
@@ -163,6 +163,9 @@ enum Opt {
         #[clap(name = "SHELL", parse(try_from_str))]
         shell: Shell,
     },
+    /// Zig linker wrapper
+    #[clap(subcommand)]
+    Zig(Zig),
 }
 
 /// Backend for the PEP 517 integration. Not for human consumption
@@ -448,6 +451,11 @@ fn run() -> Result<()> {
                     )
                 }
             }
+        }
+        Opt::Zig(subcommand) => {
+            subcommand
+                .execute()
+                .context("Failed to create zig wrapper script")?;
         }
     }
 
--- a/src/zig.rs
+++ b/src/zig.rs
@@ -1,0 +1,222 @@
+//! Using `zig cc` as c compiler and linker to target a specific glibc/musl version
+//! as alternative to the manylinux docker container and for easier cross compiling
+
+use crate::target::Arch;
+use crate::{BuildContext, PlatformTag};
+use anyhow::{bail, Context, Result};
+use fs_err as fs;
+use std::env;
+#[cfg(target_family = "unix")]
+use std::fs::OpenOptions;
+use std::io::Write;
+#[cfg(target_family = "unix")]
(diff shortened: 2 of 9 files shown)


src/zig.rs
```diff
@@ -0,0 +1,215 @@
...
+            Zig::Cxx { args } => ("c++", args),
+        };
+        // Replace libgcc_s with libunwind
+        let cmd_args: Vec<String> = cmd_args
+            .iter()
+            .map(|arg| {
+                let arg = if arg == "-lgcc_s" {
+                    "-lunwind".to_string()
+                } else if arg.starts_with('@') && arg.ends_with("linker-arguments") {
+                    // rustc passes arguments to linker via an @-file when arguments are too long
```
messense, line 49: See https://github.com/rust-lang/rust/issues/41190

src/zig.rs
```diff
@@ -0,0 +1,214 @@
...
```
konstin, line 49:
```suggestion
                    // rustc passes arguments to linker via an @-file when arguments are too long
                    // See https://github.com/rust-lang/rust/issues/41190
```

src/main.rs
```diff
@@ -163,6 +163,9 @@ enum Opt {
         #[clap(name = "SHELL", parse(try_from_str))]
         shell: Shell,
     },
+    /// Zig linker wrapper
```
konstin, line 166: Should be hidden, just like pep517

src/zig.rs
```diff
@@ -0,0 +1,214 @@
+use crate::target::Arch;
```
konstin, line 1:
```suggestion
//! Using `zig cc` as c compiler and linker to target a specific glibc/musl version as alternative to the manylinux docker container and for easier cross compiler  

use crate::target::Arch;
```

src/main.rs
```diff
@@ -449,6 +452,9 @@ fn run() -> Result<()> {
                 }
             }
         }
+        Opt::Zig(subcommand) => {
+            subcommand.execute()?;
```
konstin, line 456:
```suggestion
            subcommand.execute().context("Failed to create zig wrapper script")?;
```

src/zig.rs
```diff
@@ -0,0 +1,214 @@
...
+            .spawn()
+            .with_context(|| format!("Failed to run `zig {}`", cmd))?;
+        let status = child.wait().expect("Failed to wait on zig child process");
+        if !status.success() {
+            process::exit(status.code().unwrap_or(1));
+        }
+        Ok(())
+    }
+
+    fn find_zig() -> Result<(String, Vec<String>)> {
```
konstin, line 74:
```suggestion
    /// Search for `python -m ziglang` first and for `zig` second. That way we use the zig from `maturin[ziglang]` first, but users or distributions can also insert their own zig
    fn find_zig() -> Result<(String, Vec<String>)> {
```

src/zig.rs
```diff
@@ -0,0 +1,214 @@
...
+        "{} zig {} -- {} $@",
+        current_exe.display(),
+        command,
+        args
+    )?;
+    Ok(())
+}
+
+#[cfg(not(target_family = "unix"))]
+fn write_linker_script(path: &Path, command: &str, args: &str) -> Result<()> {
```
konstin, line 199:
```suggestion
/// Write a zig cc wrapper batch script for windows
fn write_linker_script(path: &Path, command: &str, args: &str) -> Result<()> {
```

konstin (approved):
:+1:

src/main.rs
messense: It's done here: https://github.com/PyO3/maturin/pull/763/files#diff-49b1f98d9d763bc0e0db9c28ba3dc79ca7413959f017924bc333f7a086d16189R20

messense: It turns out clap 3.0 doesn't support set subcommand to hidden in here, it only can be set in subcommand struct/enum definition site, see https://github.com/clap-rs/clap/discussions/3497

PR #764, opened 2022-01-03
Try rust-cache action

messense:
https://github.com/Swatinem/rust-cache

See also https://github.com/PyO3/pyo3/pull/2057


messense:
Looks like it works pretty well for the maturin binary.

PR #767, opened 2022-01-08
Switch to ureq and some refactorings

konstin:
The first commit is for switching from reqwest to ureq which has a lot less dependencies and therefore compiles faster.

The second commit was an attempt to take the even further that kinda failed (no significant compile time improvements) but I refactored auditwheel and upload into a better module structure: Nearly all auditwheel logic is now in the `auditwheel` module and I removed registry.rs and merged it into upload.rs


messense (approved):
👍

messense:
https://github.com/PyO3/maturin/runs/4751697023?check_suite_focus=true

Looks like we still need to provide a way to disable rustls.

PR #769, opened 2022-01-09
Allow running tests without zig

messense:
Some distros like Alpine Linux run full test suite when packaging maturin, we don't really want them to depend on zig for testing.


konstin:
Thanks for taking care of alpine, but could you please implement this with a feature flag instead? Otherwise we'll miss it when the zig installation is broken

PR #770, opened 2022-01-09
Always run zig tests on GitHub Actions

messense:
https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

Fixes https://github.com/PyO3/maturin/pull/769#issuecomment-1008298408


konstin (approved):
Thank you!

PR #786, opened 2022-01-23
Pass arguments to pep517 command via `MATURIN_PEP517_ARGS` env var

messense:
> Do you know if there currently is any way to actually use config_settings? Last time I checked it wasn't usable with pip and the general from the python packaging discourse consensus was that it's effectively unspecified so we're not doing anything with it.

Yes, this is what I've found out too from further reading. It sounds like `config_settings` doesn't work, so in general we currently don't have a way to pass options to maturin during PEP 517 build. If you don't like having a `MATURIN_PROFILE` env var an alternative could be a `MATURIN_PEP517_ARGS` environment variable which adds arguments to the `maturin pep517` Rust internal command.

_Originally posted by @davidhewitt in https://github.com/PyO3/maturin/issues/774#issuecomment-1016202143_

cc @ionenwks

### DIFF
--- a/maturin/__init__.py
+++ b/maturin/__init__.py
@@ -10,6 +10,7 @@ maturin's emojis.
 """
 
 import os
+import shlex
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


maturin/__init__.py
```diff
@@ -25,6 +25,11 @@ def get_config() -> Dict[str, str]:
     return pyproject_toml.get("tool", {}).get("maturin", {})
 
 
+def get_maturin_pep517_args():
+    args = os.getenv("MATURIN_PEP517_ARGS", "").split()
```
messense, line 29: Note to self: split by whitespace is too naive, should probably use shlex

PR #817, opened 2022-02-24
Add macOS target support for `--zig`

messense:


src/build_options.rs
```diff
@@ -334,6 +334,13 @@ impl BuildOptions {
             );
         }
 
+        let target_dir = cargo_extra_args
```
messense, line 337: TODO: read from `CARGO_BUILD_TARGET_DIR ` / `CARGO_TARGET_DIR` env vars.

PR #820, opened 2022-02-26
Refactor cargo-zigbuild integration

messense:


src/compile.rs
```diff
@@ -209,20 +193,45 @@ fn compile_target(
...
         .chain(&["--"])
         .chain(&rustc_args)
         .collect();
-    let command_str = build_args
-        .iter()
-        .fold("cargo".to_string(), |acc, x| acc + " " + x);
-
-    let mut let_binding = Command::new("cargo");
-    let build_command = let_binding
+    let command_str = build_args.iter().fold(
```
konstin, line 204: you can [join](https://doc.rust-lang.org/std/primitive.slice.html#method.join) here, fold is inscrutable to someone without functional programming background

PR #821, opened 2022-02-27
Migrate Python dependency `uiri/toml` to `tomllib` / `hukkin/tomli`

Contextualist:
Reasons for the migration:

1. [`hukkin/tomli`](https://github.com/hukkin/tomli) has been accepted as part of the Python standard library (see [PEP680](https://www.python.org/dev/peps/pep-0680/) and python/cpython#31498), named as `tomllib`, and will be released along with Python 3.11.
2. [`uiri/toml`](https://github.com/uiri/toml) is unmaintained since Oct 2020, while `hukkin/tomli` is an actively maintained implementation of TOML 1.0 spec, with 100% test coverage.

### DIFF
--- a/maturin/__init__.py
+++ b/maturin/__init__.py
@@ -17,12 +17,15 @@ import sys
 from subprocess import SubprocessError
 from typing import Dict
 
-import toml
+try:
+    import tomllib
+except ModuleNotFoundError:
+    import tomli as tomllib
 
 
 def get_config() -> Dict[str, str]:
-    with open("pyproject.toml", encoding="utf-8") as fp:
-        pyproject_toml = toml.load(fp)
+    with open("pyproject.toml", "rb") as fp:
+        pyproject_toml = tomllib.load(fp)
     return pyproject_toml.get("tool", {}).get("maturin", {})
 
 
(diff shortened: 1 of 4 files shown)


messense:
I remember setuptools-rust removed tomli dependency due to a issue from cryptography: https://github.com/pyca/cryptography/issues/6671#issuecomment-984475879 , I'm not sure if it also affects building maturin from sdist if we merge this.

Contextualist:
If I understand correctly, this is referring to the circular dependency problem between `flit` and `tomli`.

`flit` depends on `tomli` starting from version 3.4, while they fixed the circular dependency issue [in version 3.6](https://flit.readthedocs.io/en/latest/history.html#version-3-6) by vendoring `tomli`.

Building `tomli>=1.1.0` from sdist requires `flit_core>=3.2.0,<4`, so those who have to use `flit_core` version 3.4, 3.5, or 3.5.1 will probably experience the issue.

messense:
Thanks for the clarification!

> Building tomli>=1.1.0 from sdist requires flit_core>=3.2.0,<4, so those who have to use flit_core version 3.4, 3.5, or 3.5.1 will probably experience the issue.

I think this is fine now since flit_core is already fixed in recent versions.

PR #823, opened 2022-02-28
Make PEP 517 build wheel defaults to linux compatibility

messense:
I don't think pep517 build process needs to ensure manylinux/musllinux compatibility, correct me if I'm wrong.

### DIFF
--- a/maturin/__init__.py
+++ b/maturin/__init__.py
@@ -40,7 +40,15 @@ def _build_wheel(
 ):
     # PEP 517 specifies that only `sys.executable` points to the correct
     # python interpreter
-    command = ["maturin", "pep517", "build-wheel", "-i", sys.executable]
+    command = [
+        "maturin",
+        "pep517",
+        "build-wheel",
+        "-i",
+        sys.executable,
+        "--compatibility",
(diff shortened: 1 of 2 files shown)


maturin/__init__.py
```diff
@@ -40,7 +40,15 @@ def _build_wheel(
...
     # python interpreter
-    command = ["maturin", "pep517", "build-wheel", "-i", sys.executable]
+    command = [
+        "maturin",
+        "pep517",
+        "build-wheel",
+        "-i",
+        sys.executable,
+        "--compatibility",
+        "linux",
```
davidhewitt, line 50: Should this be `"off"` rather than `"linux"`?

davidhewitt:
I think I agree; if a user is building via `pip` we can leave them free to do wheel audit themselves as necessary.

maturin/__init__.py
messense: They are the same right now, but I agree `off` is better.

davidhewitt (approved):

konstin (approved):
👍 i somehow assumed we already had disabled that

PR #828, opened 2022-03-04
Use dynamic Python minor version for pyo3 bindings based on its verison

messense:
Resolves https://github.com/messense/maturin-action/issues/30

### DIFF
--- a/src/build_options.rs
+++ b/src/build_options.rs
@@ -1,7 +1,7 @@
 use crate::auditwheel::PlatformTag;
 use crate::build_context::{BridgeModel, ProjectLayout};
 use crate::cross_compile::{find_sysconfigdata, parse_sysconfigdata};
-use crate::python_interpreter::InterpreterKind;
+use crate::python_interpreter::{InterpreterKind, MINIMUM_PYTHON_MINOR};
 use crate::BuildContext;
 use crate::CargoToml;
 use crate::Metadata21;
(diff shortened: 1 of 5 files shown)


messense:
If we simply bump minimum Python version to 3.7, it's fine now, but when maturin hits 1.0, it might not be desirable.

src/build_options.rs
```diff
@@ -23,6 +24,25 @@ use std::path::PathBuf;
...
+        pyo3.insert((0, 14), 6);
+        pyo3.insert((0, 15), 6);
+        pyo3.insert((0, 16), 7);
+        map.insert("pyo3", pyo3);
+
+        let mut pyo3_ffi = HashMap::new();
+        pyo3_ffi.insert((0, 16), 7);
+        map.insert("pyo3-ffi", pyo3_ffi);
+        map
+    });
```
davidhewitt, line 44: I might be tempted to implement this as a function which assumes upper bounds. That way this only needs updating when PyO3 changes minimum version rather than every release.

```suggestion
pub fn pyo3_minimum_python_version(major_version: u8, minor_version: u8) -> Option<(u8, u8)> {
    if (major_version, minor_version) >= (0, 16) {
        Some((3, 7))
    } else if (major_version, minor_version) >= (0, 12) {
        Some((3, 6))
    } else {
        None
    }
 } 
 
pub fn pyo3_ffi_minimum_python_version(major_version: u8, minor_version: u8) -> Option<(u8, u8)> {
    if (major_version, minor_version) >= (0, 16) {
        pyo3_minimum_python_version(major_version, minor_version)
    } else {
        None
    }
 }
```

davidhewitt (approved):
Makes sense to me, just one thought which might make maintenance easier.

Issue #829, opened 2022-03-04
allow to include license text file in the wheel metadata


anthrotype:
the official `wheel` package (used by setuptools' bdist_wheel) has a handy `license_files` option (defined in setup.cfg `[metadata]` section) that allows one to list the license file paths to be included in the generated wheel.

https://wheel.readthedocs.io/en/stable/user_guide.html?highlight=license_files#including-license-files-in-the-generated-wheel-file

Some OSS licenses require one to include the license text along with distributed binaries.

I couldn't find a way to achive this when using maturin. Is it possible to add support for this?

thank you for considering.

messense:
I think you can use the `license` field defined in [PEP 621](https://www.python.org/dev/peps/pep-0621/#license): https://maturin.rs/metadata.html

anthrotype:
oh I didn't know that, thanks! Will that copy the full text of the license file to the wheel .dist-info?

anthrotype:
ok I just tried, I see it copied the full LICENSE text to the METADATA file, not as a separate LICENSE file. That's fine by me. Thank you!

anthrotype:
actually I'm having issues embedding the full text of the LICENSE file when running on Windows.
When I use `license = {file = LICENSE.txt}`, the LICENSE.txt full text is copied to the METADATA file as a [multi-line RFC5322](https://datatracker.ietf.org/doc/html/rfc5322#section-2.2.3) string separated by CRLF, but on Windows I get CRLF twice and that breaks parsing the following README section.

E.g. download the dist.zip file from [here](https://github.com/anthrotype/pngquant-wheels/suites/5545641015/artifacts/178621811), and try to run `twine check pngquant_cli-2.17.0.post2-py3-none-win32.whl`. I get this FAIL error:

```
Checking /Users/clupo/Downloads/dist (8)/pngquant_cli-2.17.0.post2-py3-none-win32.whl: FAILED
  `long_description` has syntax errors in markup and would not be rendered on PyPI.
    line 1335: Warning: Block quote ends without a blank line; unexpected unindent.
```

If I do the same for the other non-Windows wheels (e.g. the mac or linux ones in the same archive), it works. They differ only by their line endings.

messense:
@anthrotype Thanks for the report, I'll look into it soon.

anthrotype:
thanks! The windows wheels now pass twine check.
[...]

PR #836, opened 2022-03-08
Fix license line ending in wheel metadata for Windows

messense:
Fixes https://github.com/PyO3/maturin/issues/829#issuecomment-1060933617


src/metadata.rs
```diff
@@ -794,6 +794,29 @@ mod test {
...
+                } else {
+                    None
+                }
+            })
+            .next()
+            .unwrap();
+        let mut lines = license.split_inclusive('\n');
+        assert_eq!(
+            lines.next().unwrap(),
+            "Copyright (c) 2018-present konstin\r\n"
```
messense, line 813: Without `trim_end()`, this becomes `Copyright (c) 2018-present konstin\r\r\n` on Windows.

https://github.com/messense/maturin/runs/5465287747?check_suite_focus=true

PR #837, opened 2022-03-10
Package license files in `.dist-info`

### LINKED ISSUE #829: allow to include license text file in the wheel metadata

messense:
1. https://peps.python.org/pep-0639/#license-file-multiple-use
2. https://github.com/pypa/setuptools/pull/2645

Fixes #829

### DIFF
--- a/src/module_writer.rs
+++ b/src/module_writer.rs
@@ -833,5 +833,16 @@ pub fn write_dist_info(
         )?;
     }
 
+    if !metadata21.license_files.is_empty() {
+        let license_files_dir = dist_info_dir.join("license_files");
+        writer.add_directory(&license_files_dir)?;
+        for path in &metadata21.license_files {
+            let filename = path.file_name().with_context(|| {
+                format!("missing file name for license file {}", path.display())
+            })?;
+            writer.add_file(license_files_dir.join(filename), path)?;
+        }
+    }
+
     Ok(())
 }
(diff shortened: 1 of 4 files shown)


src/module_writer.rs
```diff
@@ -833,5 +833,14 @@ pub fn write_dist_info(
         )?;
     }
 
+    if !metadata21.license_files.is_empty() {
+        for path in &metadata21.license_files {
+            let filename = path.file_name().with_context(|| {
+                format!("missing file name for license file {}", path.display())
+            })?;
+            writer.add_file(dist_info_dir.join(filename), path)?;
```
anthrotype, line 841: should you not create a 'license_files' subdirectory like the PEP639 mandates?

https://peps.python.org/pep-0639/#license-files-in-project-formats

>  the .dist-info directory MUST contain a license_files subdirectory which MUST contain the files listed in the License-File fields in the METADATA file at their respective paths relative to the license_files directory.

anthrotype:
thanks for working on this! I tested it locally and it works 👍

src/module_writer.rs
messense: Thanks for the review, admittedly I haven't read through PEP 639 yet so I marked this as a draft.

Issue #861, opened 2022-03-30
Package multiple license files in .dist-info


idavis:
## Summary

Add support for [`license-files`](https://peps.python.org/pep-0639/#add-license-files-key) option in metadata for [PEP 639](https://peps.python.org/pep-0639/) so that licenses are added to the `.dist-info` wheel folder. This extends the work of #837 (#829).

There are three separate pieces covered here:
- Supporting `license-files.paths`
- Supporting `license-files.globs`
- Supporting convention for automatically including license files

I think a minimum version of this feature with the highest return would be to add the convention for automatically including license files with the `license-files` option being treated as dynamic to trigger the default behavior. I have a [working spike](https://github.com/PyO3/maturin/compare/main...idavis:pep639-default-license-files-globs) of the implementation but I need to write tests.

Looking through the code, we'd need to add `license-files` to the [pyproject-toml-rs](https://github.com/PyO3/pyproject-toml-rs/blob/main/src/lib.rs) project to deal with serializing the new metadata option to unblock support for `license-files.(paths|globs)` which can build on the work done in #837 to append the license files to the merging metadata.

## Current State

The single license specified in the `pyproject.toml` will be copied into the `.dist-info` wheel folder:

```toml
[project]
license = {file = "LICENSE"}
```
## Desired State

The [link used for multiple file use](https://peps.python.org/pep-0639/#license-file-multiple-use) in #837 is broken (and [in the code](https://github.com/PyO3/maturin/blob/52495dcf806c1e88e632f4f29c2960bd34c1d966/src/metadata.rs#L47)). The [advanced example for PEP 639](https://peps.python.org/pep-0639/#advanced-example) shows the use of `license-files.paths` or `license-files.globs` for listing multiple licenses:

Example with notice:

```toml
[project]
license-files.paths = [
    "LICENSE",
    "NOTICE.txt",
]
```

Multiple license files:

```toml
[project]
license-files.paths = [
    "LICENSE",
    "_vendor/LICENSE",
    "_vendor/LICENSE.APACHE",
    "_vendor/LICENSE.BSD",
]
```

Or alternatively, matched via glob patterns:

```toml
[project]
license-files.globs = [
    "LICENSE*",
    "_vendor/LICENSE*",
]
```

[PEP 649 - I maintain an existing package that’s already licensed](https://peps.python.org/pep-0639/#i-maintain-an-existing-package-that-s-already-licensed) gives further defaults used by `setuptools` for automatically including these files.

[...]

PR #862, opened 2022-04-01
Automatically include license files in `.dist-info/license_files`

idavis:
Implementing parts of #861 for PEP 639 not requiring updates to the metadata version. The changes in #837 added the `.dist-info/license_files` using the `License-File` field.

Any files in the project root matching glob patterns `["LICEN[CS]E*", "COPYING*", "NOTICE*", "AUTHORS*"]` (from PEP 639) are automatically added to the license files list.

Further work for PEP 639 requires update to the metadata versions and approval of 639. The serialization of the fields has been handled by [Adding license_expression and license_files serialization](https://github.com/PyO3/pyproject-toml-rs/pull/2). Additionally, the wheel writer trims paths to write only the raw file name into the `.dist-info/license_files`. Future work will need to modify this behavior as PEP 639 requires relative tree paths to be reflected in the `.dist-info/license_files` hierarchy.


src/metadata.rs
```diff
@@ -153,6 +153,25 @@ impl Metadata21 {
...
             }
 
+            // Until PEP 639 is approved with metadata 2.3, we can assume a
+            // dynamic license-files (also awaiting full 2.2 metadata support)
+            // We're already emitting the License-Files metadata without issue.
+            // license-files.globs = ["LICEN[CS]E*", "COPYING*", "NOTICE*", "AUTHORS*"]
+            let license_include_targets = ["LICEN[CS]E*", "COPYING*", "NOTICE*", "AUTHORS*"];
+            for pattern in license_include_targets.iter() {
+                for license_path in glob::glob(&manifest_path.join(pattern).to_string_lossy())
+                    .expect("No license files found for pattern")
```
konstin, line 163: the `manifest_path` needs to be escaped with `glob::escape` or it'll become part of the pattern. The expect shouldn't happen anyway, but a`?` would be better than possibly panicking

src/metadata.rs
```diff
@@ -835,4 +854,44 @@ mod test {
...
+        let cargo_toml: CargoToml = toml_edit::easy::from_str(&cargo_toml_str).unwrap();
+        let metadata = Metadata21::from_cargo_toml(&cargo_toml, &manifest_path).unwrap();
+
+        // verify Cargo.toml value came through
+        assert_eq!(metadata.license.as_ref().unwrap(), "MIT");
+
+        let license_files_strings: Vec<_> = metadata
+            .license_files
+            .iter()
+            .map(|v| v.to_str().unwrap())
```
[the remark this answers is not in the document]
idavis: Pushed fix, thank you!

src/metadata.rs
idavis: I pushed a fix for the `?` usage instead of the `expect`. I followed the pattern from [source_distribution.rs#L317-L328](https://github.com/PyO3/maturin/blob/main/src/source_distribution.rs#L317-L328). I can update the code to used `Pattern::escape` if you like.

```rust
    if let Some(include_targets) = sdist_include {
        for pattern in include_targets {
            println!("📦 Including files matching \"{}\"", pattern);
            for source in glob::glob(&manifest_dir.join(pattern).to_string_lossy())
                .expect("No files found for pattern")
                .filter_map(Result::ok)
            {
                let target = root_dir.join(&source.strip_prefix(manifest_dir)?);
                writer.add_file(target, source)?;
            }
        }
    }
```

idavis: @konstin Is this what you are looking for?
```rust
let license_include_targets = ["LICEN[CS]E*", "COPYING*", "NOTICE*", "AUTHORS*"];
let escaped_manifest_string = glob::Pattern::escape(manifest_path.to_str().unwrap());
let escaped_manifest_path = Path::new(&escaped_manifest_string);
for pattern in license_include_targets.iter() {
    for license_path in
        glob::glob(&escaped_manifest_path.join(pattern).to_string_lossy())?
            .filter_map(Result::ok)
    {
//....
```

konstin: yep that looks good!

idavis: @konstin Just pushed the change. Thanks for the tip.

messense (approved):
Thanks!

tomjakubowski:
I think that per PEP 639, the directory containing the license files is supposed to be named `licenses`, not `license_files`. https://peps.python.org/pep-0639/#term-root-license-directory

> Defined to be the [project root directory](https://packaging.python.org/en/latest/glossary/#term-Project-Root-Directory) for a [project source tree](https://packaging.python.org/en/latest/glossary/#term-Project-Source-Tree) or [source distribution](https://packaging.python.org/en/latest/glossary/#term-Source-Distribution-or-sdist); and a subdirectory named licenses of the directory containing the [built metadata](https://packaging.python.org/en/latest/glossary/#term-Built-Metadata)— **i.e., the .dist-info/licenses directory**— for a [Built Distribution](https://packaging.python.org/en/latest/glossary/#term-Built-Distribution) or [installed project](https://packaging.python.org/en/latest/glossary/#term-Installed-Project).

PR #874, opened 2022-04-06
Update the default edition for new projects

sa-:
Changing the edition in `src/templates/Cargo.toml.j2` from 2018 to 2021 since it's newer.


messense (approved):
Thanks!

Issue #830, opened 2022-03-05
Better support for cross compiling to Windows


messense:
1. Cross compile to MSVC targets can be done with integrating [cargo-xwin](https://github.com/messense/cargo-xwin).
2. Cross compile to GNU targets can already be done with `--zig` I think.
3. ~~Even with abi3 an interpreter is required on Windows, maybe we can use something like the [python3-dll-a](https://github.com/ravenexp/python3-dll-a) project to generate an Windows import library for `python3.dll` automatically.~~ Resolved in https://github.com/PyO3/pyo3/pull/2282

ravenexp:
Hi!

I've been using maturin to cross-compile PyO3 extension modules to Windows (MinGW) for years now.
I don't use MSVC, so I can't tell whether option 1 currently works. But as for cross-compiling to GNU targets, everything seems to be in place already. PyO3 already has decent Windows cross-compile support.
I've been using custom shell scripts to generate `python3.dll` import libraries, and then finally packaged this functionality into an easy-to-use `python3-dll-a` crate.

I think option 2 is not really needed because there are no libc version issues with MinGW and the default `x86_64-pc-windows-gnu` linker works just fine for Python extension modules (in my experience).

As for option 3, I think I should push for `python3-dll-a` integration into PyO3 itself (into the `pyo3-build-config` sub-crate),
which will solve the import library problem for good. I'll open an issue in the PyO3 project and see how it goes.

As for what `maturin` could do to improve the Windows cross-compile experience, I think solving https://github.com/PyO3/maturin/issues/824 would be a good start. Ideally, `maturin` should not touch the `PYO3_NO_PYTHON` and `PYO3_CROSS_LIB_DIR` env variables at all.

`cargo build --target x86_64-pc-windows-gnu` already works when `PYO3_CROSS_LIB_DIR` is provided, and I think `maturin`
should only use this variable is it wants to override the library location. As for the `PYO3_NO_PYTHON` hack, I strongly feel that it should be removed, because currently there is two competing cross-compile code paths in PyO3 `build.rs` and it's a total mess. I've already submitted numerous bug fixes for PyO3 because `cargo build` and `maturin build` behaviors keep diverging because of the `PYO3_NO_PYTHON` use by `maturin`.

messense:
> I think option 2 is not really needed because there are no libc version issues with MinGW and the default x86_64-pc-windows-gnu linker works just fine for Python extension modules (in my experience).

[...]

PR #875, opened 2022-04-06
Remove `PYO3_NO_PYTHON` hack for `pyo3 >= 0.16.4`

messense:
See https://github.com/PyO3/maturin/issues/830#issuecomment-1089904899

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+* Stop setting `PYO3_NO_PYTHON` environment variable for pyo3 0.16.4 and later in [#875](https://github.com/PyO3/maturin/pull/875)
+
 ## [0.12.12] - 2022-04-07
 
 * Migrate docker image to GitHub container registry in [#845](https://github.com/PyO3/maturin/pull/845)
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -11,6 +11,10 @@ use std::path::{Path, PathBuf};
 use std::process::Stdio;
 use std::str;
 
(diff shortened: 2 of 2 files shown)


messense:
@ravenexp Only the Docker tests failed now:

> error: cannot set a minimum Python version 3.7 higher than the interpreter version 3.6 (the minimum Python version is implied by the abi3-py37 feature)

https://github.com/PyO3/maturin/runs/5847881580?check_suite_focus=true

ravenexp:
I'm confused about where does Python 3.6 come from. `pyo3` v0.16 only supports Python 3.7 and newer.

ravenexp:
PyO3 does not support setting minimum Python version for abi3 lower than 3.7 anymore.

messense:
> I think the issue here is that unless PYO3_NO_PYTHON is set, the build script still tries to detect the host Python version, and refuses to use it when it's older than the minimum supported version.

Yeah, I don't think it makes sense for abi3 wheel on Unix now and on Windows once we integrate `python3-dll-a`.

ravenexp:
> Yeah, I don't think it makes sense for abi3 wheel on Unix now and on Windows once we integrate `python3-dll-a`.

This doesn't make much sense to me, but it's the long existing behavior someone may depend on.
It can be easily changed after https://github.com/PyO3/pyo3/pull/2276 got merged, but this probably needs more discussion as a separate PyO3 issue.

ravenexp:
> @ravenexp Only the Docker tests failed now:
> 
> > error: cannot set a minimum Python version 3.7 higher than the interpreter version 3.6 (the minimum Python version is implied by the abi3-py37 feature)
> 
> https://github.com/PyO3/maturin/runs/5847881580?check_suite_focus=true

This should also work after upgrading `pyo3` to 0.16.4

PR #879, opened 2022-04-16
Don't require a working Python interpreter when `generate-abi3-import-lib` feature is enabled for Windows

messense:
Ref #830


messense:
cc @ravenexp

ravenexp:
Oh, right, this makes it possible to omit the useless `-i python` argument when cross-compiling to Windows and `PYO3_CROSS_LIB_DIR` is not set. Does this also work in the native Windows-to-Windows compilation case when there's no Python interpreter on the system?

messense:
> Does this also work in the native Windows-to-Windows compilation case when there's no Python interpreter on the system?

I haven't test it on Windows yet, thanks for the reminder.

konstin:
:heart_eyes: 

Note that you should also update the message at https://github.com/PyO3/maturin/blob/045a120665f9b879ea2acbfc36e48873fac5bfb2/src/build_options.rs#L729

PR #886, opened 2022-04-29
Re-export `__all__` in `__init__.py` for pure Rust project

mhils:
maturin already re-exports `__doc__`, but `__all__` is also useful to be reexported. There are a bunch of Python documentation generators which make use of this. In our case (I work on [pdoc](https://pdoc.dev)), re-exporting `__all__` means we correctly detect that it's `my_project` and not `my_project.my_project`. 😃 

On that note, thank you folks for the fantastic work you are doing with PyO3/maturin! 🍰 
We're working on some Rust extensions for @mitmproxy and PyO3 has just been super wonderful. ❤️

### DIFF
--- a/src/module_writer.rs
+++ b/src/module_writer.rs
@@ -656,7 +656,7 @@ pub fn write_bindings_module(
             writer.add_bytes(
                 &module.join("__init__.py"),
                 format!(
-                    "from .{module_name} import *\n\n__doc__ = {module_name}.__doc__\n",
+                    "from .{module_name} import *\n\n__doc__ = {module_name}.__doc__\n__all__ = {module_name}.__all__\n",
                     module_name = module_name
                 )
                 .as_bytes(),
--- a/guide/src/project_layout.md
+++ b/guide/src/project_layout.md
@@ -22,7 +22,8 @@ wheel. For convenience, this file includes the following:
 ```python
 from .my_project import *
 
-__doc__ = .my_project.__doc__
+__doc__ = my_project.__doc__
+__all__ = my_project.__all__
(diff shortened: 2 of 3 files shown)


src/module_writer.rs
```diff
@@ -656,7 +656,11 @@ pub fn write_bindings_module(
             writer.add_bytes(
                 &module.join("__init__.py"),
                 format!(
-                    "from .{module_name} import *\n\n__doc__ = {module_name}.__doc__\n",
+                    "from .{module_name} import *\n\
+                    \n\
+                    __doc__ = {module_name}.__doc__\n\
+                    if hasattr({module_name}, \"__all__\"):\n\
+                    \t__all__ = {module_name}.__all__\n",
```
messense, line 663: [Raw string literal](https://stackoverflow.com/questions/26611664/what-is-the-r-operator-in-rust) should be easier.

```suggestion
                    r#"from .{module_name} import *

__doc__ = {module_name}.__doc__
if hasattr({module_name}, "__all__"):
    __all__ = {module_name}.__all__"#,
```

mhils: I didn't know what your preferred code style is, happy to adjust! 😃

messense (approved):
Looks good to me, could you please add a [changelog entry](https://github.com/PyO3/maturin/blob/main/Changelog.md), thanks!

messense:
Released in v0.12.15, should be out soon: https://github.com/PyO3/maturin/actions/runs/2285132178

mhils:
Awesome, thanks! 🍰 😃

PR #887, opened 2022-05-01
fix: Only set RUSTFLAGS when absolutely neccessary

ravenexp:
Setting `RUSTFLAGS` to an empty string has an undesirable side-effect
of overriding any `rustflags` key values from `.cargo/config`, etc.


messense (approved):
Thanks!

PR #756, opened 2021-12-27
Zig for manylinux compliance without docker

konstin:
If this actually works as I imaging it to be, we won't need the manylinux docker anymore!

---

This allows compiling for a specified manylinux version by using `zig cc`s custom glibc targeting (https://andrewkelley.me/post/zig-cc-powerful-drop-in-replacement-gcc-clang.html). You just add `--zig`, and it will target the policy you want, defaulting to manylinux2010. Example from ubuntu 20.04:

```
$ cargo run -- build -m test-crates/pyo3-pure/Cargo.toml -i python --no-sdist 2> /dev/null
🔗 Found pyo3 bindings with abi3 support for Python ≥ 3.6
🐍 Not using a specific python interpreter (With abi3, an interpreter is only required on windows)
📖 Found type stub file at pyo3_pure.pyi
📦 Built wheel for abi3 Python ≥ 3.6 to /home/konsti/maturin/test-crates/pyo3-pure/target/wheels/pyo3_pure-2.1.2-cp36-abi3-manylinux_2_24_x86_64.whl
$ cargo run -- build -m test-crates/pyo3-pure/Cargo.toml -i python --no-sdist --zig 2> /dev/null
🔗 Found pyo3 bindings with abi3 support for Python ≥ 3.6
🐍 Not using a specific python interpreter (With abi3, an interpreter is only required on windows)
📖 Found type stub file at pyo3_pure.pyi
📦 Built wheel for abi3 Python ≥ 3.6 to /home/konsti/maturin/test-crates/pyo3-pure/target/wheels/pyo3_pure-2.1.2-cp36-abi3-manylinux_2_12_x86_64.manylinux2010_x86_64.whl
```

Maturin has to be installed with `maturin[zig]` to get a compatible zig version.

Still missing:
 * musl (code is there, but untested)
 * pyproject.toml integration (do we want/need that?)
 * Documentation in the user guide

PR #896, opened 2022-05-04
Add hardcoded well-known sysconfigs for effortless cross compiling

messense:
Implements https://github.com/PyO3/maturin/pull/756#discussion_r775961992

With this when cross compiling for non-abi3 Linux wheels, there is no need to supply `PYO3_CROSS_LIB_DIR` anymore and `-i pythonX.Y` will just work™️. Internally maturin generates a pyo3 config file based on collected sysconfigs and set `PYO3_CONFIG_FILE` env var for pyo3's build script.

In theory we can extend the support to most Unix systems, ~~for example macOS~~(added macOS in this PR), but that can be done in its dedicated PRs.

### DIFF
--- a/src/build_options.rs
+++ b/src/build_options.rs
@@ -1,7 +1,7 @@
 use crate::auditwheel::PlatformTag;
 use crate::build_context::{BridgeModel, ProjectLayout};
 use crate::cross_compile::{find_sysconfigdata, parse_sysconfigdata};
-use crate::python_interpreter::{InterpreterKind, MINIMUM_PYTHON_MINOR};
+use crate::python_interpreter::{InterpreterConfig, InterpreterKind, MINIMUM_PYTHON_MINOR};
 use crate::BuildContext;
 use crate::CargoToml;
 use crate::Metadata21;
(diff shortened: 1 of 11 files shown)


src/python_interpreter/config.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /// Generate pyo3 config file content
+    pub fn pyo3_config_file(&self) -> String {
+        let mut content = format!(
+            r#"implementation={implementation}
+version={major}.{minor}
+shared=true
+abi3=false
+build_flags=WITH_THREAD
```
davidhewitt, line 69: Nb I think this flag only necessary on Python older than 3.7

src/build_options.rs
```diff
@@ -699,31 +710,77 @@ pub fn find_interpreter(
...
+                            .to_string_lossy();
+                        let (python_impl, python_ver) =
+                            if let Some(ver) = python.strip_prefix("pypy") {
+                                (InterpreterKind::PyPy, ver)
+                            } else if let Some(ver) = python.strip_prefix("python") {
+                                (InterpreterKind::CPython, ver)
+                            } else {
+                                bail!("Unsupported Python interpreter: {}", python);
+                            };
+                        let mut ver_parts = python_ver.split('.');
```
davidhewitt, line 745: You might want to use `split_once` (if msrv 1.52 or above), or check `ver_parts.next()` is none after parsing minor version, so that strings like 3.7.foobar.baz aren't accepted

src/build_options.rs
```diff
@@ -631,26 +655,13 @@ pub fn find_interpreter(
...
-                if let Some(minor) = min_python_minor {
-                    bail!("Couldn't find any python interpreters with version >= 3.{}. Please specify at least one with -i", minor);
-                } else {
-                    bail!("Couldn't find any python interpreters. Please specify at least one with -i");
-                }
-            }
-
+            let mut interpreters = Vec::new();
             if binding_name.starts_with("pyo3") && target.is_unix() && target.cross_compiling() {
                 if let Some(cross_lib_dir) = std::env::var_os("PYO3_CROSS_LIB_DIR") {
```
davidhewitt, line 660: Should there also be a check for `PYO3_CONFIG_FILE` env var, and just skip all this if so?

davidhewitt (commented):
Seems like a nice idea to me!

src/build_options.rs
messense: I think that'd be nice, but it's out of scope of this PR, I'll open a new issue to track it.

konstin (approved):
TBH i don't understand much about pyo3's current cross compiling logic, but the code looks good.

Does this mean that we can now compile and cross-compile to a specific non-abi3 version on linux and mac without having that version installed?

messense:
> Does this mean that we can now compile and cross-compile to a specific non-abi3 version on linux and mac without having that version installed?

Yes, that's the goal of this.

PR #905, opened 2022-05-08
Update 'replace_needed' to reduce total calls to 'patchelf'

messense:
Same as https://github.com/pypa/auditwheel/pull/376


messense:
@aganders3 Thanks for the idea!

PR #906, opened 2022-05-08
Add wheel data support

### LINKED ISSUE #258: Install data
Hello,

 is it possible to install data into python package?

Using `setuptools` it would've look like something this:
``` pyhton
setup(
    ...
    package_data={"module_name": ["data/*.html"]}
    ...
)
```

konstin:
This also introduces tracing for logging, which is neat because you can e.g. `RUST_LOG=...` at runtime.

Fixes #258


guide/src/project_layout.md
```diff
@@ -144,3 +144,17 @@ my-project
 └── src
     └── lib.rs
 ```
+
+## Data
+
+You can add wheel data by creating a `<module_name>.data` folder or setting its location as `data` in `[project.metadata.maturin]`.
```
messense, line 150: I think we should also allow settings it in `[tool.maturin]` section of `pyproject.toml`?

konstin: yes, thanks for the reminder!

guide/src/project_layout.md
```diff
@@ -144,3 +144,17 @@ my-project
 └── src
     └── lib.rs
 ```
+
+## Data
+
+You can add wheel data by creating a `<module_name>.data` folder or setting its location as `data` in pyproject.toml under `[tool.maturin]` or in Cargo.toml under `[project.metadata.maturin]`.
```
messense, line 150: > You can add wheel data by creating a `<module_name>.data` folder

Does this mean that creating such folder will suddenly make a pure Rust project becoming a mixed Rust/Python project since it may create a `<python_module>` folder when `module_name == python_module`?

konstin: mixed/pure and data are orthogonal, they are both options in `ProjectLayout`: https://github.com/PyO3/maturin/blob/5eb0fcb3655206a81ac6eb3a7c71a9450da23223/src/build_context.rs#L56-L67

By default, having a `<module_name>` dir will activate mixed mode, while having a `<module_name>.data` dir will activate "data mode" (btw i'm open to another default directory name, it's just what wheels are using internally and it seemed nice because it's not something people could have by accident like a pure `data` directory)

messense: The problem is that mixed mode also needs a `__init__.py`, so adding a `<module_name>.data` dir to a pure Rust mode will break the build if user forgot to add a `__init__.py`.

https://github.com/PyO3/maturin/pull/906/files#diff-87a24e61912b4f23e30d6031f5adc136adfa6008772880ae53a355cedfaee466R116-R118

konstin: maybe i'm missing something, but i thought that's exactly the case i covered with the `with-data` test crate:

```
├── Cargo.lock
├── Cargo.toml
├── check_installed
│   └── check_installed.py
├── pyproject.toml
├── src
│   └── lib.rs
└── with_data.data
    ├── data
    │   └── data_subdir
    │       └── hello.txt
    └── headers
        └── empty.h
```

There you have no `__init__.py`, but a `with_data.data` and it works as pure-rust package

messense: Oh, great!

Issue #796, opened 2022-02-05
TODO list for 1.0


konstin:
maturin is pretty stable as it stands, but I'd like to make some major changes to cli before doing a 1.0 release:

 
 - [x] Switch to no sdist as default, instead add `--sdist` flag. #955
 - [x] Switch to `-i python` as default, instead add `--find-interpreter` flag. Normally you just need to build for one python because that's the python you're testing with. We keep the interpreter search for cases where you want to build all the wheels from one machine/runner. #964 
 - [x] Try replacing `--cargo-extra-args` and `--rustc-extra-args` with `--`, e.g. `maturin build -m snake/Cargo.toml -- --release --features python-bindings -- --cfg mycfg`. I'm not sure if we really want to allow double `--`, but the single version would be much more ergonomic and intuitive. #972 
 - [x] Finish #501: Have python-first packages with a rust module, anchored in pyproject.toml, and rust-first packages that are really just one module exported from rust and maybe some typings, anchored in Cargo.toml. #1185
 - [x] https://github.com/PyO3/maturin/issues/1047
 - [x] cleanup deprecations: https://github.com/PyO3/maturin/pull/1335

messense:
I'd also like to explore https://github.com/PyO3/maturin/pull/756#discussion_r775961992 for `-i pythonX.Y` for well known Python distributions like manylinux and musllinux without the target `pythonX.Y` binary present for **effortless** cross compilation.

konstin:
would that need any backwards incompatible changes?

messense:
> Try replacing --cargo-extra-args and --rustc-extra-args with --

What do you think about just make `maturin build` behave pretty much the same as `cargo rustc` with some additional maturin specific options? @konstin

konstin:
Sounds good!

konstin:
These are all done, anything blocking remaing for having a 1.0 ~soonish?

messense:
I'd like to wait a few months to squash out bugs.

messense:
Will track new issues in 1.0 milestone: https://github.com/PyO3/maturin/milestone/3

PR #907, opened 2022-05-09
Allow use python interpreters from bundled sysconfig when not cross compiling

messense:
If cross compiling works, native build should work too.


konstin:
What do you think about making some kind of flag/option to prefer the bundled config over locally installed interpreter? I'd expect that the bundled ones would be a bit faster because we avoid pulling external stuff in, they should also work better with cargo's caching (currently sometimes get confused and recompiles stuff because pycharm's launcher's and my terminal's environment are slightly different) and probably also more stable because we don't have to deal with broken/weird python installations.

messense:
I thought about doing that once we switch to `--find-interpreter`. #796

PR #909, opened 2022-05-10
Use setuptools-rust for bootstrapping

jameshilliard:
This seems like a better approach for bootstrapping and fixes things so that the extensions are built during the build phase instead of the install phase.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,11 +9,7 @@
 # `pip install <source dir>` are supported. For creating a source distribution
 # for maturin itself use `maturin sdist`.
 
-import json
-import os
 import platform
-import shutil
-import subprocess
 import sys
 
 try:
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -41,74 +36,6 @@ def finalize_options(self):
...
-            if platform.machine() in ("ppc64le", "ppc64", "powerpc") or (
-                sys.platform == "win32" and platform.machine() == "ARM64"
-            ):
-                cargo_args.extend(
-                    ["--no-default-features", "--features=upload,log,human-panic"]
-                )
-            elif sys.platform.startswith("haiku"):
-                # mio and ring doesn't build on haiku
-                cargo_args.extend(
-                    ["--no-default-features", "--features=log,human-panic"]
```
messense, line 83: We need to keep feature flags, otherwise it won't compile on these platforms.

jameshilliard: Added those back

messense (approved):
Looks good to me, @konstin what do you think?

PR #914, opened 2022-05-13
Allow stubs-only mixed project layout

### LINKED ISSUE #792: Maturin breaks when you have a directory of stub files instead of a single stub file
### Bug Description

When you run `maturin develop` on a pure Rust project containing a directory of stub files, you get the error message `Found a directory with the module name (solders) next to Cargo.toml, which indicates a mixed python/rust project, but the directory didn't contain an __init__.py file.`

From looking at the maturin code, it looks like it thinks you have a mixed python/rust project when it sees a subdirectory with the same name as the library. If you change the stubs directory name then the type hints don't actually work.

Python 3.9

21.2.4

_No response_

### Does `cargo build` work?

- [X] Yes, it works

- [x] Yes

### Steps to Reproduce

I've made an example repo here https://github.com/kevinheavey/maturin-stub-package-error

This is what I did:

1. Create a new project with `maturin init`.
2. Make a directory of stub files where the directory name matches the library name.
3. Run `maturin develop` and get the error `Found a directory with the module name (solders) next to Cargo.toml, which indicates a mixed python/rust project, but the directory didn't contain an __init__.py file.`

messense:
Fixes #792 

Note that this only allows it to build, it doesn't re-export the Rust module to top module level as in https://github.com/PyO3/maturin/blob/4f92d4d9eeb4110ac44bc977f155ef1c8cb3f0bd/src/module_writer.rs#L646-L658

@kevinheavey I'd love to know more about your requirements.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Allow use python interpreters from bundled sysconfig when not cross compiling in [#907](https://github.com/PyO3/maturin/pull/907)
 * Use setuptools-rust for bootstrapping in [#909](https://github.com/PyO3/maturin/pull/909)
 * Allow setting the publish repository URL via `MATURIN_REPOSITORY_URL` in [#913](https://github.com/PyO3/maturin/pull/913)
+* Allow stubs-only mixed project layout in [#914](https://github.com/PyO3/maturin/pull/914)
 * Allow setting the publish user name via `MATURIN_USERNAME` in [#915](https://github.com/PyO3/maturin/pull/915)
 
 ## [0.12.15] - 2022-05-07
--- a/src/build_context.rs
+++ b/src/build_context.rs
@@ -113,7 +113,9 @@ impl ProjectLayout {
         };
 
         if python_module.is_dir() {
-            if !python_module.join("__init__.py").is_file() {
+            if !python_module.join("__init__.py").is_file()
+                && !python_module.join("__init__.pyi").is_file()
+            {
                 bail!("Found a directory with the module name ({}) next to Cargo.toml, which indicates a mixed python/rust project, but the directory didn't contain an __init__.py file.", module_name)
             }
 


kevinheavey:
> @kevinheavey I'd love to know more about your requirements.

So for my use case, I just want to mimic the standard experience of a Python project with multiple submodules (`from mypackage.module_a import foo`, `from mypackage.module_b import bar`. This is just because I'm working on something where the namespace will get very polluted otherwise.

> it doesn't re-export the Rust module to top module level

I don't understand what this entails. I'm guessing it means there's something that I can't do if I use this approach?

messense:
It means it doesn't import

```python
In [1]: import maturin_stub_package_error
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
Input In [1], in <cell line: 1>()
----> 1 import maturin_stub_package_error

ModuleNotFoundError: No module named 'maturin_stub_package_error'

In [2]: from maturin_stub_package_error import maturin_stub_package_error
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
Input In [2], in <cell line: 1>()
----> 1 from maturin_stub_package_error import maturin_stub_package_error

ModuleNotFoundError: No module named 'maturin_stub_package_error'
```

kevinheavey:
So how does one import the package then?

messense:
> So how does one import the package then?

Never mind, I think that's a problem with my IPython setup, using `python3` works.

kevinheavey:
Nice, and with this approach, do things like `maturin develop` and `maturin publish` "just work"? And everything gets uploaded to PyPI in one package?

messense:
I think it should, note that you need to add a `py.typed` file to python module folder to actually enable type hints.

kevinheavey:
Sounds awsome! The last thing I'm curious about is why we're referring to a mixed project layout. In my head this is just a pure Rust layout but with a directory of stub files instead of a single one

messense:
Using a mixed project layout is the simplest solution, and we actually generate a mixed project layout for pure Rust project internally anyway.

PR #915, opened 2022-05-13
Use `MATURIN_USERNAME` environment variable

ravenexp:
Include it for feature parity with twine's `TWINE_USERNAME`.

Refactor `MATURIN_PASSWORD` variable handling:
note that `clap` attributes can not be used until
<https://github.com/clap-rs/clap/issues/3726>
is resolved.


messense (approved):
Thanks!

PR #919, opened 2022-05-15
Integrate `cargo-xwin` for cross compiling to Windows MSVC targets

### LINKED ISSUE #830: Better support for cross compiling to Windows

messense:
Closes #830

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -194,7 +194,11 @@ fn compile_target(
 
     let target_triple = target.target_triple();
     let mut build_command = if target.is_msvc() && target.cross_compiling() {
+        #[cfg(feature = "cargo-xwin")]
         let mut build = cargo_xwin::Build::new(Some(context.manifest_path.clone()));
+        #[cfg(not(feature = "cargo-xwin"))]
+        let mut build = cargo_zigbuild::Build::new(Some(context.manifest_path.clone()));
+
         build.target = vec![target_triple.to_string()];
         // Pass zig command to downstream, eg. python3-dll-a
         if let Ok((zig_cmd, zig_args)) = cargo_zigbuild::Zig::find_zig() {
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -192,26 +192,43 @@ fn compile_target(
         .chain(&rustc_args)
         .collect();
 
-    let mut build = cargo_zigbuild::Build::new(Some(context.manifest_path.clone()));
     let target_triple = target.target_triple();
-    if !context.zig {
-        build.disable_zig_linker = true;
-        if target.user_specified {
-            build.target = vec![target_triple.to_string()];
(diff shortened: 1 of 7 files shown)


messense:
cc @ravenexp

messense:
I think `zig dlltool` is same as `llvm-dlltool` and they are not compatible with mingw64's dlltool, thus use `zig dlltool` for windows gnu targets isn't possible? @ravenexp

ravenexp:
It depends on what linker `zig cc` uses for its MinGW targets. I think Zig always links with `lld`, which uses the Windows import library format, same as `LINK.EXE`. MinGW-format import libraries are only suitable when `x86_64-w64-mingw32-gcc` is used as the linker.

ravenexp:
> It depends on what linker `zig cc` uses for its MinGW targets. I think Zig always links with `lld`, which uses the Windows import library format, same as `LINK.EXE`.

I checked this directly just to be sure, and `zig cc` indeed uses `lld-link` when compiling for `-target x86_64-windows-gnu`:

```
$ zig cc -v -target x86_64-windows-gnu test.c

--skip--
lld-link -ERRORLIMIT:0 -NOLOGO -STACK:16777216 -MACHINE:X64 -OUT:/x/.cache/zig/o/d5879460d6961e986517a114b13aaef2/a.exe -IMPLIB:/x/.cache/zig/o/d5879460d6961e986517a114b13aaef2/test.lib`
```

This means `dlltool` from MinGW binutils _can not_ be used with `cargo zigbuild` and `maturin --zig`.

src/compile.rs
```diff
@@ -192,26 +192,50 @@ fn compile_target(
...
+
+    if context.zig {
+        // Pass zig command to downstream, eg. python3-dll-a
+        if let Ok((zig_cmd, zig_args)) = cargo_zigbuild::Zig::find_zig() {
+            let zig_cmd = if zig_args.is_empty() {
+                zig_cmd
+            } else {
+                format!("{} {}", zig_cmd, zig_args.join(" "))
+            };
+            build_command.env("ZIG_COMMAND", zig_cmd);
```
messense, line 235: @ravenexp `ZIG_COMMAND` is exposed here, it will be either `zig` or `python3 -m ziglang`.

ravenexp: Good, I'll implement support for `ZIG_COMMAND` in https://github.com/PyO3/python3-dll-a/pull/18.

PR #928, opened 2022-05-19
Add support for building bin bindings wheels with multiple platform tags

### LINKED ISSUE #927: Add support for combined manylinux and musllinux platform tags for bin bindings
For example `
maturin-0.12.17-py3-none-musllinux_1_1_x86_64.manylinux_2_17_x86_64.whl ` if we are compiling bin bindings to musl target. This way you only need to ship one wheel for both GNU and MUSL libc systems.

Should help with https://github.com/PyO3/maturin/discussions/926

messense:
This allows `maturin build --manylinux manylinux_x_y musllinux_x_y -b bin` while disallows any other forms of multiple platform tags. (Which is the old behavior before this change).

Closes #927

### DIFF
--- a/src/build_options.rs
+++ b/src/build_options.rs
@@ -61,9 +61,11 @@ pub struct BuildOptions {
         name = "compatibility",
         long = "compatibility",
         alias = "manylinux",
-        parse(try_from_str)
+        parse(try_from_str),
+        multiple_values = true,
+        multiple_occurrences = true
     )]
-    pub platform_tag: Option<PlatformTag>,
+    pub platform_tag: Vec<PlatformTag>,
 
     /// The python versions to build wheels for, given as the names of the
     /// interpreters. Uses autodiscovery if not explicitly set.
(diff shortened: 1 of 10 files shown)


src/build_options.rs
```diff
@@ -328,37 +330,89 @@ impl BuildOptions {
...
+        if platform_tags
+            .iter()
+            .any(|tag| tag == &PlatformTag::manylinux1())
+        {
             eprintln!("⚠️  Warning: manylinux1 is unsupported by the Rust compiler.");
         }
 
+        match bridge {
+            BridgeModel::Bin => {
+                // Only support two different kind of platform tags when compiling to musl target
```
messense, line 375: We can lift this limitation if we want to in the future.

konstin:
i wonder if it makes sense to have a `staticlinux` or similar tag which automatically gets tagged with both just like universal2. that way we could keep the single argument logic internally

messense:
> i wonder if it makes sense to have a `staticlinux` or similar tag which automatically gets tagged with both just like universal2

I think people are more familiar with the standard manylinux/musllinux tags, a `staticlinux` tag diverges from the standard.

> that way we could keep the single argument logic internally

I was thinking about lift the limitation of single platform tags in the future. It's fine to tag a wheel with multiple manylinux/musllinux tags as long as they're audited.

PR #938, opened 2022-05-26
Add bindings detection to bin targets

### LINKED ISSUE #311: Building Rust binaries
 * Your python version (`python -V`): `3.8.1`
 * Your pip version (`pip -V`): `20.1.1` (latest version)
 * The version of the bindings you're using, if any (e.g. `pyo3`, `rust-cpython` or `cffi`): `pyo3 0.10.1`
 * Does `cargo build` work? `yes`
 * If on windows, have you checked that you aren't accidentally using unix path (those with the forward slash `/`)? `yes`

I'm attempting to build a binary crate (no lib!) as a Python package. The desire is to have an executable which is installable through `pip`. I want to use `pip` not `cargo` because it integrates strongly with the python ecosystem, and also then I can distribute prebuilt binaries without users needing to get a Rust toolchain.

Is this use case supported by maturin? Naively trying to make a binary crate and running `maturin develop` fails:

```
maturin develop
🔗 Found pyo3 bindings
⚠  Warning: You're building a library without activating pyo3's `extension-module` feature. See https://pyo3.rs/v0.10.1/
building-and-distribution.html#linking
🐍 Found CPython 3.8 at python.exe
error: no library targets found in package `pyactix`
💥 maturin failed
  Caused by: Cargo build finished with "exit code: 101": `cargo rustc --message-format json --manifest-path Cargo.toml -
-lib --
```

Perhaps a workaround instead of supporting binaries like this is for me to create a `lib`, and then add a Python script to call into it which could be installed similar to `setuptools`'s [scripts](https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-scripts-keyword-argument)

I'd be willing to contribute code to support either of these two approaches, if this is something you think is possible. I'm not familar with the maturin codebase so might need a bit of pointing in the right direction.

messense:
Closes #311


src/build_context.rs
```diff
@@ -24,7 +24,9 @@ pub enum BridgeModel {
     /// A native module with c bindings, i.e. `#[no_mangle] extern "C" <some item>`
     Cffi,
     /// A rust binary to be shipped a python package
-    Bin,
+    /// The String is the name of the bindings
+    /// providing crate, e.g. pyo3, the number is the minimum minor python version
+    Bin(Option<(String, usize)>),
```
messense, line 29: @davidhewitt I wonder should we support **abi3** bin bindings, if so I think `BridgeModel` needs a refactoring to separate `bin`/`lib` targets from `pyo3`/`rust-cpython` bindings crate. Something like

```rust
enum BuildType {
    Bin(Option<BridgeModel>),
    Library(BridgeModel),
}

enum BridgeModel {
    Cffi,
    Bindings(String, usize),
    BindingsAbi3(u8, u8),
}
```

davidhewitt: Based on https://github.com/PyO3/pyo3/pull/2328#issuecomment-1107871639 I understood that it's not really possible to link to `libpython3.so` (i.e. abi3 distribution) without fixing issues in upstream CPython.

messense:
The Windows test failure is odd, it works when execute `pyo3-bin.exe` directly from terminal, but `subprocess.checkout` returns exit code 3221225781.

image

PR #945, opened 2022-05-29
Drop python 3.6

### LINKED ISSUE #842: Drop support for Python 3.6
It won't happen now, I think we should follow manylinux: https://github.com/pypa/manylinux/issues/1260.

konstin:
3.6 is end of life since end of last year.

Also bumps the version to 0.13 since it's a breaking change.

Closes #842


tests/common/integration.rs
```diff
@@ -164,10 +164,10 @@ pub fn test_integration_conda(package: impl AsRef<Path>, bindings: Option<String
 
     // Create environments to build against, prepended with "A" to ensure that integration
     // tests are executed with these environments
-    create_conda_env("A-pyo3-build-env-36", 3, 6);
     create_conda_env("A-pyo3-build-env-37", 3, 7);
     create_conda_env("A-pyo3-build-env-38", 3, 8);
     create_conda_env("A-pyo3-build-env-39", 3, 9);
+    create_conda_env("A-pyo3-build-env-39", 3, 10);
```
messense, line 170:
```suggestion
    create_conda_env("A-pyo3-build-env-310", 3, 10);
```

messense (commented):
Maybe we should release main branch as a new 0.12.x version before merging this?

tests/common/integration.rs
konstin: oops, thanks

konstin:
:+1: rebased onto main

messense (approved):

PR #948, opened 2022-06-01
Add support for building with multiple binary targets

### LINKED ISSUE #947: Package multiple binary targets
Currently for bin bindings we pass `--bins` to `cargo rustc`, but it only compiles when `--strip` is absent, otherwise it will fail with the following error message

```bash
error: extra arguments to `rustc` can only be passed to one target, consider filtering
the package by passing, e.g., `--lib` or `--bin NAME` to specify a single target
💥 maturin failed
  Caused by: Cargo build finished with "exit status: 101": `cargo rustc --manifest-path example/Cargo.toml --message-format json --bins -- -C link-arg=-s`
```

Even when `--strip` is absent, maturin only packages one of the many binaries into the wheel.

Ideally maturin should support package multiple binary targets, which will resolve this `TODO` in https://github.com/PyO3/maturin/blob/9d197c49830bd721075b12c85c3981a6754273d5/src/compile.rs#L138-L141

messense:
Closes #947

### DIFF
--- a/src/build_context.rs
+++ b/src/build_context.rs
@@ -592,6 +592,9 @@ impl BuildContext {
     ) -> Result<PathBuf> {
         let artifacts = compile(self, python_interpreter, &self.bridge)
             .context("Failed to build a native library through cargo")?;
+        let artifacts = artifacts
+            .get(0)
+            .context("Failed to build a native library through cargo")?;
 
         let artifact = artifacts.get("cdylib").cloned().ok_or_else(|| {
             anyhow!(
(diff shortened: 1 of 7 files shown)


src/build_context.rs
```diff
@@ -709,14 +712,16 @@ impl BuildContext {
...
-            .file_name()
-            .expect("Couldn't get the filename from the binary produced by cargo");
-        self.add_external_libs(&mut writer, artifact, ext_libs)?;
+        for (artifact, artifact_ext_libs) in artifacts.iter().zip(ext_libs) {
+            // I wouldn't know of any case where this would be the wrong (and neither do
+            // I know a better alternative)
+            let bin_name = artifact
+                .file_name()
+                .expect("Couldn't get the filename from the binary produced by cargo");
+            self.add_external_libs(&mut writer, artifact, artifact_ext_libs)?;
```
messense, line 721: TODO: Multiple binaries may have same external shared library dependencies, so we should only bundle one version.

PR #955, opened 2022-06-07
Don't build sdist by default in `maturin build` command

messense:
This replaces `--no-sdist` with `--sdist` option to only build sdist
when `--sdist` is specified.

Part of #796


messense:
@konstin I wonder if we should keep `--no-sdist` in `maturin publish` command?

konstin:
good question, not sure tbh. i'd like poeple to publish sdists, on the other hand that's where the file-already-exists errors come from

messense:
Yes, publish sdist by default only works well when publish for single OS and architecture only, otherwise file-already-exists errors are inevitable unless `--skip-existing` is specified.

PR #957, opened 2022-06-08
Add a `--target` option to `maturin list-python` command

messense:


Skelectric:
How does this option work on Windows? Maturin isn't finding all the python interpreters in my environment, so I'm trying to use this option to pass the path directly but I cannot get it to work.

I've tried the following (where PATH = D:\Users\....\python.exe)

maturin list-python --target <PATH>

```
The '<' operator is reserved for future use.
    + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : RedirectionNotSupported
```

maturin list-python --target PATH

```
maturin failed 
Caused by : Unknown target triple PATH
```

maturin list-python --target --PATH

`If you tried to supply `--PATH` as a value rather than a flag, use `-- --PATH``

maturin list-python --target -- --PATH

`If you tried to supply `--D:\Users` as a value rather than a flag, use `-- --D:\Users``

PR #961, opened 2022-06-09
Pass C/C++ build variables through to cargo

ijl:
This allows the configuration of cc::Build to compile a C/C++
dependency.

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -15,6 +15,10 @@ use std::str;
 /// without `PYO3_NO_PYTHON` environment variable
 const PYO3_ABI3_NO_PYTHON_VERSION: (u64, u64, u64) = (0, 16, 4);
 
+const C_LIBRARIES_BUILD_ENVIRONMENT: [&str; 7] = [
+    "CARCH", "CHOST", "CC", "CXX", "CFLAGS", "CXXFLAGS", "LDFLAGS",
+];
+
 /// Builds the rust crate into a native module (i.e. an .so or .dll) for a
 /// specific python version. Returns a mapping from crate type (e.g. cdylib)
 /// to artifact location.
@@ -168,7 +172,7 @@ fn compile_target(
         .map(String::as_str)
         .collect();
 
-    let mut rust_flags = env::var_os("RUSTFLAGS");
+    let mut rust_flags = env::var("RUSTFLAGS").unwrap_or(String::new());
 
     // We need to pass --bin / --lib to set the rustc extra args later
     match bindings_crate {
@@ -182,9 +186,7 @@ fn compile_target(
             // We must only do this for libraries as it breaks binaries
             // For some reason this value is ignored when passed as rustc argument
             if context.target.is_musl_target() {
-                rust_flags
-                    .get_or_insert_with(Default::default)
-                    .push(" -C target-feature=-crt-static");
+                rust_flags.push_str(" -C target-feature=-crt-static");
             }
         }
     }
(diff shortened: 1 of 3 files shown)


ijl:
@messense I don't know how desirable this is to you--RFC I think, but this proves out that it works. The motivation is to compile yyjson (C) as a backend in orjson. I think this may benefit other projects that are using `setuptools-rust`.

src/compile.rs
```diff
@@ -341,6 +344,14 @@ fn compile_target(
         build_command.env("PYO3_CROSS_LIB_DIR", lib_dir);
     }
 
+    // pass context for cross-compiles with C/C++ libraries
+    for key in C_LIBRARIES_BUILD_ENVIRONMENT {
+        if let Ok(val) = env::var(key) {
```
ijl, line 349: This could also iterate through and look for `CHOST_<target>` and similar vars for cross-compilation I suppose, but I'm not really familiar.

messense:
I was under the impression that [`std::process::Command`](https://doc.rust-lang.org/std/process/struct.Command.html#method.new) inherit the current process’s environment, so why is this necessary? What error did you get without this patch?

src/compile.rs
```diff
@@ -168,7 +172,7 @@ fn compile_target(
         .map(String::as_str)
         .collect();
 
-    let mut rust_flags = env::var_os("RUSTFLAGS");
+    let mut rust_flags = env::var("RUSTFLAGS").unwrap_or(String::new());
```
messense, line 175: Making `rust_flags` an `Option` was intentional: https://github.com/PyO3/maturin/pull/887

src/compile.rs
messense: Indeed, you can refer to cc-rs: https://github.com/rust-lang/cc-rs/blob/f2e1b1c9ff92ad063957382ec445bc54e9570c71/src/lib.rs#L2354-L2372

messense:
And we can verify that cargo inherits environment by running the following command

```
❯ PYO3_PRINT_CONFIG=1 cargo run build -m test-crates/pyo3-pure/Cargo.toml
   Compiling maturin v0.13.0-beta.4 (/Users/messense/Projects/maturin)
    Finished dev [unoptimized + debuginfo] target(s) in 6.56s
     Running `target/debug/maturin build -m test-crates/pyo3-pure/Cargo.toml`
🔗 Found pyo3 bindings with abi3 support for Python ≥ 3.7
🐍 Not using a specific python interpreter
   Compiling pyo3-ffi v0.16.5
error: failed to run custom build command for `pyo3-ffi v0.16.5`

Caused by:
  process didn't exit successfully: `/Users/messense/Projects/maturin/test-crates/pyo3-pure/target/debug/build/pyo3-ffi-63f1fdf88550546f/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-env-changed=PYO3_CROSS
  cargo:rerun-if-env-changed=PYO3_CROSS_LIB_DIR
  cargo:rerun-if-env-changed=PYO3_CROSS_PYTHON_VERSION
  cargo:rerun-if-env-changed=PYO3_CROSS_PYTHON_IMPLEMENTATION
  cargo:rerun-if-env-changed=PYO3_PRINT_CONFIG

  -- PYO3_PRINT_CONFIG=1 is set, printing configuration and halting compile --
  implementation=CPython
  version=3.7
  shared=true
  abi3=true
  lib_name=python3.9
  lib_dir=/Users/messense/.pyenv/versions/3.9.11/lib
  executable=/Users/messense/.pyenv/versions/3.9.11/bin/python
  pointer_width=64
  build_flags=
  suppress_build_script_link_lines=false

  note: unset the PYO3_PRINT_CONFIG environment variable and retry to compile with the above config
💥 maturin failed
  Caused by: Cargo build finished with "exit status: 101": `cargo rustc --manifest-path test-crates/pyo3-pure/Cargo.toml --message-format json --lib -- -C link-arg=-undefined -C link-arg=dynamic_lookup -C link-args=-Wl,-install_name,@rpath/pyo3_pure`
```

You can see that `pyo3-build-config` captured the `PYO3_PRINT_CONFIG` env var and we didn't pass it in maturin explicitly.

ijl:
Oh I see, thanks.

PR #963, opened 2022-06-10
Add support for cross compiling PyPy wheels when abi3 feature is enabled

messense:
Previously it will incorrectly use host PyPy which does not work.

### DIFF
--- a/src/build_options.rs
+++ b/src/build_options.rs
@@ -987,6 +987,25 @@ pub fn find_interpreter(
                         platform: None,
                         runnable: false,
                     }])
+                } else if target.cross_compiling() {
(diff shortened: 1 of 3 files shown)


src/build_options.rs
```diff
@@ -987,6 +987,10 @@ pub fn find_interpreter(
                         platform: None,
                         runnable: false,
                     }])
+                } else if target.cross_compiling() {
+                    // cross compiling to PyPy with abi3 feature enabled,
+                    // we cannot use host pypy so switch to bundled sysconfig instead
+                    find_interpreter_in_sysconfig(interpreter, target)
```
messense, line 993: TODO: only need to process PyPy, CPython is ok since it supports abi3.

PR #964, opened 2022-06-10
Add `--find-interpreter` option to `build` and `publish` commands

messense:
Some behaviors changed:

1. `maturin build` now implicitly translate to `maturin build -i python3`, meaning that we only build for the current python3 interpreter in `PATH`
2. `maturin build --find-interpreter` makes maturin search for interpreters on host, also search for bundled sysconfig when cross compiling
3. Mix `-i` and `--find-interpreter` is not allowed
4. `maturin publish` does the same as `maturin build`

Part of #796

### DIFF
--- a/src/build_options.rs
+++ b/src/build_options.rs
@@ -70,6 +70,10 @@ pub struct BuildOptions {
     #[clap(short, long, multiple_values = true, multiple_occurrences = true)]
     pub interpreter: Vec<PathBuf>,
 
+    /// Find interpreters from the host machine
+    #[clap(short = 'f', long, conflicts_with = "interpreter")]
+    pub find_interpreter: bool,
+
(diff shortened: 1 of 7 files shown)


src/build_options.rs
```diff
@@ -70,6 +70,10 @@ pub struct BuildOptions {
     #[clap(short, long, multiple_values = true, multiple_occurrences = true)]
     pub interpreter: Vec<PathBuf>,
 
+    /// Find interpreters from the host machine
+    #[clap(short = 'F', long, conflicts_with = "interpreter")]
```
messense, line 74: I'm not sure which short one is better, `-F` or `-I`?

konstin: would go with `-f` (lowercase)

konstin (approved):
:+1:

PR #966, opened 2022-06-12
Add support for Linux armv6l

messense:
Old Raspberry Pi uses `linux_armv6l`, some projects want to support them, for example https://github.com/eclipse-zenoh/zenoh-python/issues/64.

armv6l doesn't not support manylinux/musllinux, so it will only produce `linux_armv6l` tag.


YuanYuYuan:
@messense, many thanks for your great help!

Issue #904, opened 2022-05-08
Add support for wasm32-unknown-emscripten


messense:
Status:

- [x] #974 
- [ ] What about `wasm32-wasi`?

Upstream:

* Python WASM support: https://pythondev.readthedocs.io/wasm.html
* https://github.com/pyodide/pyodide/issues/1973

messense:
Let's worry about `wasm32-wasi` when Python have better support for it.

PR #974, opened 2022-06-19
Add support for `wasm32-unknown-emscripten` target

messense:
TODO:
- [x] Workaround https://github.com/emscripten-core/emscripten/issues/17191
- [x] Add CI

Part of #904


src/target.rs
```diff
@@ -305,9 +314,13 @@ impl Target {
                     format!("macosx_{}_{}_arm64", arm64_major, arm64_minor)
                 }
             }
+            // Windows
             (Os::Windows, Arch::X86) => "win32".to_string(),
             (Os::Windows, Arch::X86_64) => "win_amd64".to_string(),
             (Os::Windows, Arch::Aarch64) => "win_arm64".to_string(),
+            // Emscripten
+            // FIXME: should include emscripten version, example: html_py_ever-0.1.0-cp310-cp310-emscripten_3_1_13_wasm32.whl
```
messense, line 322: @hoodmane Is there any documentation about wheel naming in Pyodide?

messense: https://github.com/pyodide/pyodide/blob/38fae6e9cbcdd8e932a53bf1c763279dc55729d5/pyodide-build/pyodide_build/common.py#L15-L18

I think it's generated here in Pyodide, I wonder if there is any standard between different Python Emscripten implementations.

hoodmane: We are going based on what the `packaging` package says:
```py
>>> import packaging.tags
>>> str(next(packaging.tags.sys_tags()))
'cp310-cp310-emscripten_3_1_13_wasm32'
```

.github/workflows/test.yml
```diff
@@ -148,6 +148,29 @@ jobs:
...
+          override: true
+      - uses: mymindstorm/setup-emsdk@v11
+        with:
+          version: 3.1.13
+          actions-cache-folder: emsdk-cache
+      - uses: Swatinem/rust-cache@v1
+      - name: Build wheels
+        run: |
+          set -ex
+          cargo run build -m test-crates/pyo3-mixed/Cargo.toml -o dist --target wasm32-unknown-emscripten -i python3.11
```
hoodmane, line 171: We're still on 3.10 for now:
```suggestion
          cargo run build -m test-crates/pyo3-mixed/Cargo.toml -o dist --target wasm32-unknown-emscripten -i python3.10
```

messense: Changed in `noxfile.py`.

.github/workflows/test.yml
```diff
@@ -148,6 +148,29 @@ jobs:
...
+      # TODO: Install Pyodide and run tests
```
hoodmane: We should make at least an alpha release of Pyodide since installing the development version is a bit inconvenient. It's also been longer than normal since we made a release.

src/compile.rs
```diff
@@ -214,6 +218,22 @@ fn compile_target(
             ];
             rustc_args.extend(mac_args);
         }
+    } else if target.is_emscripten() {
+        shared_args.push("-Zbuild-std");
```
hoodmane, line 222: Hopefully https://github.com/rust-lang/rust/pull/98149 will be merged soon and you won't have to use this.

src/compile.rs
```diff
@@ -214,6 +218,22 @@ fn compile_target(
...
+        rust_flags
+            .get_or_insert_with(Default::default)
+            .push(" -C relocation-model=pic");
+        let emscripten_args = &[
+            "-C",
+            "relocation-model=pic",
+            "-C",
+            "target-feature=+mutable-globals",
```
hoodmane, line 230: You probably only need to pass `relocation-model=pic` once and I think `-Ctarget-feature=+mutable-globals` isn't needed anymore.
```suggestion
        let emscripten_args = &[
            "-C",
            "relocation-model=pic",
```

hoodmane:
Thanks for working on this @messense!

src/compile.rs
hoodmane: When that happens we can also remove `-Crelocation_model=pic`.

messense:
Thanks for the review and your work on `pyo3` and `setuptools-rust` @hoodmane !

PR #978, opened 2022-06-21
Emscripten: Use -Z link-native-libraries=no and remove emcc wrapper

hoodmane:
After many hours of searching I finally found this flag.


messense (commented):
Great!, I think you can remove the whole `if` block from `L263` to `L282`

messense (approved):
Thanks, I'll do the rest of the cleanup myself.

Issue #897, opened 2022-05-04
Add support for `PYO3_CONFIG_FILE`


messense:
Should there also be a check for `PYO3_CONFIG_FILE` env var, and just skip all this if so?

_Originally posted by @davidhewitt in https://github.com/PyO3/maturin/pull/896#discussion_r864464255_

messense:
Just noticed that maturin needs more fields, namely [`abiflags`, `ext_suffix ` and `abi_tag`](https://github.com/PyO3/maturin/blob/ef654268d8c3f85a70672cc886107d0cc8a35b7c/src/python_interpreter/config.rs#L30-L41), they are not defined in [`PYO3_CONFIG_FILE`](https://docs.rs/pyo3-build-config/latest/pyo3_build_config/struct.InterpreterConfig.html).

PR #983, opened 2022-06-24
Improvements to Emscripten/Pyodide CI

hoodmane:
Since we released new Pyodide alpha, we can clean up the CI job a lot.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -162,7 +162,7 @@ jobs:
           cd tests
           npm i pyodide@0.21.0-alpha.2
           cp node_modules/pyodide/repodata.json .
-          EMSCRIPTEN_VERSION=node -p "require('./repodata.json').info.platform"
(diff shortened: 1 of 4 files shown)


hoodmane:
On my local build I get:
```
💥 maturin failed
  Caused by: Invalid PYO3_CONFIG_FILE
  Caused by: missing value for ext_suffix
```
but if I put `ext_suffix` into the `PYO3_CONFIG_FILE` it says it shouldn't be there. @messense @davidhewitt?

messense:
> 

https://github.com/PyO3/maturin/issues/897#issuecomment-1118219286

https://github.com/PyO3/pyo3/blob/3fd0c0e1421c71b0e5755427eaaa886cc67cd9c1/pyo3-build-config/src/impl_.rs#L436

This is a bug, I guess we can strip maturin specific config keys before passing to pyo3.

messense (approved):
Thanks!

PR #1001, opened 2022-07-10
Add 64-bit RISC-V support

felixonmars:
Tested with `orjson` and it built fine with all tests passed on a
riscv64 machine.

This change is not sufficient though, because `ring` doesn't support
riscv64 yet. I was building `maturin` against a custom version of `ring`
to verify this change.

Please let me know what could be done to move forward :)

### DIFF
--- a/src/target.rs
+++ b/src/target.rs
@@ -60,6 +60,7 @@ pub enum Arch {
     X86_64,
     S390X,
     Wasm32,
+    Riscv64,
 }
 
 impl fmt::Display for Arch {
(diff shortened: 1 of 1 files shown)


messense:
Thanks, please run `cargo format`.

messense (approved):
I'll fix the code format issue after merging.

felixonmars:
> I'll fix the code format issue after merging.

Thanks! I was about to do it :P

messense:
Since ring doesn't support RISC-V yet, we should also add it to https://github.com/PyO3/maturin/blob/1339808e6620cb5c8b590a7818f34fc3a9e51cef/setup.py#L49-L52

@felixonmars do you know what `platform.machine()` returns on 64-bit RISC-V?

felixonmars:
> what does `platform.machine()` return on 64-bit RISC-V?

```
>>> platform.machine()
'riscv64'
```

messense:
Released in v0.13.1.

PR #1008, opened 2022-07-17
Add support for invoking with `python3 -m maturin`

### LINKED ISSUE #1007: Allow running Maturin through Python
## Reasoning

Python has support for running the [`__main__.py`](https://docs.python.org/3/library/__main__.html#main-py-in-python-packages) file of a package through the `python -m <pkg>` command. I noticed that while installing Maturin exposes the `maturin` command, it does not allow you to run `python -m  maturin`.

One of the reasons for allowing this method is to choose which specific Python interpreter to use. I believe that using specific versions of Python is necessary for certain bindings. (I'm not positive on that, though.) If a developer has multiple versions of Maturin installed on there machine, each for a different Python version, this allows them to gracefully pick which version to use.

```shell
# Instead of
$ /path/to/python3.8/bin/maturin develop
# We could run
$ python3.8 -m maturin develop
```

Or for windows:

```shell
$ py -3.9 -m maturin develop
```

Another thing to keep in mind is that multiple popular packages use `__main__.py` as well. [PIP](https://github.com/pypa/pip/blob/main/src/pip/__main__.py), [Flask](https://github.com/pallets/flask/blob/main/src/flask/__main__.py), [Rich](https://github.com/Textualize/rich/blob/master/rich/__main__.py), and many others use it.

## Solution

I propose that Maturin exposes a `__main__.py` file in the [`maturin`](https://github.com/PyO3/maturin/tree/main/maturin) folder. This file would either call the executable file directly, or call the specific function in the Rust bindings.

## Drawbacks

While doing this would be good for compatibility, there are a few issues. The first is that the only way the Python package can access Maturin is through calling the executable with subprocess. This makes assumptions that Maturin is on the path already. If this method is used, there is no way to call a specific Python interpreter. This essentially removes the point of using `__main__` in the first place.

The fix this, `__main__` needs to figure out where the specific version's Python folder is, then call the located executable. This isn't that difficult to do, (I'm pretty sure there's a built-in function for finding the folder), but still seems overly-complicated.

On the other hand, if we didn't want to use subprocess, we'd have to ship a library file with the folder. We'd also have to abstract [`main.rs`](https://github.com/PyO3/maturin/blob/main/src/main.rs) to another file that can be imported by Python. This version probably won't work, as it requires too many breaking changes.

## Conclusion

Adding a `__main__.py` file to the Python package that can be executed with `python -m maturin` would benefit both compatibility and version-specification. This Python file would locate where binaries are installed for that specific Python version (or virtual environment), then call the Maturin executable that exists there.

Thank you for your time and consideration,
~ BD103

messense:
Closes #1007

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -385,6 +385,8 @@ jobs:
         if: steps.changes.outputs.changed == 'true'
       - run: maturin --version
         if: steps.changes.outputs.changed == 'true'
+      - run: python3 -m maturin --version
+        if: steps.changes.outputs.changed == 'true'
       - name: Upload wheel artifacts
         if: steps.changes.outputs.changed == 'true'
         uses: actions/upload-artifact@v2
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 * Add 64-bit RISC-V support by felixonmars in [#1001](https://github.com/PyO3/maturin/pull/1001)
+* Add support for invoking with `python3 -m maturin` in [#1008](https://github.com/PyO3/maturin/pull/1008)
 
 ## [0.13.0] - 2022-07-09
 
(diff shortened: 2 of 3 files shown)


messense:
The emscripten target failure is caused by https://github.com/rust-lang/rust/issues/99261

Issue #982, opened 2022-06-24
conda cannot run an aarch64 .so on an aarch64 machine


rseymour:
### Bug Description

I don't know what the bug is.  The wheel created in pyO3 and CFFI mode both seems to be incompatible with my platform.  I've tested with both, but here's the output:
```
[9:10:50] mat ❯ maturin new whatever
✔ 🤷 Which kind of bindings to use? · pyo3
  ✨ Done! New project created whatever

~/src
[9:10:54] mat ❯ cd whatever

~/src/whatever
[9:10:57] mat ❯ ls
Cargo.toml     pyproject.toml src

~/src/whatever
[9:10:58] mat ❯ maturin develop
🔗 Found pyo3 bindings
🐍 Found CPython 3.10 at /opt/homebrew/anaconda3/envs/mat/bin/python
   Compiling target-lexicon v0.12.4
   Compiling proc-macro2 v1.0.40
   Compiling once_cell v1.12.0
   Compiling unicode-ident v1.0.1
   Compiling quote v1.0.20
   Compiling libc v0.2.126
   Compiling autocfg v1.1.0
   Compiling syn v1.0.98
   Compiling parking_lot_core v0.9.3
   Compiling scopeguard v1.1.0
   Compiling cfg-if v1.0.0
   Compiling smallvec v1.8.0
   Compiling indoc v1.0.6
   Compiling unindent v0.1.9
   Compiling lock_api v0.4.7
   Compiling pyo3-build-config v0.16.5
   Compiling parking_lot v0.12.1
   Compiling pyo3-ffi v0.16.5
   Compiling pyo3 v0.16.5
   Compiling pyo3-macros-backend v0.16.5
   Compiling pyo3-macros v0.16.5
   Compiling whatever v0.1.0 (/Users/rseymour/src/whatever)
    Finished dev [unoptimized + debuginfo] target(s) in 6.55s
📦 Built wheel for CPython 3.10 to /var/folders/2c/z3xyp3hj53z0qbwqlt75tn_c0000gn/T/.tmpYipyUF/whatever-0.1.0-cp310-cp310-macosx_11_0_arm64.whl
💥 maturin failed
  Caused by: pip install in /opt/homebrew/anaconda3/envs/mat failed running ["-m", "pip", "--disable-pip-version-check", "install", "--force-reinstall"]: exit status: 1
--- Stdout:

--- Stderr:
[...]

Issue #1013, opened 2022-07-19
Can't compile using `maturin develop` on macOS(M2 chip)


dzvon:
### Bug Description

I tried follow up the example(string_num) in https://github.com/PyO3/pyo3. It shows a bunch of errors when I run `maturin develop`. Here is a screenshot of the error, not all of them.

![image](https://user-images.githubusercontent.com/3402811/179683455-37c553cc-bb9c-44b8-85f3-89e5774da8db.png)

The most important information here is `x86_64-apple-darwin`, why the compile target is `x86_64-apple-darwin`? I run those code on macOS with M2 chip, the target should be `aarch64-apple-darwin`.

BTW, I also tried `maturin develop --target aarch64-apple-darwin`, unfortunately, it doesn't work.

### Your Python version (`python -V`)

Python 3.8.9

### Your pip version (`pip -V`)

pip 22.1.2

### What bindings you're using

pyo3

### Does `cargo build` work?

- [X] Yes, it works

### If on windows, have you checked that you aren't accidentally using unix path (those with the forward slash `/`)?

- [X] Yes

### Steps to Reproduce

https://github.com/PyO3/pyo3#using-rust-from-python

messense:
Looks like you are using a x86_64 Python virtualenv. What's the output of `python3 -c 'import platform; print(platform.platform())'`?

dzvon:
```
Python 3.8.9 (default, Jul 19 2021, 09:37:30)
[Clang 13.0.0 (clang-1300.0.27.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import sysconfig; print(sysconfig.get_platform(), end='')
macosx-10.14.6-arm64>>>
```

messense:
What's the output of `rustc -vV`?

dzvon:
```
➜  ~ rustc -vV
rustc 1.62.0 (a8314ef7d 2022-06-27)
binary: rustc
commit-hash: a8314ef7d0ec7b75c336af2c9857bfaf43002bfc
commit-date: 2022-06-27
host: aarch64-apple-darwin
release: 1.62.0
LLVM version: 14.0.5
```

dzvon:
Gotcha, my scenario is arm64 Python + aarch64-apple-darwin, but the factual compile target is `x86_64-apple-darwin`

messense:
See https://github.com/PyO3/maturin/pull/1015#discussion_r924166748

messense:
[...]

Issue #979, opened 2022-06-21
python target is not used on M1


twiecki:
### Bug Description

```
>>maturin develop
🍹 Building a mixed python/rust project
🔗 Found pyo3 bindings
🐍 Found CPython 3.10 at /Users/twiecki/miniforge3/envs/pymc4/bin/python
📡 Using build options bindings from pyproject.toml
   Compiling autocfg v1.1.0
   Compiling cfg-if v1.0.0
   Compiling once_cell v1.12.0
   Compiling libc v0.2.126
   Compiling proc-macro2 v1.0.39
   Compiling target-lexicon v0.12.4
   Compiling unicode-ident v1.0.1
   Compiling syn v1.0.96
   Compiling scopeguard v1.1.0
   Compiling crossbeam-utils v0.8.9
   Compiling libm v0.2.2
   Compiling parking_lot_core v0.9.3
   Compiling rayon-core v1.9.3
   Compiling ppv-lite86 v0.2.16
   Compiling smallvec v1.8.0
   Compiling crossbeam-queue v0.3.5
   Compiling rawpointer v0.2.1
   Compiling either v1.6.1
   Compiling unindent v0.1.9
   Compiling indoc v1.0.6
   Compiling num-traits v0.2.15
   Compiling memoffset v0.6.5
   Compiling crossbeam-epoch v0.9.9
   Compiling lock_api v0.4.7
   Compiling num-integer v0.1.45
   Compiling rayon v1.5.3
   Compiling matrixmultiply v0.3.2
   Compiling itertools v0.10.3
   Compiling getrandom v0.2.7
   Compiling num_cpus v1.13.1
   Compiling rand_core v0.6.3
   Compiling quote v1.0.18
   Compiling crossbeam-channel v0.5.5
   Compiling pyo3-build-config v0.16.5
   Compiling parking_lot v0.12.1
   Compiling rand_chacha v0.3.1
   Compiling crossbeam-deque v0.8.1
   Compiling crossbeam v0.8.1
   Compiling rand v0.8.5
   Compiling num-complex v0.4.1
   Compiling ndarray v0.15.4
   Compiling rand_distr v0.4.3
   Compiling pyo3-ffi v0.16.5
   Compiling pyo3 v0.16.5
   Compiling pyo3-macros-backend v0.16.5
   Compiling thiserror-impl v1.0.31
   Compiling multiversion-macros v0.6.1
   Compiling pyo3-macros v0.16.5
   Compiling multiversion v0.6.1
[...]

PR #1015, opened 2022-07-19
fix: `maturin develop` can compile on macOS with M1/M2 chip

dzvon:
#1013 

This fix can make `maturin develop` compile successfully in my local environment.

### DIFF
--- a/src/develop.rs
+++ b/src/develop.rs
@@ -35,9 +35,9 @@ pub fn develop(
         Ok(output) if output.status.success() => {
             let platform = String::from_utf8_lossy(&output.stdout);
             if platform.contains("macos") {
-                if platform.contains("x86_64") && target.target_arch() != Arch::X86_64 {
+                if platform.contains("x86_64") && target.target_arch() == Arch::X86_64 {
                     target_triple = Some("x86_64-apple-darwin".to_string());
-                } else if platform.contains("arm64") && target.target_arch() != Arch::Aarch64 {
+                } else if platform.contains("arm64") && target.target_arch() == Arch::Aarch64 {
                     target_triple = Some("aarch64-apple-darwin".to_string());
                 }
             }
@@ -48,7 +48,7 @@ pub fn develop(
     // Store wheel in a unique location so we don't get name clashes with parallel runs
     let wheel_dir = TempDir::new().context("Failed to create temporary directory")?;
 
-    let build_options = BuildOptions {
+    let mut build_options = BuildOptions {
         platform_tag: vec![PlatformTag::Linux],
         interpreter: vec![python.clone()],
         find_interpreter: false,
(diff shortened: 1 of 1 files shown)


src/develop.rs
```diff
@@ -63,6 +63,11 @@ pub fn develop(
         },
     };
 
+    // We should respect user input.
+    if cargo_options.target.is_some() {
+        build_options.cargo.target = cargo_options.target;
```
messense, line 68: Good.

src/develop.rs
```diff
@@ -35,9 +35,9 @@ pub fn develop(
         Ok(output) if output.status.success() => {
             let platform = String::from_utf8_lossy(&output.stdout);
             if platform.contains("macos") {
-                if platform.contains("x86_64") && target.target_arch() != Arch::X86_64 {
+                if platform.contains("x86_64") && target.target_arch() == Arch::X86_64 {
```
messense, line 38: L38 and L40 change is wrong. We want to respect Python architecture to fix #982 and #979. `target` is default Rust target we get from `rustc -vV`.

messense (changes requested):

src/develop.rs
dzvon: > Then the offending code shouldn't be these two I think? It might be related to the `target_triple` variable, can you try #1016 ?

I've tried #1016, same error.

dzvon: But if I append `--target aarch64-apple-darwin`, it works fine.

messense: Try add `println!("{:#?}", build_options);` just above https://github.com/PyO3/maturin/blob/86b5dfc77254da2d01cd48a8d4128a15659d1730/src/develop.rs#L68

dzvon:
```
BuildOptions {
    platform_tag: [
        Linux,
    ],
    interpreter: [
        "/Users/pvon/Documents/string_sum/.env/bin/python",
    ],
    find_interpreter: false,
    bindings: None,
    out: Some(
        "/var/folders/0r/tkx69nhn6vg6sr0w1c6nnsxr0000gn/T/.tmpAolWXa",
    ),
    skip_auditwheel: false,
    zig: false,
    universal2: false,
    cargo: CargoOptions {
        quiet: false,
        jobs: None,
        profile: None,
        features: [],
        all_features: false,
        no_default_features: false,
        target: None,
        target_dir: None,
        manifest_path: None,
        ignore_rust_version: false,
        verbose: 0,
        color: None,
        frozen: false,
        locked: false,
        offline: false,
        config: [],
        unstable_flags: [],
        timings: None,
        future_incompat_report: false,
        args: [],
    },
}
```

messense: Nothing seems wrong, I can't reproduce this issue, so you have to debug on your own.

PR #1038, opened 2022-08-04
Fix path search upon `python3 -m  maturin`

bhrollins:
`sysconfig.get_path` was insufficient on my machine as it was unable to resolve my `scripts` directory without an explicit `scheme="nt_user"` invocation (Windows 11 running Python 3.10.6). Intention here was to implement a more generic method of resolving `scripts` directories. There may be more idiomatic ways to achieve this behavior, as well as more useful error messages than what are provided in this solution.

### DIFF
--- a/maturin/__main__.py
+++ b/maturin/__main__.py
@@ -2,8 +2,37 @@ import os
 import sys
 from pathlib import Path
 import sysconfig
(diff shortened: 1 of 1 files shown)


messense (approved):
Thanks!

PR #1041, opened 2022-08-06
Run default CI workflow on main/master push only

### LINKED ISSUE #1040: Default CI.yml results in duplicate workflow runs
### Bug Description

Having maturin put a [default CI.yml](https://github.com/PyO3/maturin/blob/2876dbecbf7df1eb8314570fa0605f57d4e7c737/src/templates/CI.yml.j2) into newly initialized projects' `.github/workflows` is a great idea, but because it specifies

```yaml
on:
  push:
  pull_request:
```

for the entire workflow, it causes duplicate runs for pull requests: once for the push and once for the PR.

Wouldn't it make sense to default to something like

```yaml
on:
  push:
    branches:
      - main
  pull_request:
```

like in [your own `test.yml`](https://github.com/PyO3/maturin/blob/2876dbecbf7df1eb8314570fa0605f57d4e7c737/.github/workflows/test.yml) instead?

If the concern was that not everyone's main branch would be called `main`, e.g. when initializing a project in an old repository that still uses `master`, I'd say that it's very unlikely considering the _vast_ majority of use cases for `maturin init` will be brand new projects, and the ones that aren't will probably have existing CI configs anyway. But if that's not good enough, there could be a check whether maturin is run inside an existing repository and if it has a `master` or `main` branch.

Thoughts?

irrelevant

irrelevant

_No response_

### Does `cargo build` work?

- [X] Yes, it works

- [X] Yes

### Steps to Reproduce

1. Initialize a project using `maturin init` (or `maturin new`).
2. Check the generated files including `.github/workflows/CI.yml` into the project's Git repository, create a GitHub project for it and push it.
3. Create a merge request and observe that the workflow specified in `CI.yml` executes twice.

smheidrich:
As suggested by @messense in https://github.com/PyO3/maturin/issues/1040#issuecomment-1207180707.

Fixes #1040.

### DIFF
--- a/src/templates/CI.yml.j2
+++ b/src/templates/CI.yml.j2
@@ -2,6 +2,9 @@ name: CI
 
 on:
   push:
+    branches:
+      - main
+      - master
   pull_request:
 
 jobs:


messense (approved):
Thanks!

Issue #1070, opened 2022-08-30
The dependency rpassword 7 means a nightly rustc


woutervb:
### Bug Description

When trying to create a wheel on Ubuntu Jammy (22.04) & Ubuntu Focal (20.04), it fails with an error that rpassword can't be build `namespaced features with the `dep:` prefix are only allowed on the nightly channel and requires the `-Z namespaced-features` flag on the command-line`

```PIP_NO_CACHE_DIR=1 PIP_WHEEL_DIR=build env/bin/pip wheel /srv/sdists/maturin-0.13.2.tar.gz
  
  × Building wheel for maturin (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [39 lines of output]
      running bdist_wheel
      running build
      running build_py
      creating build
      creating build/lib.linux-x86_64-cpython-310
      creating build/lib.linux-x86_64-cpython-310/maturin
      copying maturin/import_hook.py -> build/lib.linux-x86_64-cpython-310/maturin
      copying maturin/__init__.py -> build/lib.linux-x86_64-cpython-310/maturin
      copying maturin/__main__.py -> build/lib.linux-x86_64-cpython-310/maturin
      running egg_info
      creating maturin.egg-info
      writing maturin.egg-info/PKG-INFO
      writing dependency_links to maturin.egg-info/dependency_links.txt
      writing requirements to maturin.egg-info/requires.txt
      writing top-level names to maturin.egg-info/top_level.txt
      writing manifest file 'maturin.egg-info/SOURCES.txt'
      reading manifest file 'maturin.egg-info/SOURCES.txt'
      reading manifest template 'MANIFEST.in'
      writing manifest file 'maturin.egg-info/SOURCES.txt'
      running build_ext
      running build_rust
      cargo build --manifest-path Cargo.toml --message-format=json-render-diagnostics --release -v
      error: failed to get `anyhow` as a dependency of package `maturin v0.13.2 (/tmp/pip-req-build-pp0gr68x)`
      
      Caused by:
        failed to load source for dependency `anyhow`
      
      Caused by:
        Unable to update registry `crates-io`
      
      Caused by:
        failed to update replaced source registry `crates-io`
      
      Caused by:
        failed to parse manifest at `/srv/rust_vendored/rpassword-7.0.0/Cargo.toml`
      
      Caused by:
        namespaced features with the `dep:` prefix are only allowed on the nightly channel and requires the `-Z namespaced-features` flag on the command-line
[...]

PR #1071, opened 2022-08-30
Set MSRV to 1.59 and test on CI

messense:
Rust 1.59.0 is the packaged version on Ubuntu 22.04 LTS.

Ref https://github.com/PyO3/maturin/issues/1070

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -399,3 +399,21 @@ jobs:
         with:
           name: wheels
           path: dist
+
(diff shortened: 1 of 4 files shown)


konstin (approved):
i'd still encourage people to use rustup as old rustc versions are not supported anymore, but supporting 1.59 shouldn't be too bad. I've also opened https://github.com/conradkleinespel/rpassword/issues/73 so we can remove the rpassword pin

PR #1084, opened 2022-09-06
Fix sdist build for optional path dependencies

### LINKED ISSUE #1083: maturin fails if a local dependency is not released on `crates.io`
### Bug Description

I am not entirely certain, but I believe it is related to a subdependency not being published to crates.io.

If I comment out [this line](https://github.com/pola-rs/polars/blob/c8cfc9957de9f56a3c7a9a25752cb09ab80c687d/py-polars/Cargo.toml#L87) and try to build a source distribution with maturin it fails.

The dependency is not required for the python bindings and is not yet released on crates.io.

This only happens if we try to build a source distro.

Python 3.9.12

pip 21.2.4

pyo3

### Does `cargo build` work?

- [X] Yes, it works

- [ ] Yes

### Steps to Reproduce

Comment out this [this line](https://github.com/pola-rs/polars/blob/c8cfc9957de9f56a3c7a9a25752cb09ab80c687d/py-polars/Cargo.toml#L87) on: https://github.com/pola-rs/polars/commit/c8cfc9957de9f56a3c7a9a25752cb09ab80c687d

and then run: `$ maturin build --sdist`.

```
📦 Including license file "LICENSE"
🔗 Found pyo3 bindings with abi3 support for Python ≥ 3.7
🐍 Not using a specific python interpreter
💥 maturin failed
  Caused by: Failed to add local dependency polars at /home/ritchie46/code/polars/polars/Cargo.toml to the source distribution
  Caused by: cargo metadata does not know about the path for dependencies.polars-algo present in /home/ritchie46/code/polars/polars/Cargo.toml, which should never happen ಠ_ಠ
```

My maturin version is `0.13.2`. Sorry that I don't have an MWE other than an existing project. A bit hard to publish crates for this.

messense:
Fixes #1083 by also bundling optional path dependencies.

### DIFF
--- a/src/source_distribution.rs
+++ b/src/source_distribution.rs
@@ -1,7 +1,7 @@
 use crate::module_writer::{add_data, ModuleWriter};
 use crate::{Metadata21, SDistWriter};
 use anyhow::{bail, Context, Result};
-use cargo_metadata::{Metadata, PackageId};
+use cargo_metadata::Metadata;
 use fs_err as fs;
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
@@ -9,12 +9,6 @@ use std::process::Command;
 use std::str;
 
 const LOCAL_DEPENDENCIES_FOLDER: &str = "local_dependencies";
-
-#[derive(Debug, Clone)]
-struct PathDependency {
-    id: PackageId,
-    path: PathBuf,
-}
 
 /// We need cargo to load the local dependencies from the location where we put them in the source
 /// distribution. Since there is no cargo-backed way to replace dependencies
@@ -25,7 +19,7 @@ struct PathDependency {
 /// This method is rather frail, but unfortunately I don't know a better solution.
 fn rewrite_cargo_toml(
     manifest_path: impl AsRef<Path>,
-    known_path_deps: &HashMap<String, PathDependency>,
+    known_path_deps: &HashMap<String, PathBuf>,
     root_crate: bool,
 ) -> Result<String> {
     let text = fs::read_to_string(&manifest_path).context(format!(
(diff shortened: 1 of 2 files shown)


messense:
This does not work actually, build from sdist fails

```bash
$ cargo build
error: failed to load manifest for dependency `polars`

Caused by:
  failed to load manifest for dependency `polars-algo`

Caused by:
  failed to read `/Users/messense/Projects/maturin/dist/polars-0.14.9/local_dependencies/polars/polars-algo/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
```

src/source_distribution.rs
```diff
@@ -58,15 +58,16 @@ fn rewrite_cargo_toml(
...
                 }
-                // This is the location of the targeted crate in the source distribution
-                table[&dep_name]["path"] = if root_crate {
-                    toml_edit::value(format!("{}/{}", LOCAL_DEPENDENCIES_FOLDER, dep_name))
-                } else {
-                    // Cargo.toml contains relative paths, and we're already in LOCAL_DEPENDENCIES_FOLDER
-                    toml_edit::value(format!("../{}", dep_name))
-                };
-                rewritten = true;
                 if !known_path_deps.contains_key(&dep_name) {
```
messense, line 61: Because of https://github.com/PyO3/maturin/pull/1084#issuecomment-1237686996, I think we need to somehow add the optional path deps to `known_path_deps` even if they're not activated via Cargo features.

messense: Unfortunately there isn't enough metadata:

```json
"dependencies": [
      {
          "name": "polars-algo",
          "source": null,
          "req": "^0.23.0",
          "kind": null,
          "rename": null,
          "optional": true,
          "uses_default_features": true,
          "features": [],
          "target": null,
          "registry": null,
          "path": "/Users/messense/Projects/polars/polars/polars-algo"
      }
]
```

and the optional `polars-algo` dep isn't present at `packages`.

messense: This reminds me of https://github.com/PyO3/maturin/pull/720#issuecomment-983423640, we might be able to workaround this by using `dependencies` instead of the structured `dep`.

https://github.com/koehlma/maturin/blob/dee86e60b1771055210d04fff5d8509d80a0fe04/src/source_distribution.rs#L146-L175

PR #1085, opened 2022-09-06
Prettify repo

kianmeng:
Resolved by `prettier -w .`


konstin:
please don't run random tools over others' repository, for many files there are specific reasons why they are formatted they way they are. Additionally, if you'd like to introduce a tool this would need integration into ci and into the dev setup

kianmeng:
@konstin Noted, and thanks for letting me know.

PR #1107, opened 2022-09-15
Build platform independent rust binary wheel through wasi

konstin:
I had this idea that besides having to build a wheel for each platform and then still there are platforms without manylinux supports, we could have a single wasm wheel as a generic fallback. Currently more theoretical since wasmtime supports less platforms than manylinux. Also we could have isolation and such in the future.

We build the rust binaries to wasi and put them in a wheel with a wasmtime launcher. The resulting wheels are platform independent (py3-none-any) and run on all wasmtime platforms. 

TODO:

- [x] figure out what names are valid for binaries (esp. with dots) and how to translate them for the entrypoints scripts: cargo is stricter than pypa already
- [x] can we ship wasm entrypoints and a user python module: maybe later
- [x] https://github.com/bytecodealliance/wasmtime-py/issues/94: they don't really know either

### DIFF
--- a/src/module_writer.rs
+++ b/src/module_writer.rs
@@ -761,7 +761,7 @@ pub fn write_bin(
     writer: &mut impl ModuleWriter,
     artifact: &Path,
     metadata: &Metadata21,
-    bin_name: &OsStr,
+    bin_name: &str,
 ) -> Result<()> {
     let data_dir = PathBuf::from(format!(
         "{}-{}.data",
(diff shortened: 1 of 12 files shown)


src/module_writer.rs
```diff
@@ -759,6 +759,58 @@ pub fn write_bin(
...
+    wasi.inherit_stdin()
+    store.set_wasi(wasi)
+    linker = Linker(engine)
+    linker.define_wasi()
+    module = Module.from_file(store.engine, str(program_location))
+    linking1 = linker.instantiate(store, module)
+    # TODO: this is taken from https://docs.wasmtime.dev/api/wasmtime/struct.Linker.html#method.get_default
+    #       is this always correct?
+    start = linking1.exports(store).get("") or linking1.exports(store)["_start"]
+    start(store)
```
messense, line 798: I remember that wasmtime execution is sandboxed, so you might need to give access to file systems explicitly somehow?

messense (commented):
Interesting idea!

src/target.rs
```diff
@@ -28,6 +28,7 @@ pub enum Os {
     Illumos,
     Haiku,
     Emscripten,
+    Wasi,
```
messense, line 31: One concern though, in the future if Python gained support for `wasm32-wasi` for extension modules, it will conflicts with this wasmtime based one.

https://github.com/python/cpython/blob/main/Tools/wasm/README.md#wasi-wasm32-wasi

messense: Oh, since this is only for bin, it is not an issue.

tests/run.rs
```diff
@@ -209,16 +218,48 @@ fn integration_pyo3_ffi_pure() {
...
+// Sourced from https://pypi.org/project/wasmtime/0.40.0/#files
+// update with wasmtime updates
+#[cfg(any(
+    all(target_os = "windows", target_arch = "x86_64"),
+    all(
+        target_os = "linux",
+        any(target_arch = "x86_64", target_arch = "aarch64")
+    ),
+    all(
+        target_os = "max",
```
messense, line 246: I think you mean:

```suggestion
        target_os = "macos",
```

src/module_writer.rs
konstin: i think so, let's see if i get a response in https://github.com/bytecodealliance/wasmtime-py/issues/94 and otherwise i'll have to collect some real world use cases compatible with wasm

tests/run.rs
konstin: oops yes totally blind

messense:
`wasm32-wasi` std library also needs to be installed via `rustup`.

konstin:
yet that's the next step

messense:
Add `rustup target add wasm32-wasi` to 

https://github.com/PyO3/maturin/blob/fb130c951d780578a3b0a58666aa33f704de68fb/.cirrus.yml#L51

and https://github.com/PyO3/maturin/blob/fb130c951d780578a3b0a58666aa33f704de68fb/.cirrus.yml#L76

should resolve the test failures on Cirrus CI.

konstin:
ready to merge from my end

test-crates/cargo-mock/src/main.rs
```diff
@@ -33,7 +33,8 @@ fn run() -> Result<()> {
         .replace("--quiet", "")
         .replace(&cwd, "")
         .replace(" ", "-")
-        .replace("/", "-");
+        .replace("/", "-")
+        .replace("-----C-link-arg=-s", "");
```
messense, line 37: Why do we need to do this? I don't quite understand why it starts with so many `-`s.

konstin: I think it's `" -- -C link-arg=-s".replace(' ', '-')`. The string got too long for a filename so' i'm cutting stuff out; this could be encoded much nicer but i think for an optional test dependency it's not worth it

messense (approved):

PR #1170, opened 2022-10-06
Include `Cargo.lock` by default in source distribution

messense:
* Required if `--locked`/`--frozen` is set
* Include by default when exists, otherwise warn

https://github.com/samuelcolvin/watchfiles/issues/192#issuecomment-1269313496 cc @samuelcolvin


samuelcolvin:
LGTM 👍

PR #1185, opened 2022-10-10
Add support for python first src layout - take 2

messense:
Implements https://github.com/PyO3/maturin/pull/501#issuecomment-845343101


src/project_layout.rs
```diff
@@ -166,8 +181,10 @@ impl ProjectResolver {
         // use command line argument if specified
         if let Some(path) = cargo_manifest_path {
             let workspace_root = Self::resolve_cargo_metadata(&path, cargo_options)?.workspace_root;
+            let workspace_parent = workspace_root.parent().unwrap_or(&workspace_root);
             for parent in fs::canonicalize(&path)?.ancestors().skip(1) {
-                if !parent.starts_with(&workspace_root) {
+                // Allow looking outside to the parent directory of Cargo workspace root
```
messense, line 186: I'm a bit worried about this. It's for supporting `maturin build --manifest-path /path/to/rust/Cargo.toml` when `pyproject.toml` is located in the parent directory of the rust crate.

But it might have unexpected side effects, I'm not sure.

konstin (approved):
:100:

src/project_layout.rs
konstin: i think i'd make `--manifest-path` take either Cargo.toml or pyproject.toml, whichever is in the project root. that would be an unexpected break from the normal cargo semantics, but i think it makes sense overall to point to the most top level configuration file

messense: > make `--manifest-path` take either Cargo.toml or pyproject.toml, whichever is in the project root

It's certainly doable but not as intuitive as just `Cargo.toml`, we can explore it later.

PR #1187, opened 2022-10-10
Fix sdist build failure when using python first `src` layout

### LINKED ISSUE #1186: Failed to build sdist when using python first `src` layout
#1185 was merged too soon, opening an issue to track this.

```bash
$ maturin sdist -o dist
🔗 Found pyo3 bindings
🐍 Found CPython 3.10 at /Users/messense/.pyenv/versions/3.10.6/bin/python3
💥 maturin failed
  Caused by: Failed to add 340 bytes to sdist as /Users/messense/Projects/maturin/test-crates/pyo3-mixed-src/rust/Cargo.toml
  Caused by: paths in archives must be relative when setting path for
```

It's actually a general issue when `Cargo.toml` isn't located at the top level directory since we use `cargo package --list --allow-dirty` to get the list of files that should be included in sdist.

messense:
Fixes #1186

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ venv*/
 .pytest_cache/
 .tox/
 *.so
+*.py[cdo]
 __pycache__/
 *.egg-info/
 *.egg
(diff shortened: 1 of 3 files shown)


messense:
Odd, I can't reproduce the failure in Alpine Linux container.

PR #1198, opened 2022-10-14
Fix sdist build for maturin itself

messense:
```
cargo run sdist
   Compiling maturin v0.13.6 (/Users/messense/Projects/maturin)
    Finished dev [unoptimized + debuginfo] target(s) in 2.28s
     Running `target/debug/maturin sdist`
⚠️  Warning: `build-backend` in pyproject.toml is not set to `maturin`, packaging tools such as pip will not use maturin to build this project.
🔗 Found bin bindings
💥 maturin failed
  Caused by: prefix not found
```

### DIFF
--- a/src/source_distribution.rs
+++ b/src/source_distribution.rs
@@ -511,6 +511,7 @@ pub fn source_distribution(
                 debug!("Ignoring {}", source.display());
                 continue;
             }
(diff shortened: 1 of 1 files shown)


messense:
Will do a refactoring later to unify relative/absolute path handling related to `strip_prefix`.

PR #1214, opened 2022-10-26
Configure bors

messense:


messense:
```
Cause: Check install fail: exit status: 1 
--- Stdout:

--- Stderr:
  File "/home/runner/work/maturin/maturin/test-crates/pyo3-pure/check_installed/check_installed.py", line 5, in <module>
    import pyo3_pure
  File "/usr/share/miniconda/envs/maturin-develop-pyo3-pure-conda/lib/python3.10/site-packages/pyo3_pure/__init__.py", line 1, in <module>
    from .pyo3_pure import *
ImportError: /usr/share/miniconda/envs/maturin-develop-pyo3-pure-conda/lib/python3.10/site-packages/pyo3_pure/pyo3_pure.abi3.so: undefined symbol: PyPyLong_FromSsize_t

thread 'develop_pyo3_pure_conda' panicked at 'Check install fail: exit status: 1 
--- Stdout:

--- Stderr:
  File "/home/runner/work/maturin/maturin/test-crates/pyo3-pure/check_installed/check_installed.py", line 5, in <module>
    import pyo3_pure
  File "/usr/share/miniconda/envs/maturin-develop-pyo3-pure-conda/lib/python3.10/site-packages/pyo3_pure/__init__.py", line 1, in <module>
    from .pyo3_pure import *
ImportError: /usr/share/miniconda/envs/maturin-develop-pyo3-pure-conda/lib/python3.10/site-packages/pyo3_pure/pyo3_pure.abi3.so: undefined symbol: PyPyLong_FromSsize_t
```

Odd, looks like pyo3 does not rebuild after switching from PyPy to conda.

messense:
> Odd, looks like pyo3 does not rebuild after switching from PyPy to conda.

This one is a bit different than #1220 I think.

https://github.com/PyO3/pyo3/blob/8e8b48416940f1ca5899b5c949fd1e24a9211dd8/pyo3-build-config/src/impl_.rs#L1686-L1690

Perhaps the way that `cargo:rerun-if-env-changed` instructions are set in `pyo3-build-config` is problematic because they are only conditionally set? But I think `PYO3_PYTHON` value should have changed which should result in a rebuild, although it seems that didn't happen. I need to reproduce it locally to investigate further.

cc @davidhewitt

messense:
bors try

davidhewitt:
Hmm. Nothing obvious sticks out to me. I agree with you that it looks like these should be configuring ok. Does the `cargo nextest` run each build with a separate target directory and/or in sequence (i.e. not in parallel)?

I wonder... is there some nesting of virtualenvs / conda going on? It might be that PyO3 is being forced to ignore e.g. https://github.com/PyO3/pyo3/blob/8e8b48416940f1ca5899b5c949fd1e24a9211dd8/pyo3-build-config/src/impl_.rs#L1669-L1672

... but if maturin is setting `PYO3_PYTHON` then I don't think this code path should ever run.

messense:
> I wonder... is there some nesting of virtualenvs / conda going on?

FYI, the Python 3.10 issue dates back to https://github.com/PyO3/maturin/pull/742#issuecomment-994817705

messense:
> ImportError: dlopen(/Users/runner/work/maturin/maturin/test-crates/venvs/integration-pyo3-mixed-py-subdir-py3.9-pypy/lib/pypy3.9/site-packages/pyo3_mixed_py_subdir/_pyo3_mixed.pypy39-pp73-darwin.so, 6): Symbol not found: _PyModule_GetFilenameObject

Looks like a bug in pyo3?

* https://github.com/PyO3/pyo3/blob/d7b05cbcf5785019a097e496454b924f3e11d94f/pyo3-ffi/src/moduleobject.rs#L35
* https://github.com/PyO3/pyo3/blob/bc8641c790b22b7c456d248f8d672279aed078e5/src/types/module.rs#L206-L213

messense:
bors try

messense:
bors retry

messense:
bors try

PR #1232, opened 2022-11-01
Use generic tags when `sys.implementation.name` != `platform.python_implementation()`

### LINKED ISSUE #1230: support alternative implementations of CPython API
My problem is that a wheel generated by, for example`pyston3.8 -m pip wheel orjson`, has the wrong `implementation` and `abi` tags.  (_Pyston_ is a CPython fork with compatible API, but incompatible ABI.)  Having the wrong tags, the wheel can't be put into a package repository and be discovered properly.   (To be clear, install via `pyston3.8 -m pip install orjson` does install a correctly working package.  The issue reported here is about incorrect tags on the generated wheel file.)

As far as tagging of wheel files with `implementation` and `abi`, the maturin tool seems to only have specific, hard-coded support for CPython and PyPy implementations.

For example, here is a correctly tagged wheel file when compiling numpy with pyston:
```
$ pyston3.8 -m pip wheel numpy
--> numpy-1.23.4-pyston38-pyston_23_x86_64_linux_gnu-linux_x86_64.whl
```
and here is the incorrectly tagged wheel file produced for orjson by maturin:
```
$ pyston3.8 -m pip wheel orjson
--> orjson-3.6.7-cp38-cp38-linux_x86_64.whl
```

The maturin seems to fall back to the tags of CPython, rather than use the information available in the pip / bdist / build environment.  (Perhaps it's a bug that maturin doesn't explicitly fail the build, rather than produce a wheel with the wrong tags.)

I'm hoping that maturin can deal with arbitrary API-compatible CPython implementations, without having a hard-coded, centralized interpreter mapping.  Aren't the true strings for `implementation` and `abi` available from the environment that invoked maturin?

https://github.com/PyO3/maturin/blob/297db1f693cc9f54159d35eb7748ca7fea477b94/src/python_interpreter/mod.rs#L285-L295

messense:
Closes #1230


src/python_interpreter/mod.rs
```diff
@@ -350,6 +352,10 @@ pub struct PythonInterpreter {
     /// When cross compile the target interpreter isn't runnable,
     /// and it's `executable` is empty
     pub runnable: bool,
+    /// Comes from `sys.platform.name`
+    pub implmentation_name: String,
+    /// Comes from sysconfig var `SOABI`
+    pub soabi: Option<String>,
```
messense, line 358: Ideally `implmentation_name` and `soabi` should be moved to `InterpreterConfig`, but that requires updating all of the bundled `sysconfig` files, we can do that later.

messense:
```
FAIL [  17.167s] maturin::run develop_pyo3_ffi_pure
FAIL [  56.923s] maturin::run develop_pyo3_pure
FAIL [  11.939s] maturin::run editable_pyo3_ffi_pure
FAIL [  46.915s] maturin::run editable_pyo3_pure
FAIL [  13.128s] maturin::run integration_pyo3_ffi_pure
FAIL [  43.587s] maturin::run integration_pyo3_pure
```

Note: All of the abi3 wheel build tests failed on Pyston since it does have support for stable abi yet.

PR #1228, opened 2022-10-30
CI: use latest Python 3.10 instead of 3.10.0

messense:

PR #1237, opened 2022-11-02
Revamp GitHub Actions cache keys

messense:


.github/workflows/test.yml
```diff
@@ -147,7 +147,7 @@ jobs:
       - name: Cache cargo build
         uses: Swatinem/rust-cache@v2
         with:
-          key: maturin-${{ runner.os }}
+          shared-key: maturin-build
       - name: Cache test crates cargo build
         uses: actions/cache@v3
```
messense, line 152: It'd be great if we can drop `${{ matrix.python-version }}` from cache key once https://github.com/PyO3/maturin/pull/1228#issuecomment-1296276455 is fixed on pyo3 side.

PR #1238, opened 2022-11-02
Reduce CI time by skipping on non-code changes and reduce PR jobs

messense:


messense:
Not sure why, but `Swatinem/rust-cache@v2` seems to generate unstable cache keys.

PR #1251, opened 2022-11-05
Set default macOS deployment target version if not specified

### LINKED ISSUE #1247: macOS-12 built wheel incompatible with macOS-11, but Still has macosx_11_0 tag
macOS-12 packaging environment gives `macosx_10_9_x86_64.macosx_11_0_arm64.macosx_10_9_universal2.whl`, which turns out incompatible with macOS 11 because of `Symbol not found` in `/usr/lib/libc++.1.dylib`.

I think perhaps wheels built on macOS-12 should be tagged `macosx_12_0`, and therefore not to be downloaded by pypi to macOS-11.

messense:
Fixes #1247


messense:
bors retry

PR #1253, opened 2022-11-07
Add `ignore_git_ignore` flag to build options

mbrobbel:
This can be set to include files that are ignored by `.gitignore` files. This was added in https://github.com/PyO3/maturin/pull/695.

This may be helpful when Python modules are generated as part of a build. The module should be included in the wheel, but not in the git repo which is why it's in the `.gitignore` file.

### DIFF
--- a/src/build_context.rs
+++ b/src/build_context.rs
@@ -178,6 +178,8 @@ pub struct BuildContext {
     pub editable: bool,
     /// Cargo build options
     pub cargo_options: CargoOptions,
+    /// Ignore `.gitignore` files when including files for the Python part of a mixed project
+    pub ignore_git_ignore: bool,
 }
 
 /// The wheel file location and its Python version tag (e.g. `py3`).
@@ -474,6 +476,7 @@ impl BuildContext {
             None,
             &self.target,
             self.editable,
+            self.ignore_git_ignore,
         )
         .context("Failed to add the files to the wheel")?;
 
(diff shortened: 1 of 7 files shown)


messense:
I don't believe this is the way to go. Excluding files via `gitignore` is a sensible default, simply disable it may cause more issues.

For sdist we already have `sdist-include` configuration, but there isn't one for wheel yet. Something like the [include and exclude configuration](https://python-poetry.org/docs/pyproject/#include-and-exclude) of Poetry would be better IMO.

mbrobbel:
I totally agree with the current default behavior to excludes files listed in ignore files, and the added flag defaults to `false` which makes this a non-breaking change. Maybe the name is a bit confusing (double negative), but setting `ignore_git_ignore` to `true` disables excluding files via `gitignore`.

> For sdist we already have sdist-include configuration, but there isn't one for wheel yet. Something like the [include and exclude configuration](https://python-poetry.org/docs/pyproject/#include-and-exclude) of Poetry would be better IMO.

Are you suggesting to add that to maturin?
```toml
[tool.maturin]
include = []
exclude = []
```
I could try to implement that. Should we keep the `sdist-include` or (breaking change) copy the `format` key behavior from Poetry and deprecate the `sdist-include` configuration?

messense:
> Are you suggesting to add that to maturin?

Yes

> Should we keep the `sdist-include` or (breaking change) copy the `format` key behavior from Poetry and deprecate the `sdist-include` configuration?

We should first deprecate it and print a warning when it appears in configuration, we can then remove `sdist-include` in the next breaking change release to give users time to migrate to the new format.

PR #1259, opened 2022-11-10
Ignore sdist tar ball instead of error out

messense:
Fixes https://github.com/PyO3/maturin-action/issues/99


messense:
Alpine Linux edge `py3-pip` package has some problem.

PR #1258, opened 2022-11-09
Add sysconfig info for PyPy on ppc64le

mbargull:
This is based on the `python -m sysconfig` output from conda-forge's `pypy3.*` builds:
<details><summary>sysconfig for 3.7</summary>

```
> (conda activate linux-ppc64le-pypy3.7 && conda list '^(python|pypy.*)$' && python -m sysconfig)
# packages in environment at /opt/conda/envs/linux-ppc64le-pypy3.7:
#
# Name                    Version                   Build  Channel
pypy3.7                   7.3.7                hb71c00f_2    conda-forge
python                    3.7.12                0_73_pypy    conda-forge
Platform: "linux-ppc64le"
Python version: "3.7"
Current installation scheme: "pypy"

Paths: 
	data = "/opt/conda/envs/linux-ppc64le-pypy3.7"
	include = "/opt/conda/envs/linux-ppc64le-pypy3.7/include"
	platinclude = "/opt/conda/envs/linux-ppc64le-pypy3.7/include"
	platlib = "/opt/conda/envs/linux-ppc64le-pypy3.7/site-packages"
	platstdlib = "/opt/conda/envs/linux-ppc64le-pypy3.7/lib_pypy"
	purelib = "/opt/conda/envs/linux-ppc64le-pypy3.7/site-packages"
	scripts = "/opt/conda/envs/linux-ppc64le-pypy3.7/bin"
	stdlib = "/opt/conda/envs/linux-ppc64le-pypy3.7/lib_pypy"

Variables: 
	ABIFLAGS = ""
	AR = "ar"
	ARFLAGS = "rc"
	CC = "cc -pthread"
	CCSHARED = "-fPIC"
	CFLAGS = "-DNDEBUG -O2"
	CXX = "c++ -pthread"
	EXE = ""
	EXT_SUFFIX = ".pypy37-pp73-linux-gnu.so"
	INCLUDEPY = "/opt/conda/envs/linux-ppc64le-pypy3.7/include"
	LDLIBRARY = "libpypy3-c.so"
	LDSHARED = "cc -pthread -shared"
	LDVERSION = "3.7"
	LIBDIR = "/opt/conda/envs/linux-ppc64le-pypy3.7/lib"
	LIBRARY = ""
	MULTIARCH = "ppc_64-linux-gnu"
	OPT = "-DNDEBUG -O2"
	Py_DEBUG = "0"
	Py_ENABLE_SHARED = "0"
	SHLIB_SUFFIX = ".so"
[...]

PR #1260, opened 2022-11-10
Add Fedora's PyPy sysconfig info/dumps for ppc64le/s390x

mbargull:
This is a follow-up to gh-1258 to amend the erroneous `EXT_SUFFIX` from `conda-forge/linux-ppc64le:pypy3.7` builds.
This also adds sysconfigs for s390x since Fedora also builds PyPy for that platform.
xref: https://github.com/messense/python-sysconfig-collector/pull/22


mbargull:
Sanity check: diff from https://github.com/messense/python-sysconfig-collector/pull/22#issuecomment-1309973767 is identical to the one for `src/python_interpreter/sysconfig-linux.json` here.

messense (approved):

PR #1261, opened 2022-11-10
Lookup inherited build- and dev-dependencies in dependencies

knutwalker:
Following the docs in https://doc.rust-lang.org/stable/cargo/reference/workspaces.html#the-workspacedependencies-table
and https://doc.rust-lang.org/stable/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace
which says

    Dependencies in the [dependencies], [dev-dependencies], [build-dependencies], and [target."...".dependencies] sections support the ability to reference the [workspace.dependencies] definition of dependencies.

So, we should not lookup build- and dev-dependencies in their workspace.* equivalent but only in workspace.dependencies

### DIFF
--- a/src/source_distribution.rs
+++ b/src/source_distribution.rs
@@ -108,11 +108,13 @@ fn rewrite_cargo_toml(
                             let mut workspace_dep = workspace_dep.clone();
                             // Merge optional and features from the current Cargo.toml
                             if table[&dep_name].get("optional").is_some() {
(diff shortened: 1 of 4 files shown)


messense:
Thanks! bors r+

messense:
bors ping

messense:
Looks like the macOS runner is stuck, merging manually.

PR #1262, opened 2022-11-10
Convert workspace.dependency entries to tables if necessary

knutwalker:
When a `worksapce.dependency` is written in its short form,
e.g. `rand = "0.8"`, we need to convert that to a table before
combining it with `optional` or `features` entries from the
crate dependencies.


messense:
Thanks! bors r+

PR #1265, opened 2022-11-11
Collect python dependencies from dependent crate metadata

konstin:
This allows rust-numpy to say

```toml
[package.metadata.pyo3-packaging]
dependencies = [
    "numpy>=1.19,<2.0"
]
```

I don't know how useful this is, but it was fun to build


src/build_options.rs
```diff
@@ -725,6 +691,166 @@ impl BuildOptions {
...
+
+/// Returns the merged list of dependencies
+///
+/// Ideally, we'd like to parse the version constraints from the user and merge them with those
+/// from the transitive requirements, but that's not implemented yet
+fn collect_transitive_requires(
+    main_dependencies: &[String],
+    metadata: &Metadata,
+) -> Result<Vec<String>> {
+    let mut merged_requires = main_dependencies.to_vec();
```
adamreichold, line 751: Since you have ownership of `metadata21`, why not pass `main_dependencies` by value, e.g. `take(&mut metadata32.requires_dist)`?

src/build_options.rs
```diff
@@ -725,6 +691,166 @@ impl BuildOptions {
...
+    // https://peps.python.org/pep-0508/#names
+    // https://peps.python.org/pep-0440/#version-specifiers
+    // This isn't the full version spec regex (not sure if that even exists anywhere),
+    // but it should be good enough
+    let distribution_name_re = Regex::new(
+        r"^([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9._-]*[a-zA-Z0-9])\s*(~=|==|!=|<=|>=|<|>|===|;|$).*",
+    )
+    .unwrap();
+    // Since we don't have PEP440 parsing, we do this very simple scheme and to warn the user when we have multiple requires
+    let get_distribution_name = |dep: String| {
```
adamreichold, line 762: Why not just `dep: &str`?

src/build_options.rs
```diff
@@ -725,6 +691,166 @@ impl BuildOptions {
...
+        if sources.len() > 1 {
+            let joined = sources
+                .iter()
+                .map(|(requires, cargo_package)| {
+                    format!("'{}' in crate {}", requires, cargo_package)
+                })
+                .collect::<Vec<String>>()
+                .join(", ");
+            println!(
+                "⚠️ Warning: {dep_name} is specified multiple times: {joined}. \
```
adamreichold, line 837: To be honest, I think this is serious downside. If the issue was very common, i.e. users being swamped with managing the transitive Python dependencies of the Rust dependencies, it might make sense to handle at least the common case.

But since I assume that this is rare (Do we have any example besides NumPy which is somewhat special by exposing a capsule-based C-API for us to consume?), I am not to convince that carrying this without full PEP440 support and limited dependency resolution is really worth it.

test-crates/numpy-user/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+version = "0.1.0"
+edition = "2021"
+
+[lib]
+name = "rust_ext"
+crate-type = ["cdylib"]
+
+[dependencies]
+pyo3 = { version = "0.17", features = ["extension-module"] }
+numpy = { version = "0.17", git = "https://github.com/PyO3/rust-numpy", branch = "python-requires-in-cargo-metadata" }
```
adamreichold, line 12: I was wondering why NumPy 1.19 while our README says 1.16? (The choice predates my involvement, so I cannot say one is correct and the other is not. But I would like to understand if our README would need to be changed.)

I am also somewhat sceptical w.r.t. the upper bound as the situation with Python dependencies is significantly different from Rust dependencies due to the need for venv-wide unification, c.f. https://iscinumpy.dev/post/bound-version-constraints/

adamreichold (commented):
I have to admit to being somewhat sceptical on the value proposition here. rust-numpy's dependency on NumPy is a bit of a special case and while the code would handle that, this would open up the door for feature requests to fully flesh this out with proper PEP440 handling and unification which I suspect would take quite a bit more code to implement.

Additionally, I am not sure if Maturin really wants to be a dependency solver in the first place instead of just packaging up what the user specified unmodified. (Especially said user has no control of the contents of the Cargo.toml of the dependency and while a requirement can be overwritten as it stands, I don't it can be suppressed outright.)

src/build_options.rs
```diff
@@ -725,6 +691,166 @@ impl BuildOptions {
...
+    for dep in main_dependencies {
+        let dist_name = get_distribution_name(dep.clone())
+            .context("Your dependencies table in pyproject.toml is invalid")?;
+        main_deps_names.insert(dist_name, dep.to_string());
+    }
+    debug!("main dependencies {:?}", main_deps_names);
+
+    // distribution name -> (requires string, crate name)
+    let mut metadata_deps_names: HashMap<String, Vec<(String, String)>> = HashMap::new();
+    // I've manually tested that this doesn't include optional dependencies
```
adamreichold, line 784: But does it include enabled optional dependencies, e.g. if there is an optional dependency on `rust-numpy` but it is in use? (That is rather common in my experience, i.e. writing a Rust library that has Python bindings involving NumPy types behind a feature flag to avoid pulling in PyO3 and rust-numpy unconditionally for the Rust users.)

src/build_options.rs
adamreichold: > But writing a proper PEP440 parser is on my list anyway, so maybe i'll do that

Maybe this ties in with https://github.com/PyO3/pyproject-toml-rs or https://github.com/PyO3/python-pkginfo-rs? Or could be organized as a similar project/crate?

konstin: yep absolutely!

konstin:
> I have to admit to being somewhat sceptical on the value proposition here. rust-numpy's dependency on NumPy is a bit of a special case and while the code would handle that, this would open up the door for feature requests to fully flesh this out with proper PEP440 handling and unification which I suspect would take quite a bit more code to implement.

yeah same :D it's more of a pet feature for me, seeing how we could make cross ecosystem dependencies work

> Additionally, I am not sure if Maturin really wants to be a dependency solver in the first place instead of just packaging up what the user specified unmodified. (Especially as said user has no control of the contents of the Cargo.toml of the dependency and while a requirement can be overwritten as it stands, I don't think it can be suppressed outright.)

We could potential just package up all constraints and let the downstream tooling figure it out, but then we'd risk producing cursed package specifications. Just adding everything up unfortunately also doesn't work because we have environment markes

konstin:
it's not worth the extra complexity this adds

Issue #1268, opened 2022-11-14
Raise minimum supported macOS version to 10.12


messense:
rustc planned to do so: https://github.com/rust-lang/rust/pull/104385

- [x] https://github.com/PyO3/maturin/pull/1269
- [ ] Update fallback version accordingly once the upstream change lands

messense:
https://blog.rust-lang.org/2023/09/25/Increasing-Apple-Version-Requirements.html

Issue #1252, opened 2022-11-07
Generate macOS platform tag based on the dylib build version


messense:
Currently it's hard coded if `MACOSX_DEPLOYMENT_TARGET` env var isn't set

https://github.com/PyO3/maturin/blob/cb0b6c6d39777952e7e22e3e88dff61b5b7c3edb/src/target.rs#L685-L686

messense:
I think in reality it's not a good idea to do this because it will end up generating wheels that can only be installed on newer versions of macOS.

Always setting `MACOSX_DEPLOYMENT_TARGET` is still a good idea, but I think we should add a option in `[tool.maturin]` like `macos-deployment-target` to allow projects to declare a default target version, instead of relying on querying the one that rustc support because there might be some C/C++ dependencies that requires higher macOS SDK versions than rustc.

PR #1269, opened 2022-11-14
Query macOS deployment target version from rustc

messense:
Part of #1268

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -1,5 +1,5 @@
 use crate::build_context::BridgeModel;
-use crate::target::{Arch, RUST_1_64_0};
+use crate::target::RUST_1_64_0;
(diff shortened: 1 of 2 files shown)


davidhewitt:
@messense this looks interesting, do you think we should add the same to setuptools-rust?

messense:
setuptools already have it's only way to deal with macOS deployment target version from `wheel`: https://github.com/pypa/wheel/blob/814c2efe8e40051039c5a6de6945e04ecdd162ee/src/wheel/macosx_libfile.py#L392

I also want to implement some similar in maturin (#1252) once https://github.com/m4b/goblin/pull/341 is released.

davidhewitt:
Ah ok 👍

PR #1278, opened 2022-11-19
Use `uniffi.toml` config file if it exist

messense:


messense:
bors retry

messense:
bors retry

PR #1279, opened 2022-11-20
Downgrade cargo_metadata to 0.15.0

messense:
See https://github.com/oli-obk/cargo_metadata/issues/212


messense:
Get Rust 1.48.0 working on CI is too much trouble, giving up for now.

bors r+

PR #1292, opened 2022-11-23
Fix auditwheel with read-only libraries

### LINKED ISSUE #1287: Permission denied error in add_external_libs when libary is not writable by owner
### Bug Description

The issue is that `patchelf` fails when invoked on the copy of the to-be-included library as the original permissions are retained and nothing guaranties that they are owner writable which is the case for all libraries on NixOS

Example:

```console
$ patchelf --set-soname libhs-19c0d7fc.so.5 /tmp/.tmpcBPM8J/libhs-19c0d7fc.so.5
patchelf: open: Permission denied
$ stat -c "Access: %a, UID: %u, GID: %g" /tmp/.tmpcBPM8J/libhs-19c0d7fc.so.5
Access: 555, UID: 1000, GID: 100
$ stat -c "Access: %a, UID: %u, GID: %g" /nix/store/l793dc6r9xagw7cfxv2w3bgwy9cy67yi-hyperscan-5.4.0/lib/libhs.so.5.4.0
Access: 555, UID: 0, GID: 0
```

I understand, there is an ongoing work to get rid of `patchelf`. In the meantime it could be mitigated by writing to a separate file via the `--output` flag of patchelf or fixing the permissions beforehand.

Python 3.8.15

pip 22.2.2

_No response_

### Does `cargo build` work?

- [X] Yes, it works

- [ ] Yes

### Steps to Reproduce

I guess it is enough to set permissions to e..g `555` of a library dependency which gets picked up by auditwheel

vlaci:
`std::fs::copy` also copies the original file permissions. If the original library at `lib_path` was not writable, e.g. because it had its permission set to 0o555, the copy inherited the same permission causing later file manipulations, like `patchelf --set-soname` to fail.

Fixes #1287

### DIFF
--- a/src/build_context.rs
+++ b/src/build_context.rs
@@ -392,6 +392,12 @@ impl BuildContext {
             fs::copy(&lib_path, &dest_path)?;
             libs_copied.insert(lib_path);
 
+            // fs::copy copies permissions as well, and the original
+            // file may have been read-only
+            let mut perms = fs::metadata(&dest_path)?.permissions();
+            perms.set_readonly(false);
+            fs::set_permissions(&dest_path, perms)?;
+
             patchelf::set_soname(&dest_path, &new_soname)?;
             if !lib.rpath.is_empty() || !lib.runpath.is_empty() {
                 patchelf::set_rpath(&dest_path, &libs_dir)?;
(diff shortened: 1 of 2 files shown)


src/build_context.rs
```diff
@@ -392,6 +392,12 @@ impl BuildContext {
             fs::copy(&lib_path, &dest_path)?;
             libs_copied.insert(lib_path);
 
+            // fs::copy copies permissions as well, and the original
+            // file may have been read-only
+            let mut perms = fs::metadata(&dest_path)?.permissions();
+            perms.set_readonly(false);
+            fs::set_permissions(&dest_path, perms)?;
```
vlaci, line 399: I thought about only setting the permission when needed, but didn't think that it worth the extra if condition to spare a few syscalls, as a stat is needed for each file anyway.

messense (approved):
Looks good to me, thanks!

bors r+

messense:
Thanks for adding the changelog entry.

bors retry

vlaci:
Thanks for the fast response.

Issue #1289, opened 2022-11-23
Generate correct wheel name for armv8l


messense:
See https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/41663#note_275648

armv8l is 32-bit mode on aarch64.

```bash
root@raspberrypi:~# uname -a
Linux raspberrypi 5.15.61-v8+ #1579 SMP PREEMPT Fri Aug 26 11:16:44 BST 2022 aarch64 GNU/Linux
root@raspberrypi:~# linux32 uname -a
Linux raspberrypi 5.15.61-v8+ #1579 SMP PREEMPT Fri Aug 26 11:16:44 BST 2022 armv8l GNU/Linux
```

nekopsykose:
for the specific issue there- armv8l wouldn't be the correct name either. it means 32-bit on 64, but the actual wanted output is `armv7` (actual 32-bit) - which is the rust toolchain triple (`rustc -vV` host: line). the uname value should not be the source of truth here, as it's not related to the actual toolchain target.

i'm not sure how it would affect every case, but defaulting to the rustc host instead of uname sounds like it would just always be correct without this edgecase, and then --target can be used for people that cross compile or want to set it manually like now with no change there.

messense:
Maybe we can just add an escape hatch like `_PYTHON_HOST_PLATFORM`?

messense:
We used to generate armv7l but the Alpine Linux CI will fail to install it because pip wants armv8l, see https://github.com/PyO3/maturin/issues/490#issuecomment-840634183

nekopsykose:
> pip wants armv8l

you mean pip installs things based on the current sysconfig thing that reads uname? inconvenient.. not sure how to really fix it then without adding shims like that everywhere

messense:
> you mean pip installs things based on the current sysconfig thing that reads uname?

Exactly

```bash
$ root@raspberrypi:~# linux32 pip debug --verbose
WARNING: This command is only meant for debugging. Do not use this with automation for parsing and getting these details, since the output and options of this command may change without notice.
pip version: pip 22.0.4 from /usr/local/lib/python3.9/dist-packages/pip (python 3.9)
sys.version: 3.9.2 (default, Feb 28 2021, 17:03:44)
[GCC 10.2.1 20210110]
sys.executable: /usr/bin/python3
sys.getdefaultencoding: utf-8
sys.getfilesystemencoding: utf-8
locale.getpreferredencoding: UTF-8
[...]

PR #1301, opened 2022-11-27
Fix wrong platform tag when building in i386 docker container on x86_64 host

messense:
Fixes the i386 issue reported in https://github.com/PyO3/maturin/issues/1289#issuecomment-1328225431.

We can not trust the return value of `uname -m` in Docker container.


messense:
bors retry

messense:
bors seems stuck.

PR #1302, opened 2022-11-27
CI: Stop caching test crates target directory

messense:
CI spends quite a lot of time uploading cache, let's see how it goes without test crates cache.


messense:
bors try

messense:
bors try

messense:
bors retry

messense:
The total duration is reduced from [58m 16s](https://github.com/PyO3/maturin/actions/runs/3551820475/usage) to [48m 45s](https://github.com/PyO3/maturin/actions/runs/3561305674/usage).

PR #1313, opened 2022-11-30
Add riscv64 target support for FreeBSD

messense:
Tested in QEMU riscv64 FreeBSD.


src/target.rs
```diff
@@ -256,8 +259,8 @@ impl Target {
                     Arch::X86 => "i386",
                     Arch::Aarch64 => "arm64",
                     Arch::Armv6L | Arch::Armv7L => "arm",
-                    Arch::Powerpc64 => "powerpc64",
-                    Arch::Powerpc64Le => "powerpc64le",
+                    Arch::Powerpc | Arch::Powerpc64 | Arch::Powerpc64Le => "powerpc",
```
messense, line 262: Per

https://github.com/PyO3/maturin/blob/4d6147553ca835403a38c2d55555a54dacb01311/sysconfig/cpython-freebsd-13.0-powerpc64.txt#L1

and 

https://github.com/PyO3/maturin/blob/4d6147553ca835403a38c2d55555a54dacb01311/sysconfig/cpython-freebsd-13.0-powerpc64le.txt

They are all just `powerpc`.

PR #1334, opened 2022-12-05
Support SOURCE_DATE_EPOCH environment variable in wheel building

### LINKED ISSUE #1320: Use SOURCE_DATE_EPOCH environment variable to set mtime of files in wheel
Currently running `maturin build` does not have reproducible wheel output even if the input files have not changed. This can be observed by doing the following:

1. Running `maturin build`
2. Record the md5sum of the output wheel
3. Running the same `maturin build` command again without changing any of the python or rust files
4. Observe the md5sum of the output wheel is different

I noticed in my case it's because the `mtime` timestamp of the files in the wheel have been updated.

It would be great if `maturin` respected the `SOURCE_DATE_EPOCH` environment variable like wheel already does:
https://github.com/pypa/wheel/blob/70d8f7f5189fac0dfbf21b83adbd2781d2537c17/src/wheel/wheelfile.py#L26-L31

If this environment variable is set, it would be used to set the mtime of the files in the wheel, making the output reproducible. 

Having reproducibility is really handy for me because a build system may skip tasks if the wheel is the same.

I think the change would be as simple as adjusting https://github.com/PyO3/maturin/blob/6e7b02ffb043bced722d5b163f16ae10accdac17/src/module_writer.rs#L237-L241 to explicitly set the mtime.

zmanji:
This adds support for the `SOURCE_DATE_EPOCH` environment variable like wheel does when building wheels. If this environment variable is set, the `mtime` of the files in the wheel are set to this value to ensure reproducible output.

Fixes #1320

### DIFF
--- a/src/module_writer.rs
+++ b/src/module_writer.rs
@@ -14,6 +14,7 @@ use ignore::WalkBuilder;
 use normpath::PathExt as _;
 use sha2::{Digest, Sha256};
 use std::collections::{HashMap, HashSet};
+use std::env;
 use std::ffi::OsStr;
 use std::fmt::Write as _;
 #[cfg(target_family = "unix")]
@@ -26,8 +27,10 @@ use std::path::{Path, PathBuf};
 use std::process::{Command, Output};
 use std::str;
 use tempfile::{tempdir, TempDir};
+use time::macros::datetime;
+use time::OffsetDateTime;
 use tracing::debug;
-use zip::{self, ZipWriter};
+use zip::{self, DateTime, ZipWriter};
 
 /// Allows writing the module to a wheel or add it directly to the virtualenv
 pub trait ModuleWriter {
@@ -234,9 +237,15 @@ impl ModuleWriter for WheelWriter {
         } else {
             zip::CompressionMethod::Deflated
         };
-        let options = zip::write::FileOptions::default()
+
+        let mut options = zip::write::FileOptions::default()
             .unix_permissions(permissions)
             .compression_method(compression_method);
+        let mtime = self.mtime().ok();
+        if let Some(mtime) = mtime {
+            options = options.last_modified_time(mtime);
+        }
+
         self.zip.start_file(target.clone(), options)?;
         self.zip.write_all(bytes)?;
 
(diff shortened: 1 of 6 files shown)


src/module_writer.rs
```diff
@@ -309,14 +318,36 @@ impl WheelWriter {
...
+        let dt = OffsetDateTime::from_unix_timestamp(epoch)?;
+        let min_dt = datetime!(1980-01-01 0:00 UTC);
+        let dt = dt.max(min_dt);
+
+        let dt = DateTime::from_time(dt);
+
+        match dt {
+            Ok(dt) => return Ok(dt),
+            Err(_) => return Err(anyhow!("failed to build zip DateTime")),
+        }
```
messense, line 334: I think it can be simplified to

```suggestion
        let dt = DateTime::from_time(dt).context("Failed to build zip DateTime")?;
        Ok(dt)
```

messense (commented):
Thanks for the PR! Could you please fix the clippy warnings and add a changelog entry in `Changelog.md`?

src/module_writer.rs
zmanji: This does not work due to the following error:
```
error[E0599]: the method `context` exists for enum `std::result::Result<zip::DateTime, ()>`, but its trait bounds were not satisfied
   --> src/module_writer.rs:329:42
    |
329 |         let dt = DateTime::from_time(dt).context("Failed to build zip DateTime")?;
    |                                          ^^^^^^^ method cannot be called on `std::result::Result<zip::DateTime, ()>` due to unsatisfied trait bounds
    |
   ::: /home/zmanji/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/result.rs:504:1
    |
504 | pub enum Result<T, E> {
    | --------------------- doesn't satisfy `_: anyhow::Context<zip::DateTime, ()>`
    |
    = note: the following trait bounds were not satisfied:
            `(): anyhow::context::ext::StdError`
            which is required by `std::result::Result<zip::DateTime, ()>: anyhow::Context<zip::DateTime, ()>`
```

I am not sure how to resolve it. I have updated this to satisfy clippy.

src/module_writer.rs
```diff
@@ -309,14 +318,36 @@ impl WheelWriter {
...
     pub fn finish(mut self) -> Result<PathBuf, io::Error> {
         let compression_method = if cfg!(feature = "faster-tests") {
             zip::CompressionMethod::Stored
         } else {
             zip::CompressionMethod::Deflated
         };
-        let options = zip::write::FileOptions::default().compression_method(compression_method);
+
+        let mut options = zip::write::FileOptions::default().compression_method(compression_method);
+        let mtime = self.mtime().ok();
```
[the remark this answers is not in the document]
zmanji: Wheel does not print a warning: https://github.com/pypa/wheel/blob/70d8f7f5189fac0dfbf21b83adbd2781d2537c17/src/wheel/wheelfile.py#L26

I think based on that it's not required to do so here. I do not recall other tools that support `SOURCE_DATE_EPOCH` printing out a warning as well.

zmanji:
> Thanks for the PR! Could you please fix the clippy warnings and add a changelog entry in `Changelog.md`?

This should be done now.

messense (approved):
bors r+

messense:
Released in v0.14.5.

Issue #1337, opened 2022-12-06
Reduce duplicate dependencies


messense:
Currently we have lots of duplicate dependencies, see https://github.com/PyO3/maturin/pull/1336/files, we should reduce them for better compile time.

- [x] goblin 0.5: https://github.com/mozilla/uniffi-rs/pull/1405
- [x] clap 0.3: https://github.com/mozilla/uniffi-rs/pull/1417
- [x] terminal_size 0.1: https://github.com/console-rs/console/pull/146
- [x] windows-sys 0.36: https://github.com/steffengy/schannel-rs/pull/82
- [ ] sha2 0.9: https://github.com/hwchen/secret-service-rs/pull/61 or https://github.com/PyO3/maturin/pull/1331

PR #1352, opened 2022-12-13
Run `uniffi-bindgen` command instead of depending on `uniffi_bindgen` crates

messense:
`uniffi_bindgen` crate is too heavy weight and people uses `uniffi` usually have already installed uniffi-bindgen.

Part of #1337, reduces `--no-default-features` compile time on M2 MacBook Air from 26.2s to 22.3s.


messense:
bors retry

Issue #1372, opened 2022-12-24
Complex Python + Rust layout


AleCandido:
At the moment, I have a Python distribution package managed by Poetry, and I'd like to implement one of its modules in Rust.

On one side, I already had successful and pleasant experience using `maturin`, but I also like Poetry environment management, development dependencies, and lock files.

I already had a look into issues and discussions, and found out that the two things are not incompatible: what I like of Poetry is not the build backend (i.e. `poetry-core`), but everything else.
Following #1246, I'm willing to switch build backend from `poetry-core` to `maturin`, while retaining Poetry for development.

Unfortunately, my package is already more complex than the proposed layouts, and I'd like to also introduce an extended Rust layout.
In the following, I will outline the two levels of support I'm looking for: the first consist in better integration with pretty standard Python layout (and possibly Poetry, but I believe this not to be `maturin`'s responsibility), the second is an extended Rust layout, yet not departing from standard practices.
In order to clarify, I tried to collect the two layouts in a dedicated repo:

https://github.com/AleCandido/atuin

## Multi-package Python

At the moment, my current repository contains a single Python [distribution package](https://packaging.python.org/en/latest/glossary/#term-Distribution-Package), but multiple [import packages](https://packaging.python.org/en/latest/glossary/#term-Import-Package).

I tried to follow the advice in #1246, retaining `[tool.poetry]`, but implementing the [src layout](https://www.maturin.rs/project_layout.html#alternate-python-source-directory-src-layout).
In the repo, this is done in the `alternate` folder.

With this layout, most of the expected workflow works smoothly, i.e.:

- I can install Python import packages with Poetry (declaring them in the `[tools.poetry]` section)
- I run `poetry run maturin develop` to recompile the Rust part and install in development
- I can package a working wheel by running `poetry run maturin build`

The few places in which this is falling short are:

- `maturin` is currently not packaging the extra Python import packages
- Poetry front-end is not properly working (I'd expect `poetry build` to work similar to `pip wheel ...`, but it isn't, since it is not invoking the build backend)
  - this I will raise on Poetry repo (after someone else's confirmation, if possible)

It seems like nor PEP 517 neither PEP 621 define any way to specify multiple import packages, so I'd suggest to detect multiple folders in the `src` directory, and check they contain a top-level `__init__.py` file.

## Rust workspace

One of the reasons to switch to Rust was to release this part with API for a different language (i.e. make a Rust crate), and possibly provide C bindings as well.

[...]

PR #1378, opened 2022-12-28
Add support for packaging multiple pure Python packages

messense:
Implements https://github.com/PyO3/maturin/issues/1372#issuecomment-1366368629


AleCandido:
The PR looks sensible to me, but I'm not the best person to review, since I'm not experienced in `maturin` internals.

What I can do is testing in the demo project, building from source.

messense:
Please try v0.14.8b1: https://pypi.org/project/maturin/0.14.8b1/

PR #1386, opened 2023-01-03
chore: cargo update `ureq` and `console`

messense:


messense:
Wait for https://github.com/algesten/ureq/pull/578

bors r-

Issue #1397, opened 2023-01-08
Provide more guidance on which kind of bindings to use


not-my-profile:
Running `maturin init` or `maturin new` currently gives the following prompt:
```
? 🤷 Which kind of bindings to use? ›
❯ pyo3
  rust-cpython
  cffi
  uniffi
  bin
```

I think it would make sense to link some website that explains the differences (advantages/disadvantages) of these various bindings. Perhaps in a new section in the [user guide](https://www.maturin.rs/)? https://pyo3.rs/v0.17.3/rust_cpython.html would probably be a good page to link from that section but ideally the section would also explain (or link pages that explain) the differences between cffi, uniffi, etc.

Lastly as I understand it the `bin` option actually isn't a binding at all, so this should probably be better explained in the CLI prompt. And it also does not appear to be documented in the user guide.

messense:
https://www.maturin.rs/bindings.html

not-my-profile:
Ah thanks, I did indeed miss that page. I mistakenly interpreted the sidebar of the user guide to be a table of contents for the current page (just because I saw "Contributing" to be the last section of both).

Ok great that page is indeed helpful but I think it be more explanatory for people who are unfamiliar with the various binding technologies. And I still think that `bin` technically isn't a binding. I can look into improving that page but I'll have to familiarize myself firstly more with these various technologies since I don't have much experience with them.

What do you think should `maturin init/new` display a link to https://www.maturin.rs/bindings.html?

messense:
> What do you think should `maturin init/new` display a link to https://www.maturin.rs/bindings.html?

Sure, added in https://github.com/PyO3/maturin/pull/1403.

messense:
> And I still think that `bin` technically isn't a binding. I

Yeah it's also getting complicated now that maturin also supports [bin bindings that uses `pyo3`](https://github.com/PyO3/maturin/pull/938).

messense:
> Ok great that page is indeed helpful but I think it could be more explanatory for people who are unfamiliar with the various binding technologies.

Closing for now, feel free to send pull requests to improve the code or docs, thanks!

PR #1403, opened 2023-01-09
Add link to bindings guide in `new` and `init` commands

messense:
See #1397 

image


not-my-profile:
Related note: I think it would also make sense to link https://www.maturin.rs/ in the `--help` output.

PR #1405, opened 2023-01-09
Respect `rustflags` settings in cargo configuration file

messense:
https://github.com/taiki-e/cargo-config2

`CARGO_ENCODED_RUSTFLAGS` requires Rust 1.55+ so technically it's a breaking change even though our MSRV is 1.62 we can actually build packages with older versions of Rust.


messense:
bors try

PR #1421, opened 2023-01-19
Completing type hints for python files

ischaojie:
The result:

```
➜  maturin git:(type-hints) mypy maturin
Success: no issues found in 3 source files
```


messense:
Thanks! Maybe we should also add mypy check to CI in `.github/workflows/lint.yaml`?

ischaojie:
> Thanks! Maybe we should also add mypy check to CI in `.github/workflows/lint.yml`?

I will add it later.

messense (approved):
Looks great, thanks again!

bors r+

PR #1435, opened 2023-01-30
Don't package dev-only path dependencies in sdist

messense:
When building from sdist, `cargo` only needs `dependencies` and `build-dependencies`, removing `dev-dependencies` entirely from `Cargo.toml` makes sdist size smaller and avoid some problems with dev-only path deps like https://github.com/PyO3/maturin/issues/1083#issuecomment-1407642659.


konstin (approved):

messense:
bors r=konstin

PR #1462, opened 2023-02-05
Detect .venv in current or parent folder in maturin develop

konstin:
See discussion in #1461

---

`maturin develop` now looks for a virtualenv .venv in the current or any parent directory if no venv or conda environment is active.

---

If someone knows a good `execv` equivalent on windows i'll switch over to that, currently it's a subcommand which is still fine.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * **Breaking Change**: Remove deprecated `python-source` option in `Cargo.toml` in [#1335](https://github.com/PyO3/maturin/pull/1335)
 * **Breaking Change**: Turn `patchelf` version warning into a hard error in [#1335](https://github.com/PyO3/maturin/pull/1335)
 * **Breaking Change**: [`uniffi_bindgen` CLI](https://mozilla.github.io/uniffi-rs/tutorial/Prerequisites.html#the-uniffi-bindgen-cli-tool) is required for building `uniffi` bindings wheels in [#1352](https://github.com/PyO3/maturin/pull/1352)
+* `maturin develop` now just like `maturin run` also look for a virtualenv .venv in the current or any parent directory if no environment is active.
 * Add Cargo compile targets configuration for filtering multiple bin targets in [#1339](https://github.com/PyO3/maturin/pull/1339)
 * Respect `rustflags` settings in cargo configuration file in [#1405](https://github.com/PyO3/maturin/pull/1405)
 * Bump MSRV to 1.63.0 in [#1407](https://github.com/PyO3/maturin/pull/1407)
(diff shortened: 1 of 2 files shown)


konstin:
image

i tried restarting but it only failed again

Cargo.toml
```diff
@@ -83,6 +83,7 @@ multipart = { version = "0.18.0", features = ["client"], default-features = fals
 ureq = { version = "2.6.1", features = ["gzip", "socks-proxy"], default-features = false, optional = true }
 native-tls-crate = { package = "native-tls", version = "0.2.8", optional = true }
 keyring = { version = "1.1.1", optional = true }
+nix = { version = "0.26.2", default-features = false, features = ["process"], optional = true }
```
messense, line 86: How about

```suggestion
[target.'cfg(unix)'.dependencies]
nix = { version = "0.26.2", default-features = false, features = ["process"], optional = true }
```

messense:
Should `maturin run` build the project automatically before executing the given command?

Cargo.toml
konstin: good catch, thanks

konstin:
i'd say no, because building is slow (vs. run being near-instant) and we'd need to handle all the cli args to develop. Maybe we could add a `--run <python options>` flag to develop instead, so could e.g. do `maturin develop -m python_bindings/Cargo.toml --run "-m pytest"`

messense:
> `maturin develop` now just like `maturin run` also look for a virtualenv .venv in the current or any parent directory if no environment is active.

Maybe we should split `maturin develop` changes into another PR given that `maturin run` needs some more discussions.

konstin:
I've changed this to only include the `maturin develop` change and moved maturin run to #1475

messense (approved):

Changelog.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * **Breaking Change**: Remove deprecated `python-source` option in `Cargo.toml` in [#1335](https://github.com/PyO3/maturin/pull/1335)
 * **Breaking Change**: Turn `patchelf` version warning into a hard error in [#1335](https://github.com/PyO3/maturin/pull/1335)
 * **Breaking Change**: [`uniffi_bindgen` CLI](https://mozilla.github.io/uniffi-rs/tutorial/Prerequisites.html#the-uniffi-bindgen-cli-tool) is required for building `uniffi` bindings wheels in [#1352](https://github.com/PyO3/maturin/pull/1352)
+* `maturin develop` now just like `maturin run` also look for a virtualenv .venv in the current or any parent directory if no environment is active.
```
messense, line 15:
```suggestion
* `maturin develop` now looks for a virtualenv `.venv` in the current or any parent directory if no virtual environment is active.
```

PR #1469, opened 2023-02-09
Fix link to ballista

kngwyu:
It seems to have its own home now. Also, I changed the datafusion link title to 'datafusion-python' for consistency.


messense (approved):
Thanks！

PR #1475, opened 2023-02-11
maturin run, a new subcommand that launches python with the correct venv

konstin:
Based on #1462.

I don't think this needs to build project before running, it's effectively just `python <args>` except it picks the right `python` by default.


konstin:
Obsolete through `uv run`.

PR #1476, opened 2023-02-12
Use extension_name instead of module_name for CFFI extensions in develop mode

luizirber:
I noticed when running `maturin develop` with [sourmash](https://github.com/sourmash-bio/sourmash) that the CFFI extension ended in `src/sourmash/sourmash` instead of the expected `src/sourmash/_lowlevel.py` (to match the [`[package.maturin.metadata.name]`](https://github.com/sourmash-bio/sourmash/blob/fa3ead658e189cfb196c1ee0c5ffd2047c23af70/src/core/Cargo.toml#L95L96) field).
`maturin build` puts the extension in the right place, so checking how it worked I noticed that changing the line from this PR would make the behavior match in `develop` and `build`.

(and now I'm opening a PR to see if I missed some test and I'm breaking other use cases 😬 )


messense (approved):
LGTM, could you please add a changelog entry in `CHANGELOG.md`? Thanks!

messense:
I'll backport this to v0.14 soon and release a new version soon.

luizirber:
I was looking at the project history and just noticed this PR is the CFFI equivalent of https://github.com/PyO3/maturin/pull/1416. Mostly commenting to link both PRs =]

Issue #325, opened 2020-07-08
When cargo building pyo3 projects auto-add in the extension-module feature please


gilescope:
At the moment maturin warns if a pyo3 project is built without the extension-module feature.

We can already do:

maturin develop --cargo-extra-args="--features pyo3/extension-module"

Why not add that flag in automatically if it's clear that it's not there? Having pyo3/extension-module turned on as a default feature seems like a footgun as people will just run cargo test and expect stuff to work on cli, IDEs etc. but will run into linking problems if they don't use --no-default-features. It's both unergonomic and confusing as the linker errors give no clue that you should have disabled a feature so it's tough on new users.

It would be great if maturin just by default did the right thing :-) - 
(it could still squark that it wasn't found so adding it in explicitly.)

thoughs?

davidhewitt:
I think that's a very good idea!

konstin:
Blocked on https://github.com/PyO3/pyo3/pull/1040 being resolved.

messense:
After #605 lands, you can add the `cargo-extra-args = "--features pyo3/extension-module"` to the `[tool.maturin]` section in `pyproject.toml` and maturin will pick it up automatically.

konstin:
I believe `[tool.maturin] cargo-extra-args` in is the way to go

jmhodges:
Reading between the lines here, I think this issue closing means we could take an update to [getting_started.md](https://github.com/PyO3/pyo3/blob/4d80475008801c428dc0dd74c4aefaa2be447ee3/guide/src/getting_started.md#pyprojecttoml)'s examples and maybe to the [FAQ](https://github.com/PyO3/pyo3/blob/4d80475008801c428dc0dd74c4aefaa2be447ee3/guide/src/faq.md#i-cant-run-cargo-test-or-i-cant-build-in-a-cargo-workspace-im-having-linker-issues-like-symbol-not-found-or-undefined-reference-to-_pyexc_systemerror) to allow folks to run `cargo test` work out of the box

I'm willing to do that, would like to confirm what's desired.

I think what we want is the example `pyproject.toml` to gain a

```toml
[tool.maturin]
features = ["pyo3/extension-module"]
```

while the example `Cargo.toml`'s `dependencies` looks like:

```toml
[dependencies]
pyo3 = "0.18.0"
```

(Or I could be convinced to make that `pyo3 = { version = "0.18", features = ["anyhow"] }` since that's so handy)

jmhodges:
[...]

PR #1479, opened 2023-02-13
Enable `extension-module` feature in `pyproject.toml` in project templates

messense:
Implements https://github.com/PyO3/maturin/issues/325#issuecomment-1416697270.

Thanks for the idea @jmhodges!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1457,9 +1457,9 @@ dependencies = [
 
 [[package]]
 name = "minijinja"
-version = "0.30.3"
+version = "0.30.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "277b87a5d253b7cfbfe410802abf884bd1d96ccba8afa7af1b02265c43cb0f14"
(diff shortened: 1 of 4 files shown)


messense:
This has the benefits of that `cargo test` will work out of the box for non-Python related tests written in Rust. I guess we can also add `maturin test` command that invokes `cargo test` with `auto-initialize` feature enabled to support testing Python related tests written in Rust?

What do you guys think? @konstin @davidhewitt

konstin (approved):
good idea!

messense:
bors r=konstin

messense:
bors retry

davidhewitt:
> This has the benefits of that `cargo test` will work out of the box for non-Python related tests written in Rust. I guess we can also add `maturin test` command that invokes `cargo test` with `auto-initialize` feature enabled to support testing Python related tests written in Rust?

I think it's less necessary, because it's easy to add a `pyo3` dev-dependency with the `auto-initialize` feature. Plus, not all projects use `auto-initialize`, some use custom interpreter setup at the beginning of each test.

PR #1492, opened 2023-02-18
Backport and release v0.14.14

messense:


messense:
Dropped `package.metadata.maturin.name` deprecation for now to unblock release, we can do a new patch release after https://github.com/PyO3/maturin/pull/1493.

Issue #1489, opened 2023-02-16
Tracking issue for 1.0.0 release


messense:
## Breaking Changes

* [x] move `RemainingCoreMetadata` over to `pyproject.toml`

## Bug fixes

* [ ] https://github.com/PyO3/maturin/issues/1490
* [x] ~~https://github.com/hwchen/keyring-rs/pull/118~~

## Documentation

* [x] Add migration guide
* [x] Add documentation for `maturin generate-ci`
* [x] Add documentation for maturin related environment variables like the ones added in https://github.com/PyO3/maturin/pull/1483

## Nice to have
* [x] https://github.com/PyO3/maturin/issues/1252

konstin:
I think we should also move `RemainingCoreMetadata` over to pyproject.toml before 1.0. Currently there's a weird split where you need to configure maturin in both Cargo.toml and pyproject.toml; I only realized this when i hit https://www.maturin.rs/project_layout.html#import-rust-as-a-submodule-of-your-project with pep440_rs. Afaik `targets = ` and `data = ` of `[package.metadata.maturin]` are also currently not part of the docs.

landhb:
This should no longer be blocked on `linux-keyutils`. The build has been fixed in the `0.2.3` release.

PR #1493, opened 2023-02-19
Move maturin metadata from `Cargo.toml` to `pyproject.toml`

messense:
Implements https://github.com/PyO3/maturin/issues/1489#issuecomment-1432975142

### DIFF
--- a/src/pyproject_toml.rs
+++ b/src/pyproject_toml.rs
@@ -81,6 +81,19 @@ impl GlobPattern {
     }
 }
 
+/// Cargo compile target
+#[derive(Serialize, Deserialize, Debug, Clone, Default)]
+#[serde(rename_all = "kebab-case")]
+pub struct CargoTarget {
+    /// Name as given in the `Cargo.toml` or generated from the file name
+    pub name: String,
+    /// Kind of target ("bin", "lib")
+    pub kind: Option<String>,
+    // TODO: Add bindings option
+    // Bridge model, which kind of bindings to use
+    // pub bindings: Option<String>,
+}
+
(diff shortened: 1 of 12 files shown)


src/pyproject_toml.rs
```diff
@@ -86,6 +86,8 @@ impl GlobPattern {
 #[serde(rename_all = "kebab-case")]
 pub struct ToolMaturin {
     // maturin specific options
+    // extension module name, accepts setuptools style import name like `foo.bar`
+    name: Option<String>,
```
messense, line 90: I feel that `name` isn't a good name for this, how about `extension-name` / `extension-module`?

@konstin

konstin: yeah, maybe `rust-module-name` or `native-module-name`? i think `extension` a bit misleading because in maturin's case it's not an extension, it's the main thing (and the metaphor of extending cpython by adding new built-in modules is lost on most users)

konstin:
What does this pull request need to be ready to be merged?

src/pyproject_toml.rs
messense: I think `module-name` is a good alternative.

konstin (approved):
Neat! 

This is the last breaking change we need for 1.0, right?

messense:
Yeah, I'd also like to release a new 0.14 version to deprecate `package.metadata.maturin` first.

konstin:
sounds good! i also thought about publishing a v0.15 first so people can test all the breaking changes and then publish a 1.0 two weeks or so later

messense:
bors r=konstin

PR #1500, opened 2023-02-25
Use `keyutils` instead of `secret-service` for accessing Linux keyring

messense:
Removes a lot of async related dependencies.


messense:
Release workflow passed: https://github.com/PyO3/maturin/actions/runs/4270534216

PR #1503, opened 2023-02-27
Build maturin with native-tls support on aarch64 Windows

messense:


messense:
It works: https://github.com/PyO3/maturin/actions/runs/4278069348/jobs/7447447482

Issue #368, opened 2020-10-25
Ship both python library AND binary command line tool in wheel


PeterQLee:
I'm in a situation where I would like to build both a python library and a command line tool in a wheel for distribution. (e.g., https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html , but all the code would be in rust.)

The Cargo.toml would contain:
```toml
...
[lib]
crate-type = ["staticlib", "cdylib"]
name = "my_library"
path = "src/lib.rs"

[[bin]]
name = "my_commandline"
path= "src/cli.rs"
...
```

Is it possible to do this in maturin?

I've tried doing `maturin build -b "bin"`, which only builds the command line tool (my_commandline) and not the library.
`maturin build -b "pyo3"` likewise only builds the library (my_library) and not the command line tool.

Is it currently possible to build with both "bin" and "pyo3" bindings in a single wheel?

konstin:
For shipping both library and binary, you'd need to build twice and ship both binaries, having two copies of each dependency in the wheel. Would it be possible to work around issue by adding a pseudo-main to the rust library and creating a small python script that  only calls the main of library?

rmcgibbo:
Does anyhow know how to do this from the pyproject.toml, so that `pip wheel` or `pip install` builds either the po3 or the bin type of wheel, since we can't get them both into the same wheel?

messense:
After #948 we now have the skeleton to allow shipping both lib and bin, the remaining question is how to let users to configure it. We'd still want maturin to by default only build `lib` or `bin`, build for both of them when asked to.

I'm thinking about adding the following options to `pyproject.toml` (and `Cargo.toml`):

```toml
[[tool.maturin.targets]]
name = "example"
kind = "bin"

[[tool.maturin.targets]]
name = "example"
kind = "lib"
```

Then you will get a `libexample` and `example` binary in the wheel.

I don't quite like the `targets` metadata name, easily gets confused with cargo's `--target`.

davidhewitt:
Sounds good to me. Agreed `targets` is a little confusing, though it really is the name used by Cargo for these things: https://doc.rust-lang.org/cargo/reference/cargo-targets.html#cargo-targets

dylanbstorey:
I like it and while `targets` might be a little confusing at first blush, it seems like a community standard and not an inordinate cognitive barrier to understand and easy to document.

messense:
Thanks for the feedback!

[...]

PR #1565, opened 2023-04-11
Document binary and library in a single package by entrypoint workaround

konstin:
Inspired by https://github.com/PyO3/maturin/issues/368#issuecomment-1503752982

Preview: https://deploy-preview-1565--maturin-guide.netlify.app/bindings.html#both-binary-and-library

CC @nanthony007

### DIFF
--- a/guide/src/bindings.md
+++ b/guide/src/bindings.md
@@ -91,6 +91,42 @@ directory of a virtual environment) once installed.
 > **Note**: Maturin _does not_ automatically detect `bin` bindings. You _must_
 > specify them via either command line with `-b bin` or in `pyproject.toml`.
 
+### Both binary and library?
+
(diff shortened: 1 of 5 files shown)


nanthony007:
This is awesome!

messense:
File path on Windows is ridiculously [complicated](https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats), maybe we should simplify the test case to not assert file paths.

konstin:
The tests are passing now, i've removed checks for platforms with inconsistent behaviors. I think the current `check_installed.py` is also good as documentation.

messense (approved):
bors r+

konstin:
bors retry

looks like a network error: https://github.com/PyO3/maturin/actions/runs/4760507648/jobs/8460910927

konstin:
bors r+

discovering more strange platform behavior as i go

konstin:
bors retry

PR #1573, opened 2023-04-20
Fix module_writers::write_python_part include target check

viccie30:
It is only called from functions that create wheels, not sdists.


messense (approved):
Thanks!

bors r+

PR #1578, opened 2023-04-26
Add trusted publisher support

### LINKED ISSUE #1575: Trusted Publisher support for PyPI
Recently, PyPI has introduced a new authorization method called "Trusted publishing", which is based on [OpenID Connect (OIDC)](https://openid.net/connect/).

https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/

I wonder if it makes sense to support it in Maturin directly, or it is better to delegate it to [`maturin-action`](https://github.com/PyO3/maturin-action). I don't know if PyPI is ever going to support other OIDC identity providers besides GitHub.

Technical details: https://docs.pypi.org/trusted-publishers/using-a-publisher/

messense:
Closes #1575


messense:
Manually tested in `typst-py` project and it works:

image

src/upload.rs
```diff
@@ -196,12 +199,23 @@ fn resolve_pypi_cred(
...
     registry_name: Option<&str>,
+    registry_url: &str,
 ) -> (String, String) {
     // API token from environment variable takes priority
     if let Ok(token) = env::var("MATURIN_PYPI_TOKEN") {
         return ("__token__".to_string(), token);
     }
 
+    // Try to get a token via OIDC exchange
+    match resolve_pypi_token_via_oidc(registry_url) {
```
konstin, line 210: what happens if someone uses this with a registry that isn't pypi?

messense: It will print a failure warning, but we can certainly do better to not show a warning. I'll improve it later.

messense: https://github.com/PyO3/maturin/commit/6daaed5aac9f08820bd7cc2b5f3033ad470b5c30

konstin: thank you!

Issue #1576, opened 2023-04-22
`maturin develop` doesn't include text color from rust build errors


Thenerdstation:
When using `maturin develop`, any rust error's that appear are just displayed in system default color text, making it slightly harder to read.

<img width="782" alt="image" src="https://user-images.githubusercontent.com/8702042/233760846-494fda6a-892d-4c3b-bfa1-05b3adc80978.png">

Vs when using say `cargo build`

<img width="690" alt="image" src="https://user-images.githubusercontent.com/8702042/233760803-be94cadd-f228-485f-bfde-d01bb757980c.png">

  - How is this important to you? How would you use it?
  Maybe a 3/10 importance. Certainly not a blocking issue, but it is a strong QOL issue.

  - Can you think of any alternatives?
  Leaving it as is.
  
  - Do you have any ideas about how it can be implemented? Are you willing/able to implement it? Do you need mentoring?
  Not sure why the coloring is missing, but I could figure it out I would be happy to open a PR.

PR #1591, opened 2023-05-05
Enabled color output for error message

ischaojie:
Closes https://github.com/PyO3/maturin/issues/1576

It looks like you just need to change ` message-format` to `json-render-diagnostics`.

Example Result:
image


messense (approved):
bors r+

PR #1592, opened 2023-05-06
Always set `PYO3_PYTHON` if interpreter is runnable regardless of abi3

messense:
It can be useful in build scripts, like https://github.com/pyca/cryptography/pull/8815#issuecomment-1535461786


messense:
bors retry

messense:
bors retry

PR #1603, opened 2023-05-09
Publish maturin itself using trusted publisher

messense:


messense:
image

Works great, I've removed `MATURIN_PASSWORD` from repository secret.

PR #1604, opened 2023-05-11
describe the need for module-name

krpatter-intc:
This discussion was pretty critical for me:

https://github.com/PyO3/maturin/discussions/1540

And helped unblock some usages, so proposing that it get added to the readme here

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ my-project
     └── lib.rs
 ```
 
-When doing this also be sure to set the module name in your code to match the last part of module-name (don't include the package path):
+When doing this also be sure to set the module name in your code to match the last part of `module-name` (don't include the package path):
 
 ```
 #[pymodule]
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ You can specify a different python source directory in `pyproject.toml` by setti
 ```toml
 [tool.maturin]
 python-source = "python"
+module-name = "my_project._lib_name"
 ```
 
 then the project structure would look like this:
@@ -153,7 +154,7 @@ my-project
 ├── my_project
 │   ├── __init__.py
 │   ├── bar.py
-│   └── my_project.cpython-36m-x86_64-linux-gnu.so
+│   └── _lib_name.cpython-36m-x86_64-linux-gnu.so
 ├── README.md
 └── src
     └── lib.rs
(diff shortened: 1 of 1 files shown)


messense:
Thanks, but I dont think it's the right place to add it, better add it in user guide instead.

krpatter-intc:
> Thanks, but I dont think it's the right place to add it, better add it in user guide instead.
> (We can keep the readme change)

Does that mean this PR is ok and I should also submit one for the user guide? or I should close/reject this PR as you dont want it here in the readme ?

messense:
You can push new commits to add it to user guide, and we should mention that you need to also change your `#[pymodule]` name to `_lib_name`.

krpatter-intc:
@messense updated with the need to specify the libname in the module

I'd like to do the User Guide as a separate PR. I think this is enough to help with some bare minimums. I'd like to spend a bit more time ramping on user guide layout in general. There's a few spots where I think an example or two would be helpful for newbies like myself.

README.md
```diff
@@ -153,12 +154,24 @@ my-project
...
 │   ├── __init__.py
 │   ├── bar.py
-│   └── my_project.cpython-36m-x86_64-linux-gnu.so
+│   └── _lib_name.cpython-36m-x86_64-linux-gnu.so
 ├── README.md
 └── src
     └── lib.rs
 ```
 
+When doing this also be sure to set the module name in our code to match the Cargo.toml's `lib.name`
```
messense, line 163: It's actually the last part of `module-name` (split by `.`), `lib.name` in Cargo.toml doesn't matter when using `module-name`.

README.md
```diff
@@ -153,12 +154,24 @@ my-project
...
 │   ├── __init__.py
 │   ├── bar.py
-│   └── my_project.cpython-36m-x86_64-linux-gnu.so
+│   └── _lib_name.cpython-36m-x86_64-linux-gnu.so
 ├── README.md
 └── src
     └── lib.rs
 ```
 
+When doing this also be sure to set the module name in your code to match the last part of module-name (don't include the package path):
```
messense, line 163:
```suggestion
When doing this also be sure to set the module name in your code to match the last part of `module-name` (don't include the package path):
```

messense (approved):

krpatter-intc:
Thanks @messense !

PR #1607, opened 2023-05-12
Warn on missing python-source contents

konstin:
Warns when the user explicitly set `python-source` but that folder doesn't contain a python module.

This came up in ruff (https://github.com/charliermarsh/ruff/pull/4397, https://github.com/charliermarsh/ruff/pull/4399), where the crate name is `ruff_cli` and i didn't realize this takes precedence as module name over the package name.

The extra boolean isn't pretty but i wanted to avoid refactoring the entire method just for that.


messense (approved):

messense:
Just need to resolve the conflict.

PR #1613, opened 2023-05-16
Fix cross-compile to Apple universal2 in the other platform

MisLink:
Since the `--universal2` option is deprecated, only providing `--target=universal2-apple-darwin` cannot set the correct target for Linux or Windows platforms. I suggest that we disregard the current target if the user specifies the `universal2-apple-darwin` target.


messense (approved):
Make sense, thanks!

PR #1614, opened 2023-05-16
Add license classifiers to package metadata

chriskuehl:
This adds standard license classifiers from https://pypi.org/classifiers/ to the package metadata for maturin itself.

This is useful because some internal Python package registries check licenses when importing packages, and the license value of `MIT OR Apache-2.0` generally can't be automatically parsed since it is a freeform text field.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,6 +11,8 @@ classifiers = [
     "Programming Language :: Rust",
     "Programming Language :: Python :: Implementation :: CPython",
     "Programming Language :: Python :: Implementation :: PyPy",
+    "License :: OSI Approved :: MIT License",
+    "License :: OSI Approved :: Apache Software License",
 ]
 dependencies = ["tomli>=1.1.0 ; python_version<'3.11'"]
 dynamic = [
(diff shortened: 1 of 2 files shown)


messense:
License classifiers will be deprecated in [PEP 639](https://peps.python.org/pep-0639/#deprecate-license-classifiers) so I don't want to include them.

`MIT OR Apache-2.0` is not just a freeform text, it's a [SPDX expression](https://spdx.github.io/spdx-spec/v2-draft/SPDX-license-expressions/), surely you can find ways to parse it.

chriskuehl:
> `MIT OR Apache-2.0` is not just a freeform text, it's a [SPDX expression](https://spdx.github.io/spdx-spec/v2-draft/SPDX-license-expressions/), surely you can find ways to parse it.

We could surely parse it for maturin, but as far as I know this isn't any kind of standard used across the ecosystem so it's not something we can use as a general solution. The draft PEP you linked also agrees that it is in general unstructured text: https://peps.python.org/pep-0639/#deprecate-license-field

By contrast the classifiers are standardized so they would be nice for our use-case as they have reliable structured values. But I understand if you don't want to use them.

PR #1619, opened 2023-05-20
Add PEP 517 `config_settings` support

### LINKED ISSUE #1541: PEP 517 `config_settings` support
I think we should explore `config_settings` support for replacing `MATURIN_PEP517_ARGS`, and maybe #1419.

See also https://github.com/pypa/pip/issues/5771

messense:
Replaces `MATURIN_PEP517_ARGS` environment variable with `--config-settings build-args=<cargo build args>`.

Closes #1541

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -11,8 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.15.2] - 2023-05-16
 
- * When determining the python module name, use pyproject.toml `project.name` over Cargo.toml `package.name` in [#1608](https://github.com/PyO3/maturin/pull/1608)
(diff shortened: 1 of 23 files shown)


konstin:
I like replacing environment variables with a PEP mechanism!

Do the frontends most users use (e.g. pip, `python -m build`, tox and nox) all support `--config-settings`? If yes, deprecating `MATURIN_PEP517_ARGS` is the right thing to do, if no, i'd keep both `MATURIN_PEP517_ARGS` and `--config-settings` without any deprecation.

messense:
`pip` and `python -m build` has direct support for `--config-settings`, `nox` can pass `--config-settings` or any other options in `session.install`, `tox` needs to use a customized [`install_command`](https://tox.wiki/en/latest/config.html#install_command) to add `--config-settings` I think.

konstin (approved):

PR #1631, opened 2023-05-26
Upgrade `platform-info` to 2.0, `base64` to 0.21

messense:


messense:
Blocked on https://github.com/uutils/platform-info/issues/39

PR #1644, opened 2023-06-01
Improve error message when python fails in develop

konstin:
Inspired by https://www.youtube.com/live/zepPZ6MFiGs?feature=share&t=648, i wanted to improve the error message when `maturin develop` fails because `python` doesn't work.

I unfortunately failed to reproduce the error in the video, but i'm nonetheless confident that this is the correct change: It shouldn't say "The given list of python interpreters is invalid" when you didn't provide any list. iirc the error message comes from a time when i assumed users would provide a list like python3.5 python3.6 python3.7 instead of building for one version per CI job.

I thought about redoing the error system around interpreters but i think just pointing out the faulty interpreter is all a user needs to understand and debug the problem

### DIFF
--- a/src/build_options.rs
+++ b/src/build_options.rs
@@ -563,6 +563,8 @@ impl BuildOptions {
                     vec![PathBuf::from("python3")]
                 }
             } else {
+                // XXX: False positive clippy warning
+                #[allow(clippy::redundant_clone)]
                 self.interpreter.clone()
             };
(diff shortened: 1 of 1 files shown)


konstin:
looks to me like it's a new lint that's broken, the lint doesn't fire on 1.69 and removing the clone results in a "borrow of partially moved value"

messense (approved):

Issue #1545, opened 2023-03-30
Maturin does not calculate wheel tags the same way as pypa/packaging


timfel:
### Bug Description

With https://github.com/pypa/packaging/pull/607, the official way that pip and wheel calculate wheel tags has changed. It looks like maturin is replicating a different (old) strategy in `get_interpreter_metadata.py`. The generated wheel tags will thus be incompatible and refuse installation of Rust packages e.g. on GraalPy.

### Your maturin version (`maturin --version`)

all

### Your Python version (`python -V`)

all

### Your pip version (`pip -V`)

all

### What bindings you're using

None

### Does `cargo build` work?

- [ ] Yes, it works

### If on windows, have you checked that you aren't accidentally using unix path (those with the forward slash `/`)?

- [ ] Yes

### Steps to Reproduce

Currently no steps to reproduce, as the old way still works for now, but maturin should upgrade to replicate the new behaviour (as e.g. wheel has already upgraded their vendored packaging version, pip is about to, and meson-build already replicated that new behaviour also).

messense:
Thanks for the report. Looks like GraalPy is currently using a very old version of `packaging` (installed via `graalpy -m ginstall install packaging`) so I can't test that change right now:

```python
Python 3.8.5 (Thu Jan 19 11:35:16 PST 2023)
[Graal, GraalVM CE, Java 19.0.2] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import packaging
>>> packaging.__version__
'19.0'
```

timfel:
Hi, this is about the packaging as vendored in wheel, for example. If you pip install the latest wheel, you get a newer version of packaging, and the generated wheels will have the newer tags.

messense:
Thanks! BTW I think the wheel tags is still a bit odd (I'm using arm64 macOS), for example in `graalpy38-graalpy_38_native_aarch64_darwin-macosx_13_0_i386` the `i386` part doesn't make much sense?

```bash
$ graalpy -m pip debug --verbose
WARNING: This command is only meant for debugging. Do not use this with automation for parsing and getting these details, since the output and options of this command may change without notice.
pip version: pip 23.0.1 from /Users/messense/.local/lib/python3.8/site-packages/pip-23.0.1-py3.8.egg/pip (python 3.8)
sys.version: 3.8.5 (Thu Jan 19 11:35:16 PST 2023)
[Graal, GraalVM CE, Java 19.0.2]
[...]

PR #1645, opened 2023-06-02
Add basic support for GraalPy

timfel:
I am working on making PyO3 support GraalPy in https://github.com/timfel/pyo3/tree/timfel/graalpy-support. To run the tests, I noticed I need maturin, so I had to make changes here as well. However, to run the maturin tests, I need the PyO3 changes. 

Right now with this PR and my PyO3 branch, I can run the maturin tutorial example and a small number of PyO3 tests. Since the PyO3 changes will grow, whereas I expect the changes here to not be much more, I would like to open this for discussion.

### DIFF
--- a/src/python_interpreter/get_interpreter_metadata.py
+++ b/src/python_interpreter/get_interpreter_metadata.py
@@ -19,6 +19,23 @@ if platform.python_implementation() == "PyPy":
 else:
     ext_suffix = sysconfig.get_config_var("EXT_SUFFIX")
 
+
+def get_abi_tag():
+    # This should probably return the ABI tag based on EXT_SUFFIX in the same
+    # way as pypa/packaging. See https://github.com/pypa/packaging/pull/607.
+    # For simplicity, we just fix it up for GraalPy for now and leave the logic
+    # for the other interpreters untouched, but this should be fixed properly
+    # down the road.
+    if platform.python_implementation() == "GraalVM":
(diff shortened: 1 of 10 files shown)


src/python_interpreter/get_interpreter_metadata.py
```diff
@@ -19,6 +19,23 @@
 else:
     ext_suffix = sysconfig.get_config_var("EXT_SUFFIX")
 
+
+def get_abi_tag():
+    # This should probably return the ABI tag based on EXT_SUFFIX in the same
+    # way as pypa/packaging. See https://github.com/pypa/packaging/pull/607.
+    # For simplicity, we just fix it up for GraalPy for now and leave the logic
+    # for the other interpreters untouched, but this should be fixed properly
+    # down the road.
```
konstin, line 28: i'd be happy to switch this over to what pypa/packaging does entirely, their logic is authoritative for all other tools anyway

konstin (approved):
Personally, i'm fine with merging this before pyo3 support is finished, especially if this unblocks you on adding support to pyo3. We already have some targets that are supported but not really tested (similar to tier 3 platform in rust in many ways)

src/python_interpreter/get_interpreter_metadata.py
messense: It's tracked here: https://github.com/PyO3/maturin/issues/1545

PR #1657, opened 2023-06-09
Fix emscripten action

ManonMarchand:
Hello and thanks for your software!

## What did not work

In the output of `generate-ci` with the `emscripten` option, the version of python in `maturin_args` was fixed at "3.10". But the stable version of pyodide-build is only compatible with "3.11" right now. 

## What I did

Now the ci fetches the current python version for pyodide with `pyodide config get python_version` similarly to what was already done for the emscripten_version. Like so, it should adapt to any future new versions

## Doubts

I'm new to Rust, so feel free to change anything. Tests are passing and the ci runs there in my library: https://github.com/cds-astro/mocpy/actions/workflows/emscripten.yml but I never know what I could have broken.

### DIFF
--- a/guide/src/develop.md
+++ b/guide/src/develop.md
@@ -29,6 +29,11 @@ Options:
 
           Use as `--extras=extra1,extra2`
 
+      --skip-install
+          Skip installation, only build the extension module inplace
+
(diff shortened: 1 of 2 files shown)


messense (approved):
Thanks!

PR #1670, opened 2023-06-19
Fix crashing on unknown platforms

siepkes:
`platform-info` has an issue which causes it to fail hard on unknown platforms. This issue was fixed in newer versions however the latest released version (`2.0.1`) does not have the fix in it.


messense:
Thanks, but usually we only allow crates.io dependencies, so we need to wait for a new `platform-info` release.

T-256:
@siepkes you can pin `2.0.2` instead of git rev as it's now live on crates.io.

siepkes:
@T-256 Thanks for the heads-up! I've updated the PR.

messense (approved):
Thanks

PR #1696, opened 2023-07-18
remove __init__.py requirement

### LINKED ISSUE #1605: support for implicit namespace packages
There is a check for requiring an `__init__.py` file in the levels of the python source directory.  Can we get that check removed so that maturin can be used as part of packages implementing implicit namespaces ? [PEP420](https://peps.python.org/pep-0420/)

https://github.com/PyO3/maturin/blob/e378ddde4b5566850f16b8a6829f89f3079b2fc7/src/project_layout.rs#L384

krpatter-intc:
Closes #1605


krpatter-intc:
> I'm not opposed to support this, to get it merged we need to add a test case.

Thank you. I'll look in to the existing tests and see if there's something I can leverage.

Marlin-Na:
+1 for this PR. For test, should we just remove the `__init__.py` file in the [pyo3-mixed-src test example](https://github.com/PyO3/maturin/tree/main/test-crates/pyo3-mixed-src/src/pyo3_mixed_src)?

krpatter-intc:
> should we just remove the __init__.py file in the [pyo3-mixed-src test example](https://github.com/PyO3/maturin/tree/main/test-crates/pyo3-mixed-src/src/pyo3_mixed_src)?

I couldn't quite figure out why the original tests were failing online.  I went with this suggestion for now. I could create a new "pyo3-mixed-without-init" based on the "pyo3-mixed" if that is preferred.

krpatter-intc:
> I could create a new "pyo3-mixed-without-init" based on the "pyo3-mixed" if that is preferred.

I went this way and added a new pyo3_mix_implicit test with the specific use case (and the original failures seemed to have gone away after rebasing on main)

krpatter-intc:
Sorry for the push thrash @messense, but I think this is complete. It adds a new test for implicit namespaces and updates the changelog (let me know if squashing is needed in the PR vs. doing in the merge)

PR #1722, opened 2023-08-09
Add non-interactive mode to `upload` command

messense:
Same as `twine upload --non-interactive`.


konstin:
Can we set the same behavior if `CI` is set to true?

messense:
> Can we set the same behavior if `CI` is set to true?

https://github.com/PyO3/maturin/pull/1723

PR #1723, opened 2023-08-09
Defaults to non interactive mode when running on CI

messense:
https://github.com/PyO3/maturin/pull/1722#issuecomment-1670698622


konstin:
thanks!

PR #1728, opened 2023-08-11
Handle renamed Rust dependency in sdist

messense:
Fixes https://github.com/PyO3/maturin/issues/1083#issuecomment-1674085413

cc @kylebarron


kylebarron:
Thanks! You beat me to making the issue 😅

PR #1763, opened 2023-09-03
Pre commit

mbway:
maturin has various checks that run on the CI so it would be helpful for contributors if they could easily run the same checks locally so they can find out if there is a problem before their code reaches the CI.

[pre-commit](https://pre-commit.com/) is a tool used by many projects to run checks before every commit (and can also be run manually with `pre-commit run --all`)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,65 @@
...
+        pass_filenames: false
+        types: [rust]
+      # cargo install --locked cargo-deny
+      - id: cargo-deny
+        name: cargo deny
+        entry: cargo deny --all-features check --
+        language: system
+        pass_filenames: false
+      # rustup component add clippy
+      - id: cargo-clippy
```
messense, line 24: I don't think we should add `cargo check` and `cargo clippy` in pre-commit hooks because they may take a long time which makes the git commit UX bad.

mbway: I was able to set these checks to run only for the 'manual' stage which means only when pre-commit is run with `pre-commit run --hook-stage manual --all`.
It's debatable whether another task runner like makefile, [cargo make](https://github.com/sagiegurari/cargo-make) or [just](https://github.com/casey/just) would be easier to use for this, but at least pre-commit is able to serve as a commit hook and also as a way to fairly easily run all the checks at once with a single tool.

messense (approved):
Thanks!

messense:
FYI, I've also set up [pre-commit.ci](https://pre-commit.ci/) for this repo.

PR #1793, opened 2023-10-02
Warn about incorrect version in `build-system.requires`

### LINKED ISSUE #1792: `build` command ignores `build-system.requires` specifier
### Bug Description

My `pyproject.toml` contains the following snippet:

```toml
[build-system]
requires = ["maturin==1.2.1"]
build-backend = "maturin"
```

I have the latest maturin installed (`1.2.3`). This does not match the build requirement. Yet the build still completes without any issue.
I have tried different combinations of version specifiers (minimum, maximum, exact versions), but I cannot get the build to fail.

I would expect an error to be thrown.

1.2.3

3.11.2

23.2.1

pyo3

### Does `cargo build` work?

- [X] Yes, it works

- [ ] Yes

### Steps to Reproduce

Set the `build-system.requires` entry to not match the installed version of maturin, then run `maturin build`.

konstin:
Maturin should warn if the version in the pyproject.toml build system requirements is incompatible with the one that is currently running (#1792). We only warn instead of erroring to allow e.g. testing a prerelease version of maturin with a project without having to edit the requires.

pyproject.toml:

```toml
[build-system]
requires = ["maturin==0.0.1"]
build-backend = "maturin"
```

Output:

```
⚠️  Warning: You specified maturin ==0.0.1 in pyproject.toml under `build-system.requires`, but the current maturin version is 1.2.3
```

Fixes #1792.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+* Warning about incorrect maturin version pyproject.toml `[build-system] requires`
+
 ## [1.2.3] - 2023-08-17
 
 * Fix sdist build failure with workspace path dependencies by HerringtonDarkholme in [#1739](https://github.com/PyO3/maturin/pull/1739)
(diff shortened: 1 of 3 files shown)


messense (approved):
Thanks!

PR #1800, opened 2023-10-09
Add trusted publishing section to the user guide

Phil-V:
Hi and thank you for maturin, it's great!

I couldn't figure out how to use the workflow provided by `generate-ci` with PyPI's trusted publishing, but found the solution in issue https://github.com/PyO3/maturin/issues/1575#issuecomment-1702761101. Here's a PR to add a short section about it to the distribution page in the user guide. Feedback is welcome.

PS: I've also toyed with making trusted publishing the default, but I don't know if this is desirable as some users will have different registries than PyPI. If you prefer, I can make a different PR with [those changes](https://github.com/PyO3/maturin/compare/main...Phil-V:maturin:trusted-publishing) after doing some further testing.

### DIFF
--- a/guide/src/distribution.md
+++ b/guide/src/distribution.md
@@ -289,3 +289,13 @@ Options:
   -h, --help
           Print help information (use `-h` for a summary)
 ```
+
+### Using PyPI's trusted publishing
+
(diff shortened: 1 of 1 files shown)


messense (approved):
Thanks!

Personally I don't think we should make trusted publishing the default because it requires much more efforts to set up. However we could add an option like `--trusted-publisher` to the `generate-ci` command for users to opt-in to trusted publishing, pull requests are welcome!

PR #1802, opened 2023-10-10
Fix wheel filename for GraalPy

msimacek:
The wheel filename needs to separate the ABI from the platform with a dash, otherwise pip fails to parse the name.

CC @timfel


messense (approved):
Thanks!

PR #1808, opened 2023-10-17
Keep trailing newlines from templates

nils-werner:
Both rustfmt and pycodestyle/PEP8 require a trailing newline in code files. These newlines are produced by most code editors and are there in the code templates, however `minijinja` is configured to remove any trailing newlines. This means maturin produces templates that don't adhere to the code styles in both Rust an Python.

This PR

 1. Bumps the `minijinja` requirement to `1.0.7` (that's when the [option was introduced](https://github.com/mitsuhiko/minijinja/commit/53da32be602bc50fde4c3d603aba4dcd74bd314d))
 1. Sets [`Environment::set_keep_trailing_newline(true)`](https://docs.rs/minijinja/latest/minijinja/struct.Environment.html#method.set_keep_trailing_newline)

I'm relatively new to Rust and Cargo, and I am a bit worried by the massive change in `Cargo.lock`. Is this supposed to happen, or did I do something wrong?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1170,9 +1170,9 @@ dependencies = [
 
 [[package]]
 name = "minijinja"
-version = "1.0.6"
+version = "1.0.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e436cdb19900a08197c6b874280d24f9289d127275eadc23623276e0f5ec97e9"
+checksum = "80084fa3099f58b7afab51e5f92e24c2c2c68dcad26e96ad104bd6011570461d"
 dependencies = [
  "serde",
 ]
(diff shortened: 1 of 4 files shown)


messense:
> I'm relatively new to Rust and Cargo, and I am a bit worried by the massive change in `Cargo.lock`. Is this supposed to happen, or did I do something wrong?

Don't run `cargo update`, `cargo build` will take care of it automatically when `Cargo.toml` changed.

nils-werner:
> Don't run `cargo update`, `cargo build` will take care of it automatically when `Cargo.toml` changed.

Ah, thanks! I've force-pushed an amendment to get rid of the mess.

messense (approved):
Thanks!

PR #1828, opened 2023-10-30
use file_name to get bindings manifest name

nreinicke:
Attempts to update how the python binding crate name is parsed by just taking the last member in the path rather than removing a given prefix. 

With the existing code, in the new test, `pyo3-mixed-workspace`, the `main_member_name` gets parsed as `rust/pyo3-mixed-workspace-py` and then does not get added into the resulting workspace Cargo.toml file in the source distribution. This causes a failure when trying to install the package from the resulting source distribution. 

The new test is based on this project structure: https://github.com/NREL/routee-compass


src/source_distribution.rs
```diff
@@ -456,7 +456,7 @@ fn add_cargo_package_files_to_sdist(
             let mut deps_to_keep = known_path_deps.clone();
             // Also need to the main Python binding crate
             let main_member_name = abs_manifest_dir
-                .strip_prefix(&sdist_root)
+                .file_name()
```
messense, line 459: What if `pyo3-mixed-workspace-py` is located at `rust/python/pyo3-mixed-workspace`? Then in workspace `Cargo.toml`, `members` would be

```toml
members = [
    "pyo3-mixed-workspace",
    "python/pyo3-mixed-workspace",
]
```

Only use the file name would be wrong? I think we maybe it should be changed to `strip_prefix(&workspace_root)`?

nreinicke: Yes, that's a great point. I just updated to strip the `workspace_root` prefix and updated the test to reflect the nested case you described and it seems to be working.

nreinicke:
Sorry, I accidentally closed this by reference in a different PR

messense:
The test failure on Windows is probably related to `/` v.s. `\`.

PR #1879, opened 2023-12-11
Allow identical `VIRTUAL_ENV` and `CONDA_PREFIX` env vars

### LINKED ISSUE #1878: Compatibility with conda and poetry used simultaneously
When a project uses both maturin and poetry simultaneously, it's common to run commands such as `poetry run maturin develop`.

However, this doesn't work in a conda environment. Here's an example using the [robyn](https://github.com/sparckles/Robyn) repository:
```shell
$ git clone https://github.com/sparckles/Robyn.git
[...]
$ cd Robyn
$ conda create -n robyn python
[...]
$ conda activate robyn
$ poetry install
[...]
± poetry run maturin develop
💥 maturin failed
  Caused by: Both VIRTUAL_ENV and CONDA_PREFIX are set. Please unset one of them
```
This is because `poetry` seems to set the `VIRTUAL_ENV` environment variable when it detects a conda env:
```
$ env
[...]
CONDA_PREFIX=/home/vinmic/miniconda/envs/parsecv3
[...]
# No VIRTUAL_ENV defined here
$ poetry run env
[...]
CONDA_PREFIX=/home/vinmic/miniconda/envs/robyn
[...]
VIRTUAL_ENV=/home/vinmic/miniconda/envs/robyn
```

Would it be possible to allow both environment variables to be set when their values are equal?

konstin:
Fixes #1878

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -191,6 +191,9 @@ fn detect_venv(target: &Target) -> Result<PathBuf> {
     match (env::var_os("VIRTUAL_ENV"), env::var_os("CONDA_PREFIX")) {
         (Some(dir), None) => return Ok(PathBuf::from(dir)),
         (None, Some(dir)) => return Ok(PathBuf::from(dir)),
+        (Some(venv), Some(conda)) if venv == conda => {
+            return Ok(PathBuf::from(venv))
+        }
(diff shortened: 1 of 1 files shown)


messense (approved):
👍

PR #1909, opened 2024-01-09
Support uniffi-bindgen in cargo workspaces

### LINKED ISSUE #1905: Uniffi-bindgen not detected in cargo workspace
### Bug Description

Maturin includes support for workspace bundled uniffi-bindgen in https://github.com/PyO3/maturin/blob/09e4f55b0d1636e7494b8654dfa5f05883d0dba7/src/module_writer.rs#L876-L882

However this doesn't seem to work correctly and instead I get the following error message.
```
💥 maturin failed
  Caused by: Failed to run uniffi-bindgen, did you install it? Try `pip install uniffi-bindgen`
```

I can resolve it by adding my target dir to the path, but this "should" be unnecessary with the above logic.

1.4.0

3.9.6

21.2.4

uniffi

### Does `cargo build` work?

- [X] Yes, it works

- [ ] Yes

### Steps to Reproduce

2. Run `maturin build`, note it fails due to `unffi-bindgen` not being in path.

Hinton:
Uniffi suggests using a separate crate for uniffi-bindgen in a multi-crate workspace, https://mozilla.github.io/uniffi-rs/tutorial/foreign_language_bindings.html#multi-crate-workspaces. This workflow is currently not supported in maturin.

This PR expands the `uniffi_bindgen_command` to evaluate if the workspace has a bindgen package, and if so it runs the bindgen command from the root of the workspace. This seems to work correctly in my project https://github.com/Hinton/maturin-repro.

Resolves #1905

### DIFF
--- a/src/module_writer.rs
+++ b/src/module_writer.rs
@@ -880,13 +880,27 @@ fn uniffi_bindgen_command(crate_dir: &Path) -> Result<Command> {
                 .any(|target| target.name == "uniffi-bindgen" && target.is_bin())
         })
         .unwrap_or(false);
+    let has_uniffi_bindgen_workspace_package = cargo_metadata.packages.iter().any(|pkg| {
+        pkg.targets
+            .iter()
+            .any(|target| target.name == "uniffi-bindgen" && target.is_bin())
+    });
+
     let command = if has_uniffi_bindgen_target {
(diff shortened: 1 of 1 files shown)


messense (approved):
LGTM, thanks!

PR #1910, opened 2024-01-09
PoC for supporting uniffi bindgen with multiple crates

### LINKED ISSUE #1904: Uniffi multiple crates not supported
### Bug Description

Uniffi supports using types from dependent crates, https://mozilla.github.io/uniffi-rs/proc_macro/index.html#types-from-dependent-crates. Consider a crate `a` which depends on `b`. This results in two python files being generated, `a.py` and `b.py`.

The types exposed by crate `a` are exported in `a.py`, and `b` in `b.py`. In order to correctly use the library both python files needs to be imported.

The sample repository https://github.com/Hinton/maturin-repro produces the following wheel:

```
a.whl
  a/
    __init__.py
    a.py
    liba.dylib
  a-0.1.0.dist-info/
```

As noted, `b.py` is missing which is referenced by `a.py`.

1.4.0

3.9.6

21.2.4

uniffi

### Does `cargo build` work?

- [X] Yes, it works

- [ ] Yes

### Steps to Reproduce

2. Run `maturin build`
3. Unpack `target/wheels/a-0.1.0-py3-none-macosx_11_0_arm64.whl`
4. Note missing `b.py`.
5. Attempt to run and get an error complaining about `b` missing since it's imported from `a.py`.

Hinton:
Uniffi supports using types from dependent crates, https://mozilla.github.io/uniffi-rs/proc_macro/index.html#types-from-dependent-crates. Consider a crate `a` which depends on `b`. This results in two python files being generated, `a.py` and `b.py`.

The previous solution relied on manually building the expected bindgen filename, which worked well with just a single python output. This atlernative approach relies on finding all python files in the target directory and including them in the wheel.

At the current stage this is a fairly brute force approach which resolves my use case. I would appreciate feedback on the general approach.

Resolves #1904

### DIFF
--- a/src/module_writer.rs
+++ b/src/module_writer.rs
@@ -14,6 +14,7 @@ use fs_err::File;
 use ignore::overrides::Override;
 use ignore::WalkBuilder;
 use indexmap::IndexMap;
+use itertools::Itertools;
 use normpath::PathExt as _;
 use sha2::{Digest, Sha256};
 use std::collections::{HashMap, HashSet};
@@ -859,9 +860,9 @@ struct UniFfiBindingsConfig {
 
 #[derive(Debug, Clone)]
 struct UniFfiBindings {
-    name: String,
+    bindings: Vec<String>,
     cdylib: String,
-    path: PathBuf,
+    dir: PathBuf,
 }
 
 fn uniffi_bindgen_command(crate_dir: &Path) -> Result<Command> {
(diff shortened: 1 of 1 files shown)


Nickersoft:
@Hinton The fact this hasn't landed is causing https://github.com/PyO3/maturin/issues/2294 for me :( Any chance these changes will get in?

Hinton:
@Nickersoft unfortunately I haven't had time to work on this since I opened the PR. If I remember correctly I believe some tests failed in a non trivial way. I was hoping to get a initial review on if the changes were appropriate as I'm not familiar with the Maturin codebase.

At the moment I'm not actively working on anything producing python bindings either, which means I probably won't be able to prioritize this PR in the near future. If anyone wants to continue on my rough PoC they have my permission.

Nickersoft:
I might be able to help! I think a good chunk of this PR was already landed in #2208, short of updating the __init__ file correctly, which was causing my issue. So this is good info to know – I'll see if I can get a PR open, as it's been a huge blocker for me for months :)

Nickersoft:
Opened #2294 😊

PR #1918, opened 2024-01-21
Use extension name as library name, instead of hardcoded 'native.so'

### LINKED ISSUE #1919: [cffi] Allow generated shared library to have a name other than `native.so`
When using the `cffi` backend, the generated shared library is always named `native.so`. It would be good if this:

- was configureable or
- matched the name of the library (e.g. `my_extension` would generated `my_extension.so` or `libmy_extension.so`)

timkpaine:
Opening this PR for some conversation, I'm hoping to distribute an extension via CFFI, but it's a bit awkward integrating it with other projects when the library name is always `native.so`. It also makes it difficult to link against multiple.

In this PR I've just slightly tweaked the generation bits to make the library name the same as the extension name, so instead of `native.so` it will have the same name as the outer extension folder. I'm tempted to make it `lib<extension_name>.so` as it's not an importable python module, so seems like it should use the standard library prefix?

partially fixes #1919

### DIFF
--- a/src/module_writer.rs
+++ b/src/module_writer.rs
@@ -508,15 +508,17 @@ fn entry_points_txt(
 }
 
 /// Glue code that exposes `lib`.
-fn cffi_init_file() -> &'static str {
-    r#"__all__ = ["lib", "ffi"]
+fn cffi_init_file(extension_name: &str) -> String {
+    format!(
+        r#"__all__ = ["lib", "ffi"]
 
 import os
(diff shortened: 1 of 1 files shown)


messense (approved):
Looks fine to me.

messense:
> I'm tempted to make it `lib<extension_name>.so` as it's not an importable python module, so seems like it should use the standard library prefix?

IMO it's not a big issue.

Issue #1398, opened 2023-01-08
`maturin init/new` should validate the package name


not-my-profile:
Package names of Rust libraries and package names of Python libraries both need to fulfill certain criteria. `maturin init/new` apparently currently doesn't check for any of these criteria e.g. `maturin init "foo bar"` works fine even though neither Rust nor Python support package names with a space.

I think ideally maturin would check the supplied name against two regexes (one for Rust package names and one for Python package names) and report which check failed in case the supplied name doesn't match.

Note that I think PyPI disallows certain package names like names of modules in the standard library ... it would be nice if maturin could check for these as well but that might not be easily doable (see also https://github.com/pypi/warehouse/pull/8533).

messense:
Agreed it's nice to have, pull requests are welcome.

messense:
@mahtantaoui No need to ask permission to work on issues, feel free to send pull requests, thanks!

sanmai-NL:
See also: https://github.com/rust-lang/cargo/pull/12766

alonme:
> Can we validate the package name using a combination of an approach similar to the one in [Cargo](https://github.com/rust-lang/cargo/blob/master/src/cargo/util/restricted_names.rs) for Rust names and the [constraint](https://github.com/pypi/warehouse/blob/main/warehouse/packaging/models.py#L168-L171) in PYPI for Python?

[permalink](https://github.com/pypi/warehouse/blob/8f79d90a310f0243ab15f52c41de093708a61dfd/warehouse/packaging/models.py#L211C9-L214C10) to the pypi/warehouse name constraint

PR #1943, opened 2024-02-17
Add validation for crate/package name in new/init

alonme:
partially solves https://github.com/PyO3/maturin/issues/1398  (only cargo)

The current commit only adds a very basic validation as a POC before implementing the needed cargo / pypi checks
A few questions i have before finishing the implementation

1. The cargo validation seems to be comprised of 2 main parts, 
    a. [Basic validation](https://github.com/rust-lang/cargo/blob/7b7af3077bff8d60b7f124189bc9de227d3063a9/crates/cargo-util-schemas/src/restricted_names.rs#L42C1-L42C11) which is part of the crate [cargo-util-schemas](https://crates.io/crates/cargo-util-schemas/versions), and is not made public outside the crate

    b. [Extra validations](https://github.com/rust-lang/cargo/blob/master/src/cargo/util/restricted_names.rs) that are public from the cargo package

    I believe we should vendor these logics - one is private, and the other requires the cargo package which i believe will add many dependencies to the project 

2. The `generate_project` function is de-facto defining a default value for `GenerateProjectOptions.name` - would it be better to move this logic into `GenerateProjectOptions`?

### DIFF
--- a/src/package_name_validations.rs
+++ b/src/package_name_validations.rs
@@ -1,0 +1,128 @@
+// Based on: https://github.com/rust-lang/cargo/blob/e975158c1b542aa3833fd8584746538c17a6ae55/src/cargo/ops/cargo_new.rs#L169
+pub fn cargo_check_name(name: &str) -> anyhow::Result<()> {
+    // Instead of `PackageName::new`, which performs these checks
+    validate_package_name(name)?;
+
+    if is_keyword(name) {
+        anyhow::bail!(
+            "the name `{}` cannot be used as a package name, it is a Rust keyword",
+            name,
+        );
+    }
+    if is_conflicting_artifact_name(name) {
+        anyhow::bail!(
+            "the name `{}` cannot be used as a package name, \
+                it conflicts with cargo's build directory names",
+            name,
+        );
+    }
+    if name == "test" {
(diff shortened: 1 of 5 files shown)


messense:
> 1. I believe we should vendor these logics - one is private, and the other requires the cargo package which i believe will add many dependencies to the project

Agreed.

> 2\. The `generate_project` function is de-facto defining a default value for `GenerateProjectOptions.name` - would it be better to move this logic into `GenerateProjectOptions`?

Makes sense.

src/package_name_validations.rs
```diff
@@ -0,0 +1,116 @@
...
+        );
+    }
+    if ["core", "std", "alloc", "proc_macro", "proc-macro"].contains(&name) {
+        eprintln!(
+            "⚠️  Warning: the name `{}` is part of Rust's standard library\n\
+            It is recommended to use a different name to avoid problems.",
+            name,
+        );
+    }
+    if is_windows_reserved(name) {
```
alonme, line 32: The original code bailed if the platform is actually windows, I think this should be good enough for now

alonme:
@messense I believe this is ready for review 😄

alonme:
@messense Thanks!

PR #1958, opened 2024-02-25
Isolated import hook changes

mbway:
I have been working on `maturin_import_hook` in [my fork](https://github.com/mbway/maturin-import-hook), sorry it's taken so long. I'm nearly finished. This PR consists of the changes in the original PR https://github.com/PyO3/maturin/pull/1748 that apply to maturin directly instead of the import hook.

The main change here is that packages installed with `maturin develop` are installed in editable mode so they are suitable for use with the import hook. I'm thinking that the initial release of the import hook can also make the necessary edits after calling `maturin develop` and in future releases the minimum supported maturin version can be raised.


guide/src/local_development.md
```diff
@@ -117,15 +117,26 @@ requires = ["maturin>=1.0,<2.0"]
 build-backend = "maturin"
```
messense, line 117: `SUMMARY.md` also needs to be updated to link to the new file name.

messense (approved):
Thanks!

PR #1969, opened 2024-02-27
Add 32-bit RISC-V support

kraj:
Tested with qemuriscv32 and it builds fine with all tests passed on a qemu machine.


konstin (approved):

messense:
Thanks!

PR #1978, opened 2024-03-04
Only set rustflags when necessary

taiki-e:
Similar to https://github.com/PyO3/maturin/pull/887, but handles more cases.

As explained in that PR, explicitly setting rustflags has several side effects.
A concrete example of where this has caused problems is https://github.com/taiki-e/portable-atomic/issues/151, but I know of several other instances where setting rustflags might cause problems: https://github.com/taiki-e/cargo-llvm-cov/issues/332, https://github.com/taiki-e/cargo-llvm-cov/issues/212 (Those are examples from cargo-llvm-cov, but I think similar problems could occur in maturin).


messense (approved):
Thanks!

PR #2001, opened 2024-03-21
Add MacOS m1 Runner & move target aarch64 to this runner

### LINKED ISSUE #1952: Use macOS M1 runner in generated GitHub Actions configuration
No need to cross compiling to macOS arm64 on x86_64 now that GitHub Actions has macOS M1 runners (`macos-14`).

https://github.com/PyO3/maturin/blob/b209c4b7b10f5bcf6141c85ed26999401568b2fa/src/ci.rs#L213-L222

Owen-CH-Leung:
fixes #1952 

This PR enhances the `generate-ci` command for github actions. Specifically, it addes a new CI job to be running on macos m1 runner, and move away cross compilation to aarch64 from the original mac runner.

### DIFF
--- a/src/ci.rs
+++ b/src/ci.rs
@@ -32,11 +32,18 @@ pub enum Platform {
     Macos,
     /// Emscripten
     Emscripten,
+    /// macOSM1
+    MacosM1,
 }
 
 impl Platform {
     fn defaults() -> Vec<Self> {
-        vec![Platform::Linux, Platform::Windows, Platform::Macos]
+        vec![
+            Platform::Linux,
+            Platform::Windows,
+            Platform::Macos,
+            Platform::MacosM1,
+        ]
     }
 
     fn all() -> Vec<Self> {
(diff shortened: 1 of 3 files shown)


src/ci.rs
```diff
@@ -32,11 +32,18 @@ pub enum Platform {
     Macos,
     /// Emscripten
     Emscripten,
+    /// macOSM1
+    MacosM1,
```
messense, line 36: For future proof, how about naming this `MacosArm64`?

Owen-CH-Leung: Thanks. Renamed & fixed the failing CI

messense (approved):
Thanks!

I might tweak it a bit later to merge macos arm64 with macos using a different matrix like:

```yaml
matrix:
  platform:
    - os: macos-latest
      target: x86_64
    - os: macos-14
      target: aarch64
```

Feel free take this refactoring task if you want to.

PR #2005, opened 2024-03-23
Add JSON schema generation

### LINKED ISSUE #1995: Add JSON schema generation for pyproject.toml with schemars for SchemaStore
Adding an automatic JSON schema generation of the maturin configuration in pyproject.toml would allow update SchemaStore easily and enable autocompletion and validation in pyproject.toml.
Ruff could be taken as an example, as it has an automatic JSON schema generation that use [schemars](https://github.com/GREsau/schemars).
Schemars only need to add some macro on the configuration struct.
The generation would be available through a feature.
I am willing and able to implement it.

This feature has been requested on the SchemaStore repo.
- SchemaStore/schemastore#3564

JonathanPlasse:
- Fix #1995

This PR implements the Maturin JSON schema for the section `tool.maturin` in `pyproject.toml` files.

To run the generation, you can use the following command:
```sh
cargo run --bin generate_json_schema --features schemars
```

I use Ruff [generate_json_schema.rs](https://github.com/astral-sh/ruff/blob/0a99bd84ce90564b10dbb7c3cf51c70922f4daed/crates/ruff_dev/src/generate_json_schema.rs) as a base to implement the JSON schema generation.

Remaining tasks:
- [ ] Integrate the generation better
- [ ] Check in CI if the Maturin schema is up to date


messense:
Thanks, IMO it's better to be placed in a separated crate that depends on `maturin` instead, we don't want two `bin` targets in `maturin` crate.

konstin:
Can we make this a subcommand instead of a separate crate? I'd avoid splitting out into a new crate over helper functionality

konstin:
I've just forked https://github.com/PyO3/schemastore so we can port (copy?) https://github.com/astral-sh/ruff/blob/0c194f55e81c217b7ac63e7136ea3aabffd7c127/scripts/update_schemastore.py for maturin too

konstin:
With switched to a (hidden?) subcommand and CI passing i'm r+ on the code.

JonathanPlasse:
I switched to the JSON generation to a hidden sub-command.
I would like to merge this and leave the update script for a following PR.

messense (approved):
LGTM, thank you!

messense:
> leave the update script for a following PR.

It seems that the schema generation currently does not put a newline in the end, which triggers the pre-commit autofixes. We should get this fixed when adding schema checks in CI.

PR #2006, opened 2024-03-23
feat: add allow prerelease flag

fatelei:


src/build_options.rs
```diff
@@ -175,6 +175,10 @@ pub struct BuildOptions {
     #[arg(short = 'f', long, conflicts_with = "interpreter")]
     pub find_interpreter: bool,
 
+    /// Allow prerelease python version
+    #[arg(short, long)]
```
messense, line 179:
```suggestion
    #[arg(long)]
```

Let's drop the short form.

src/build_options.rs
```diff
@@ -560,6 +564,15 @@ impl BuildOptions {
         let generate_import_lib = is_generating_import_lib(&cargo_metadata)?;
         let interpreter = if self.find_interpreter {
             // Auto-detect interpreters
+            let requires_python = metadata23.requires_python.as_ref();
+            if requires_python.is_some() {
+                let versions = requires_python.unwrap();
+                let has_pre_release = versions.iter().any(|v| v.any_prerelease());
+                if has_pre_release && !self.allow_prereleases {
+                    bail!("⚠️  Warning: python version is pre release, need pass flag --allow-prereleases");
```
messense, line 572: I think we should just skip it and print the warning if `allow_prereleases` isn't set, no need to bail.

fatelei: ok

fatelei: code changed, but i not sure what i do is right, i not familiar with rust.

src/build_options.rs
```diff
@@ -560,6 +563,16 @@ impl BuildOptions {
         let generate_import_lib = is_generating_import_lib(&cargo_metadata)?;
         let interpreter = if self.find_interpreter {
             // Auto-detect interpreters
+            let requires_python = metadata23.requires_python.as_ref();
```
messense, line 566: How about moving this check after L588, and use it as a filter to [retain](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.retain) only non-prerelease versions? Something like this:

```rust
// existing code
let mut interpreter = if self.find_interpreter { ... } else { ... };
interpreter.retain(|interp| {
    // your added code here
});
```

src/build_options.rs
```diff
@@ -560,6 +563,16 @@ impl BuildOptions {
...
+            if requires_python.is_some() {
+                let versions = requires_python.unwrap();
+                let has_pre_release = versions.iter().any(|v| v.any_prerelease());
+                if has_pre_release && !self.allow_prereleases {
+                    print!("⚠️  Warning: python version is pre release, need pass flag --allow-prereleases");
```
messense, line 571: Usually we use `eprintln!` to print to stderr.

src/build_options.rs
```diff
@@ -6,10 +6,10 @@ use crate::project_layout::ProjectResolver;
 use crate::pyproject_toml::ToolMaturin;
 use crate::python_interpreter::{InterpreterConfig, InterpreterKind, MINIMUM_PYTHON_MINOR};
 use crate::{BuildContext, PythonInterpreter, Target};
-use anyhow::{bail, format_err, Context, Result};
+use anyhow::{bail, format_err, Context, Ok, Result};
```
messense, line 9:
```suggestion
use anyhow::{bail, format_err, Context, Result};
```

src/build_options.rs
```diff
@@ -175,6 +175,9 @@ pub struct BuildOptions {
     #[arg(short = 'f', long, conflicts_with = "interpreter")]
     pub find_interpreter: bool,
 
+    #[arg(long)]
```
messense, line 178:
```suggestion
    /// Allow prerelease python interpreters
    #[arg(long)]
```

src/build_options.rs
```diff
@@ -587,6 +591,16 @@ impl BuildOptions {
             self.find_interpreters(&bridge, &interpreter, &target, None, generate_import_lib)?
         };
 
+        if !self.allow_prereleases {
+            interpreter.retain(|interp| {
+                let version = Version::new([interp.config.major as u64, interp.config.minor as u64]);
```
messense, line 596: This won't be enough, I think it also need patch version.

messense:
```
Python 3.13.0a6 (main, Apr 10 2024, 01:20:43) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.version_info
sys.version_info(major=3, minor=13, micro=0, releaselevel='alpha', serial=6)
>>> sys.version
'3.13.0a6 (main, Apr 10 2024, 01:20:43) [GCC 12.2.0]'
```

fatelei: ok

fatelei: interp config only has major and minor, missing micro，we should pass minor to it?

messense: You can modify interp config to add more detailed version info.

messense: Perhaps we can remove `major` and `minor`, add `version: pep440_rs::Version` instead to interpreter config and maybe other places. I might give it a try on the weekend.

fatelei: ok

kmk142789 (approved):

PR #2007, opened 2024-03-23
Add pypi name validation

### LINKED ISSUE #1398: `maturin init/new` should validate the package name

alonme:
closes #1398

### DIFF
--- a/src/new_project.rs
+++ b/src/new_project.rs
@@ -1,4 +1,4 @@
-use self::package_name_validations::cargo_check_name;
+use self::package_name_validations::{cargo_check_name, pypi_check_name};
 use crate::ci::GenerateCI;
 use crate::BridgeModel;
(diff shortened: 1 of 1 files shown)


messense (approved):
Thanks!

PR #2010, opened 2024-03-26
Detect compiling from Linux gnu to Linux musl as cross compiling

messense:
The original code logic was derived from pyo3, I don't remember why it does not consider cross-compiling if arch-vendor-os is all the same.

For maturin to correctly use bundled sysconfig Python when compiling from Linux gnu host to Linux musl target, we need to detect it as cross compiling, otherwise it will only find Python interpreters on the host that was compiled for glibc.

cc @davidhewitt

### DIFF
--- a/src/cross_compile.rs
+++ b/src/cross_compile.rs
@@ -27,16 +27,9 @@ pub fn is_cross_compiling(target: &Target) -> Result<bool> {
         // Not cross-compiling to compile for 32-bit Python from windows 64-bit
         return Ok(false);
     }
-
-    if let Some(target_without_env) = target_triple
-        .rfind('-')
-        .map(|index| &target_triple[0..index])
-    {
-        if host.starts_with(target_without_env) {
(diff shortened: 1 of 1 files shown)


davidhewitt:
> I don't remember why it does not consider cross-compiling if arch-vendor-os is all the same.

Probably this was because Windows 32/64bit and macOS arm/x86_64 are somewhat special cases. I wonder whether the logic is backwards though, something to check in PyO3.

davidhewitt:
EDIT: rereading, I think my comment above is not relevant.

As far as I was concerned in PyO3 my thinking what that cross compiling is "can't run the target interpreter to discover config". I think Linux gnu systems can run musl binaries (but maybe only if the musl binary is statically linked?)

messense:
> I think Linux gnu systems can run musl binaries (but maybe only if the musl binary is statically linked?)

Yeah, I think currently the only major player is Alpine Linux which is dynamically linked musl libc.

PR #2015, opened 2024-03-28
Add uv as develop backend command

### LINKED ISSUE #1959: option to use uv
It may be beneficial to offer [uv](https://github.com/astral-sh/uv) as a backend for `maturin develop` rather than hard-coding `pip` as the only available backend.

I was thinking about this while thinking of ways to speed up the import hook tests as a significant part of those tests is waiting for `maturin develop`. Compilation still probably takes the majority of the time but there may be a small benefit.

`uv pip` offers a pip-compatible interface for installing packages but may be faster than using pip. Since it's written in rust perhaps it could be bundled into maturin directly as a library?

dmatos2012:
Hi, 
This should close #1959. However, I would like some feedback/help before I imagine this is ready to merge. I plan on adding more tests to see that it works accross crates etc,  but got into a weird test error loop, that I hope someone can help me here.

Basically about making the `develop_hello_world_uv` test pass. If I activate my `venv` then test fails with `no-such file or directory hello-world`. If I dont activate it, then it passes. Sadly, if I `deactivate` then the other tests dont pass as I need `cffi` to have my tests pass, what gives? 

Doing a `/target/debug/maturin develop --uv` on 3-4 crates from `test-crates` worked fine, thus not sure whats with the test harness that doenst make it happy. 

If anyone else more experienced can help me out, it would be great. Maybe not the best place for long comment here but didnt know where else :). Thanks for your time

### DIFF
--- a/src/develop.rs
+++ b/src/develop.rs
@@ -16,21 +16,16 @@ use std::process::Command;
 use tempfile::TempDir;
 use url::Url;
 
-#[derive(PartialEq, Eq, Clone)]
-struct Pip {
-    path: Option<PathBuf>,
-}
-
 #[derive(PartialEq, Eq)]
 enum InstallBackend {
-    Pip(Pip),
+    Pip { path: Option<PathBuf> },
     Uv,
 }
 
 impl InstallBackend {
     fn make_command(&self, python_path: &Path) -> Command {
         match self {
-            InstallBackend::Pip(pip) => match &pip.path {
+            InstallBackend::Pip { path } => match &path {
                 Some(path) => {
                     let mut cmd = Command::new(path);
                     cmd.arg("--python")
@@ -185,7 +180,7 @@ fn pip_install_wheel(
     fix_direct_url(
         build_context,
         python,
-        &InstallBackend::Pip(Pip { path: pip_path }),
+        &InstallBackend::Pip { path: pip_path },
     )?;
     Ok(())
 }
@@ -263,9 +258,9 @@ pub fn develop(develop_options: DevelopOptions, venv_dir: &Path) -> Result<()> {
     let install_backend = if uv {
         InstallBackend::Uv
     } else {
-        InstallBackend::Pip(Pip {
+        InstallBackend::Pip {
             path: pip_path.clone(),
-        })
+        }
     };
     let mut target_triple = cargo_options.target.as_ref().map(|x| x.to_string());
     let target = Target::from_target_triple(cargo_options.target)?;
--- a/src/develop.rs
+++ b/src/develop.rs
@@ -16,6 +16,10 @@ use std::process::Command;
 use tempfile::TempDir;
 use url::Url;
 
+enum Backend {
+    Pip,
+    Uv,
+}
 /// Install the crate as module in the current virtualenv
 #[derive(Debug, clap::Parser)]
 pub struct DevelopOptions {
@@ -57,20 +61,30 @@ pub struct DevelopOptions {
     /// `cargo rustc` options
     #[command(flatten)]
     pub cargo_options: CargoOptions,
+    /// Use `uv` as develop backend instead of `pip`
+    #[arg(long)]
+    pub uv: bool,
 }
 
-fn make_pip_command(python_path: &Path, pip_path: Option<&Path>) -> Command {
-    match pip_path {
-        Some(pip_path) => {
-            let mut cmd = Command::new(pip_path);
-            cmd.arg("--python")
-                .arg(python_path)
-                .arg("--disable-pip-version-check");
-            cmd
-        }
-        None => {
+fn make_backend_command(python_path: &Path, pip_path: Option<&Path>, backend: &Backend) -> Command {
+    match backend {
+        Backend::Pip => match pip_path {
+            Some(pip_path) => {
+                let mut cmd = Command::new(pip_path);
+                cmd.arg("--python")
+                    .arg(python_path)
+                    .arg("--disable-pip-version-check");
+                cmd
+            }
+            None => {
+                let mut cmd = Command::new(python_path);
+                cmd.arg("-m").arg("pip").arg("--disable-pip-version-check");
+                cmd
+            }
+        },
+        Backend::Uv => {
             let mut cmd = Command::new(python_path);
-            cmd.arg("-m").arg("pip").arg("--disable-pip-version-check");
+            cmd.arg("-m").arg("uv").arg("pip");
             cmd
         }
     }
@@ -81,6 +95,7 @@ fn install_dependencies(
     extras: &[String],
     interpreter: &PythonInterpreter,
     pip_path: Option<&Path>,
+    backend: &Backend,
 ) -> Result<()> {
     if !build_context.metadata23.requires_dist.is_empty() {
         let mut args = vec!["install".to_string()];
@@ -110,7 +125,7 @@ fn install_dependencies(
             }
             pkg.to_string()
         }));
-        let status = make_pip_command(&interpreter.executable, pip_path)
+        let status = make_backend_command(&interpreter.executable, pip_path, backend)
             .args(&args)
             .status()
             .context("Failed to run pip install")?;
@@ -127,8 +142,9 @@ fn pip_install_wheel(
     venv_dir: &Path,
     pip_path: Option<&Path>,
     wheel_filename: &Path,
+    backend: &Backend,
 ) -> Result<()> {
-    let mut pip_cmd = make_pip_command(python, pip_path);
+    let mut pip_cmd = make_backend_command(python, pip_path, backend);
     let output = pip_cmd
         .args(["install", "--no-deps", "--force-reinstall"])
         .arg(dunce::simplified(wheel_filename))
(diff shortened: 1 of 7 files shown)


src/develop.rs
```diff
@@ -16,6 +16,10 @@ use std::process::Command;
 use tempfile::TempDir;
 use url::Url;
 
+enum Backend {
```
konstin, line 19: nit: This should have a more specific name, like `InstallBackend`

src/develop.rs
```diff
@@ -57,20 +61,30 @@ pub struct DevelopOptions {
     /// `cargo rustc` options
     #[command(flatten)]
     pub cargo_options: CargoOptions,
+    /// Use `uv` as develop backend instead of `pip`
```
konstin, line 64:
```suggestion
    /// Use `uv` to install packages instead of `pip`
```

src/develop.rs
```diff
@@ -57,20 +61,30 @@ pub struct DevelopOptions {
...
-    match pip_path {
-        Some(pip_path) => {
-            let mut cmd = Command::new(pip_path);
-            cmd.arg("--python")
-                .arg(python_path)
-                .arg("--disable-pip-version-check");
-            cmd
-        }
-        None => {
+fn make_backend_command(python_path: &Path, pip_path: Option<&Path>, backend: &Backend) -> Command {
```
konstin, line 69: This should be a method on `Backend`

src/develop.rs
```diff
@@ -16,6 +16,10 @@ use std::process::Command;
 use tempfile::TempDir;
 use url::Url;
 
+enum Backend {
+    Pip,
```
konstin, line 20: Can you make this `Pip { path: Option<PathBuf> }`?

src/develop.rs
```diff
@@ -148,12 +164,15 @@ fn pip_install_wheel(
...
         );
     }
+    // uv pip install sends logs to stderr thus only print this warning for pip
     if !output.stderr.is_empty() {
-        eprintln!(
-            "⚠️ Warning: pip raised a warning running {:?}:\n{}",
-            &pip_cmd.get_args().collect::<Vec<_>>(),
-            String::from_utf8_lossy(&output.stderr).trim(),
-        );
+        if let Backend::Pip = backend {
```
konstin, line 169: You can `#[derive(PartialEq, Eq)]` on `Backend` and use the normal `==`

konstin (commented):
Thanks for starting this!

konstin:
> Basically about making the develop_hello_world_uv test pass. If I activate my venv then test fails with no-such file or directory hello-world. If I dont activate it, then it passes.

Can you check where this hello-world comes from with `which hello-world`? The tests should be independent of the containing venv not counting cffi.

dmatos2012:
So,
```rust
# develop_hello_world_uv
#[test]
fn develop_hello_world_uv() {
    handle_result(develop::test_develop(
        "test-crates/hello-world",
        None,
        "develop-hello-world-uv",
        false,
        true,
    ));
}

```
```zsh
> virtualenv venv
created virtual environment CPython3.8.10.final.0-64 in 102ms
  creator CPython3Posix(dest=/home/david/oss/maturin/venv, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/home/david/.local/share/virtualenv)
    added seed packages: pip==24.0, setuptools==69.1.1, wheel==0.42.0
```
```zsh
> cargo test develop_hello_world_uv (no venv activated)
running 1 test
test develop_hello_world_uv ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 56 filtered out; finished in 3.48s
```
```zsh
> source venv/bin/activate
> pip install cffi

cargo test develop_hello_world_uv
Errors with: 

---- develop_hello_world_uv stdout ----
🔗 Found bin bindings
🎯 Found 2 Cargo targets in `Cargo.toml`: hello-world, foo
📦 Built wheel to /tmp/.tmpgHeVhr/hello_world-0.1.0-py3-none-linux_x86_64.whl
✏️  Setting installed package as editable
🛠 Installed hello-world-0.1.0
Cause: Check install fail: exit status: 1 
--- Stdout:

--- Stderr:
  File "test-crates/hello-world/check_installed/check_installed.py", line 16, in <module>
    main()
  File "test-crates/hello-world/check_installed/check_installed.py", line 5, in main
    output = check_output(["hello-world"]).decode("utf-8").strip()
  File "/usr/lib/python3.8/subprocess.py", line 415, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.8/subprocess.py", line 493, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/lib/python3.8/subprocess.py", line 858, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.8/subprocess.py", line 1704, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'hello-world'

thread 'develop_hello_world_uv' panicked at tests/common/mod.rs:88:13:
Check install fail: exit status: 1 
--- Stdout:

--- Stderr:
  File "test-crates/hello-world/check_installed/check_installed.py", line 16, in <module>
    main()
  File "test-crates/hello-world/check_installed/check_installed.py", line 5, in main
    output = check_output(["hello-world"]).decode("utf-8").strip()
  File "/usr/lib/python3.8/subprocess.py", line 415, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.8/subprocess.py", line 493, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/lib/python3.8/subprocess.py", line 858, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.8/subprocess.py", line 1704, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'hello-world'

note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    develop_hello_world_uv

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 56 filtered out; finished in 7.91s

error: test failed, to rerun pass `--test run`
```

Then, with the `venv` activated and then after the test failure I do:
```zsh
> pip list
Package     Version
----------- -------
cffi        1.16.0
hello-world 0.1.0
pip         24.0
pycparser   2.22
setuptools  69.1.1
wheel       0.42.0
```
```zsh
> hello-world  
Hello, world!

> which hello-world
/home/david/oss/maturin/venv/bin/hello-world
```

But if I change `test_develop_uv` to just use pip , by setting `false` in last arg and recreate then the whole `venv`
```rust
#[test]
fn develop_hello_world_uv() {
    handle_result(develop::test_develop(
        "test-crates/hello-world",
        None,
        "develop-hello-world-uv",
        false,
        false,
    ));
}

```
then 
```zsh
> pip list
Package    Version
---------- -------
cffi       1.16.0
pip        24.0
pycparser  2.22
setuptools 69.1.1
wheel      0.42.0

```
Doesnt show `hello-world` installed
So it looks like when using `uv`, the `hello-world` is installed to the `venv` but when just using `pip` is not?

dmatos2012:
I have also updated the code to reflect the changes @konstin . Thanks for your review! :)

src/develop.rs
```diff
@@ -16,6 +16,42 @@ use std::process::Command;
...
 use url::Url;
 
+#[derive(PartialEq, Eq, Clone)]
+struct Pip {
+    path: Option<PathBuf>,
+}
+
+#[derive(PartialEq, Eq)]
+enum InstallBackend {
+    Pip(Pip),
```
messense, line 26: I think this can be simplified to just `Pip { path: Option<PathBuf> }`?

tests/common/develop.rs
```diff
@@ -12,6 +12,7 @@ pub fn test_develop(
     bindings: Option<String>,
     unique_name: &str,
     conda: bool,
+    uv: bool,
```
messense, line 15: Might be clearer to also use a enum instead, like

```rust
enum TestInstallBackend {
    Pip,
    Uv,
}
```

messense:
> If I activate my `venv` then test fails with `no-such file or directory hello-world`. If I dont activate it, then it passes. Sadly, if I `deactivate` then the other tests dont pass as I need `cffi` to have my tests pass, what gives?

This might be an existing issue, I do remember encountering similar issue in that past. Usually I just install `cffi` globally to fix it.

dmatos2012:
Yeah, addressed your comments @messense. Yeah, and I guess just installing `cffi` globally fixed all my issues and run the tests without a `venv`.  

Finally, before I add the `TestInstallBackend::Uv` for all tests, would you folks be ok with adding `rstest` as a `dev-dependency` crate? That way we can add both `backend` cases without duplicating the test. Let me know, and then ill change/add accordingly to cover all the tests that `TestInstallBackend::Pip` covers

messense:
Sounds reasonable to me.

konstin:
rstest is a great addition!

For the test suite performance, i'd keep only running the develop test with uv, maybe we can eventually even switch the other tests to uv.

messense (approved):
LGTM, we should probably also support `uv` other installations not from pypi, but it doesn't need to block merging this.

messense:
I'm a bit confused why CI passes without `uv` installed.

messense:
I missed that, thanks for pointing it out!

PR #2024, opened 2024-04-05
Import hook upgrade

mbway:
This PR updates the documentation to point to the new import hook and adds a deprecation warning to the old one.


messense (approved):
Thanks!

PR #2030, opened 2024-04-09
Add support for AIX

bzEq:
Add support for `powerpc64-ibm-aix` which is a tier-3 target.


messense:
Can you also add an AIX python3 sysconfig dump file in https://github.com/PyO3/maturin/tree/main/sysconfig?

messense (approved):
LGTM, thanks!

PR #2032, opened 2024-04-09
Bump version of cargo-config2 to 0.1.24

nipunn1313:
Picks up the fix to https://github.com/taiki-e/cargo-config2/issues/17 which was leading to spurious recompiles.


messense (approved):
Thanks!

Issue #2036, opened 2024-04-10
Remove rust-cpython support


messense:
### Discussed in https://github.com/PyO3/maturin/discussions/2031

<div type='discussions-op-text'>

<sup>Originally posted by **henryiii** April  9, 2024</sup>
Currently, there's a list of options when starting a new project that includes `rust-cpython`. Upon inspection, it turns out [that `rust-cpython` was declared abandoned](https://github.com/dgrunwald/rust-cpython) 7 months ago (2+ years since last meaningful commit, not counting this notice) and doesn't support Python 3.12+. There's a note in the readme that projects should move to PyO3.

Shouldn't this be removed from the new project options, to simplify the user choice? Maybe the auto detection code could be removed too, but at least taking it out of the templates seems reasonable at this point. I have actually had someone ask me whether they should pick pyo3 or rust-python a while back because they saw it in this list.</div>

### Phase out plan

- [x] Remove from new/init templates: #2034
- [x] Remove from readme/documentation: #2057
- [x] Remove build support: #2044

PR #2044, opened 2024-04-16
Remove support for rust-cpython

Owen-CH-Leung:
As per #2036 , this PR removes support for `rust-cpython`. Comments & tests related to rust-cpython are all removed in this PR

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -42,7 +42,6 @@ body:
       label: What bindings you're using
       options:
         - "pyo3"
-        - "rust-cpython"
         - "cffi"
         - "uniffi"
         - "bin"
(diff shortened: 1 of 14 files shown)


messense:
Thanks, it might be better to split this into two PRs, one to remove docs/comments, another one to remove build support, because we haven't decided when to drop build support yet.

messense:
BTW, there are also some leftovers in https://github.com/PyO3/maturin/tree/main/guide

Owen-CH-Leung:
> Thanks, it might be better to split this into two PRs, one to remove docs/comments, another one to remove build support, because we haven't decided when to drop build support yet.

Ok - In this PR I'll focus on removing build support first. Will file another PR to remove docs / comments

Owen-CH-Leung:
@messense Can we restart the failing CI ? looks like there's some intermitent network error

PR #2076, opened 2024-05-09
Ban std::fs trough clippy

konstin:
Maturin is already using `fs_err` in most places, which attaches the filename to io errors. This change enforces this through clippy and fixes the remaining `std::fs` usages. This is the same setup we're successfully using in uv.


messense (approved):
Good idea!

PR #2080, opened 2024-05-09
Add a global verbose option

konstin:
Debugging https://github.com/PyO3/maturin/discussions/2079 i realized that we currently require `RUST_LOG` for users to debug issues like missing inclusions. I added a global counting `-v` flag to allow users to get more verbose logging from maturin to help them diagnose their problems.

We currently use `-v` to configure cargo verbosity level. For simplicity, it now configures both at the same time. Fine-grained control is still possible by setting `RUST_LOG`.

The levels:
* Default: Show build information and `cargo build` output.
* `-v`: Use `cargo build -v`.
* `-vv`: Show debug logging and use `cargo build -vv`.
* `-vvv`: Show trace logging.

This can be extended to support a quiet option.

### DIFF
--- a/src/build_options.rs
+++ b/src/build_options.rs
@@ -94,6 +94,7 @@ pub struct CargoOptions {
     pub ignore_rust_version: bool,
 
     /// Use verbose output (-vv very verbose/build.rs output)
+    // Note that this duplicates the global option, but clap seems to be fine with that.
     #[arg(short = 'v', long, action = clap::ArgAction::Count)]
     pub verbose: u8,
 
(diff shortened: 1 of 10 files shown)


src/build_options.rs
```diff
@@ -94,6 +94,7 @@ pub struct CargoOptions {
     pub ignore_rust_version: bool,
 
     /// Use verbose output (-vv very verbose/build.rs output)
+    // Note that this duplicates the global option, but clap seems to be fine with that.
```
konstin, line 97: This feels wrong but unfortunately i don't understand the cargo argument handling enough to solve this properly

PR #2084, opened 2024-05-12
Ignore PyPi in generated CI for GitHub

shpaker:
This PR adds `--no-index` pip parameter for ignoring PyPi, if pip don't find package in `dist` (dir from `--find-links`).


messense:
this will not work if the package has dependencies.

Issue #2041, opened 2024-04-14
Building an x86 extension module against an x86 Python on an ARM macOS machine


hauntsaninja:
### Discussed in https://github.com/PyO3/maturin/discussions/1980

<div type='discussions-op-text'>

<sup>Originally posted by **hauntsaninja** March  5, 2024</sup>
Currently, `python -m pip install -e .` on a maturin project on an ARM macOS machine using an x86 Python will give me something like:
```
ImportError: dlopen(/Users/shantanu/dev/watchfiles/watchfiles/_rust_notify.cpython-311-darwin.so, 0x0002): tried: '/Users/shantanu/dev/watchfiles/watchfiles/_rust_notify.cpython-311-darwin.so' (mach-o file, but is an incompatible architecture (have 'arm64', need 'x86_64')), '/System/Volumes/Preboot/Cryptexes/OS/Users/shantanu/dev/watchfiles/watchfiles/_rust_notify.cpython-311-darwin.so' (no such file), '/Users/shantanu/dev/watchfiles/watchfiles/_rust_notify.cpython-311-darwin.so' (mach-o file, but is an incompatible architecture (have 'arm64', need 'x86_64'))
```
As far as I can tell, `setuptools-rust` correctly determines target, I think by: https://github.com/PyO3/setuptools-rust/blob/d2f3b3a1e336061604f4eb0fce8c46df6b24e9a9/setuptools_rust/build.py#L700
</div>

messense's reply indicates this works in `maturin develop` and is a missing feature in `maturin pep517 build-wheel`.

The workaround for now is `CARGO_BUILD_TARGET=x86_64-apple-darwin`

PR #2086, opened 2024-05-14
Detect target based on interpreter for pep517 build-wheel

kcking:
This PR detects the desired target based on the python interpreter for pep517 commands, similar to the detection already in place for `maturin develop`.

This fixes `maturin pep517 build-wheel` building for the host architecture, even when the current python interpreter is a different architecture, ie. when using an x86_64 python on arm64 macos host.

Related issue: https://github.com/PyO3/maturin/issues/2041

### DIFF
--- a/src/build_options.rs
+++ b/src/build_options.rs
@@ -551,16 +551,7 @@ impl BuildOptions {
             target_triple = None;
         }
 
-        let mut target = Target::from_target_triple(target_triple)?;
-        if !target.user_specified && !universal2 {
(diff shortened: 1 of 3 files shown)


messense:
Thanks for the PR, consider reopen it regardless of the CI issue, it's a known issue/bug of virualenv on Windows that occurs from time to time.

kcking:
Ah gotcha, thanks! Re-created as https://github.com/PyO3/maturin/pull/2088 since I wasn't able to reopen this one.

PR #2094, opened 2024-05-31
Use base executable when possible in PEP 517 build

messense:
Same as https://github.com/PyO3/setuptools-rust/pull/429


konstin (approved):
That's a smart trick

PR #2097, opened 2024-06-04
Release v1.6.0

messense:


mbway:
Would this release be a good time to remove the old import hook? I can post a PR later if you agree

messense:
I think we should remove it after this release (in v1.7.0), so feel free to post a PR anytime.

PR #2102, opened 2024-06-05
Initial iOS support

FeodorFitsner:
This PR adds minimal iOS support to Maturin.

Related issue: #1742

Solution is merely based on https://github.com/PyO3/maturin/issues/1742 by @holzschu

It works! I've been able to build [pydantic-core](https://pypi.org/project/pydantic-core/) package for iOS and simulator using [crossenv](https://pypi.org/project/crossenv/).

The only issue is wheel tag which is always `-cp312-cp312-ios_23_5_0_arm64.whl` for all three triples passed to `CARGO_BUILD_TARGET`:
* `aarch64-apple-ios`
* `aarch64-apple-ios-sim`
* `x86_64-apple-ios`

though binaries inside `.whl` have correct architecture.

`_PYTHON_HOST_PLATFORM` is also set, but is not respected.

I'd like to improve that in this PR as well (if that's possible), but don't know where to dig. Some hint/guidance would be appreciated.

### DIFF
--- a/src/target.rs
+++ b/src/target.rs
@@ -22,6 +22,7 @@ pub enum Os {
     Linux,
     Windows,
     Macos,
+    Ios,
     FreeBsd,
     NetBsd,
     OpenBsd,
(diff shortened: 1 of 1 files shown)


messense (approved):
Thanks, I think we can merge this now and improve it in future PRs.

messense:
> The only issue is wheel tag which is always `-cp312-cp312-ios_23_5_0_arm64.whl`

What are the correct wheel tags?

PR #2105, opened 2024-06-06
Remove old import hook

mbway:
This removes the old import hook that is superseded by https://github.com/PyO3/maturin-import-hook

I don't think the old hook is referenced anywhere else so simply removing the module is enough.


messense (approved):

messense:
Thanks!

PR #2134, opened 2024-07-02
pep517: only use base python when `MATURIN_PEP517_USE_BASE_PYTHON` env var is set

### LINKED ISSUE #2133: use of base python in pep517 builds breaks eg pydantic-core
### Bug Description

#2094 arranged that maturin passes around the base python interpreter, rather than the python interpreter associated with the isolated build environment.

but then build-requirements declared in projects are not available to that interpreter

this causes the failure reported in https://github.com/pydantic/pydantic-core/issues/1352, per https://github.com/pydantic/pydantic-core/issues/1352#issuecomment-2201121666

1.6.0

3.12

n/a

pyo3

### Does `cargo build` work?

- [ ] Yes, it works

- [ ] Yes

### Steps to Reproduce

- `git clone git@github.com:pydantic/pydantic-core.git`
  - eg `v2.20.0` will work as a tag to reproduce the issue
- create and activate virtual env eg `uv venv -p python3.12 ~/.virtualenvs/foo && workon foo`
- install `build` eg `uv pip install build`
- `python -m build`

messense:
Closes #2133

### DIFF
--- a/guide/src/environment-variables.md
+++ b/guide/src/environment-variables.md
@@ -14,6 +14,7 @@ See [environment variables Cargo reads](https://doc.rust-lang.org/cargo/referenc
 * `_PYTHON_SYSCONFIGDATA_NAME`: Name of a `sysconfigdata*.py` file
 * `MATURIN_PYPI_TOKEN`: PyPI token for uploading wheels
 * `MATURIN_PASSWORD`: PyPI password for uploading wheels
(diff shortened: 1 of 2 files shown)


alex:
Thanks for this fix (it also impacts cryptography). Anything I can help with to get a release with this?

messense:
> Thanks for this fix (it also impacts cryptography). Anything I can help with to get a release with this?

#2139

PR #2158, opened 2024-07-23
Don't check .gitignore files in parent directories

jameshilliard:
When building inside a .gitignore'd output directory as part of another build system maturin may accidentially pick up the parent .gitignore. Lets just disable searching for parent .gitignore files since this functionality is broken anyways as the logic does not match git's .gitignore logic anyways.

For example in buildroot we have an output directory that has its contents gitignored in which maturin will build packages, changing some minor implementation details in how we ignore this directory triggered this bug.

https://github.com/buildroot/buildroot/commit/a14c862c08865e053d5fce90a0d0323b8f9e4bc0

As the output dir files are supposed to be ignored both before and after this change this indicates that ignore::WalkBuilder parent matching is fundamentally broken and can not be used reliably in the first place as it does not match git's ignore logic.

Also disable git exclude and git global excludes which may cause problems.


messense:
> Also disable git exclude and git global excludes which may cause problems.

I'm not sure about disabling git global excludes, I often put things in it (like `.vscode`/`.idea` etc.) to avoid writing in every repo myself.

jameshilliard:
> I'm not sure about disabling git global excludes, I often put things in it (like `.vscode`/`.idea` etc.) to avoid writing in every repo myself.

Hmm, why would those be getting picked up in the build though? They shouldn't be part of the package paths right?

jameshilliard:
By the way any idea why tests are failing on windows? I don't really have a windows system set up for testing right now.

messense:
> By the way any idea why tests are failing on windows? I don't really have a windows system set up for testing right now.

Probably because the disk is full.

messense:
> > By the way any idea why tests are failing on windows? I don't really have a windows system set up for testing right now.
> 
> Probably because the disk is full.

no, it's because this change breaks something because it's not ignoring a `.pyd` file.

```
Cause: File pyo3_mixed_py_subdir\_pyo3_mixed.cp312-win_amd64.pyd was already added from D:\a\maturin\maturin\test-crates\pyo3-mixed-py-subdir\python\pyo3_mixed_py_subdir\_pyo3_mixed.cp312-win_amd64.pyd, can't added it from D:\a\maturin\maturin\test-crates/targets/integration-pyo3-mixed-py-subdir\debug\maturin\pyo3_mixed_py_subdir.dll
Cause: Failed to write to pyo3_mixed_py_subdir\_pyo3_mixed.cp312-win_amd64.pyd
Cause: Failed to add the files to the wheel
```

jameshilliard:
What's the correct way to get the directory containing `pyproject.toml` from `write_python_part`?

jameshilliard:
> no, it's because this change breaks something because it's not ignoring a `.pyd` file.

I think I have this fixed(along with some other random issues such as windows path length issues).

jameshilliard:
@messense Does this look ok to merge? The only test failure I'm seeing appears to be spurious.

messense (approved):

PR #2164, opened 2024-07-26
Fix rust 1.80 clippy errors

jameshilliard:
Was hitting a bunch of weird errors that had no relation to the code changes in #2158, looks like they were related to a recent rust 1.80 release.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -106,6 +106,10 @@ jobs:
       SCCACHE_GHA_ENABLED: "true"
       RUSTC_WRAPPER: "sccache"
     steps:
+      - name: Enable long paths
+        if: startsWith(matrix.platform.os, 'windows')
+        run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -106,6 +106,10 @@ jobs:
       SCCACHE_GHA_ENABLED: "true"
       RUSTC_WRAPPER: "sccache"
     steps:
+      - name: Enable long paths
+        if: startsWith(matrix.platform.os, 'windows')
+        run: |
+          reg add HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem /v LongPathsEnabled /t REG_DWORD /d 1 /f
```
jameshilliard, line 112: Apparently some of the tests use paths that are above the default windows path limit.

messense (approved):

PR #2173, opened 2024-08-04
Use modern stripping option

konstin:
Use `-C strip=symbols` instead of `-C link-arg=-s`. Still broken due to https://github.com/rust-lang/cargo/issues/14346, but i think it's the right modernization anyway.

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -275,9 +275,10 @@ fn cargo_build_command(
     }
 
     if context.strip {
+        // https://doc.rust-lang.org/rustc/codegen-options/index.html#strip
         cargo_rustc
             .args
-            .extend(["-C".to_string(), "link-arg=-s".to_string()]);
+            .extend(["-C".to_string(), "strip=symbols".to_string()]);
     }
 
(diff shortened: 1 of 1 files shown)


messense (approved):
LGTM, note that this only supports Rust 1.45+ which should be acceptable.

edmorley:
> Still broken due to [rust-lang/cargo#14346](https://github.com/rust-lang/cargo/issues/14346), but i think it's the right modernization anyway.

Another option which would both work around the cargo regression and perhaps make things more robust long term would be to switch to setting `CARGO_PROFILE_RELEASE_STRIP=true` instead of passing manual linker args. 

I think passing manual linker args is always going to risk conflicts with Cargo implementation details, whereas using the Cargo provided options (via env vars) will mean Maturin benefits from whatever improvements Cargo makes to features over time (for example, perhaps in the future strip=true might also enable other optimisations, that the manual linker options wouldn't) - plus means Maturin works out of the box more like how end users might invoke Cargo outside of Maturin.

Issue #2174, opened 2024-08-05
Move maturin project metadata from `setup.py` to `pyproject.toml`


messense:
https://github.com/PyO3/maturin/blob/ed8ba1e2f3dccabd969c2500ac4330596d96641a/setup.py#L53-L75

Use [PEP 621](https://peps.python.org/pep-0621/) instead for maturin itself.

PR #2175, opened 2024-08-05
Move project metadata from setup.py to pyproject.toml

Owen-CH-Leung:
As per #2174 , this PR moves the metadata defined under `setup.py` to `pyproject.toml`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,9 +38,6 @@ try:
 except ImportError:
     bdist_wheel = None
 
-with open("README.md", encoding="utf-8", errors="ignore") as fp:
-    long_description = fp.read()
-
 with open("Cargo.toml", "rb") as fp:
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -51,25 +51,10 @@ def finalize_options(self):
...
 
 setup(
-    name="maturin",
-    author="konstin",
-    author_email="konstin@mailbox.org",
-    url="https://github.com/pyo3/maturin",
-    description="Build and publish crates with pyo3, cffi and uniffi bindings as well as rust binaries as "
-    "python packages",
     long_description=long_description,
     long_description_content_type="text/markdown",
```
messense, line 55: long description should also be removed now we use `readme` in `pyproject.toml`.

messense (approved):
LGTM, thanks!

Issue #2120, opened 2024-06-25
Install rust toolchain on building source distribution


padix-key:
Currently, `maturin` requires the rust toolchain (`cargo`, `rustc`) already to be installed. This is no problem in a development scenario as Rust developers presumably have Rust installed.

However, when a user installs the source distribution of a package built with `maturin`, the user also requires the Rust toolchain, although they may be only interested in the package, but not in Rust. Rust is also not a part (yet) of the major OS distributions, so that user would need to go the extra mile of running `rustup` first. 

Therefore, I propose installing the Rust toolchain in an isolated environment (e.g. similar to https://github.com/chriskuehl/rustenv) when `maturin` is used as `[build-system]`, at least in case the rust toolchain is not installed, yet.

messense:
I don't think this should be done in `maturin`, when using source distribution, users should be aware of the build system requirement.

Even though gcc/clang is part of the major OS distributions, when using sdist, it's still possible that they are missing from the OS environment of the end user, in such case should `setuptools` or other build backends install gcc/clang for the users? Absolutely not, IMO.

konstin:
Personally, i think a source distribution should install all build dependencies it needs; it should not be the responsibility of a python user to understand the native build system of their transitive dependencies only because they are let's say on an arm platform without wheels. We'd need a way to bootstrap rust and cargo for this though that doesn't change the users environment (like rustup does). It would be ideal to have wheels for cargo and rustc on pypi like we have with zig. Relying on (linux) distros is often insufficient, they require root access to install new packages that is not available in many environments and only ship specific versions of tools. It would honestly be amazing if we could get gcc and clang from pypi too.

messense:
It sounds amazing but surely there will be people start complain that using sdist download lots of huge dependencies.

IMO if it's really desirable, perhaps it should just be another dependency in `[build-system]` table that gets installed by a build frontend, instead of installing by `maturin`?

konstin:
We can check for rust and cargo in `get_requires_for_build_wheel` and only if it is not installed add a dependency on the pypi package(s). For me, maturin should handle as much as possible end-to-end without the developer specifying something additionally.

messense:
> We can check for rust and cargo in `get_requires_for_build_wheel` and only if it is not installed add a dependency on the pypi package(s).

I'm fine with this solution if it works, but consider when a sdist requires a specific 
[...]

PR #2177, opened 2024-08-08
Implement new sdist feature to download rust toolchain

Owen-CH-Leung:
Solves #2120

This PR implements feature to download rust toolchain when trying to pip install a python package built by maturin. Specifically, when maturin execute `WriteDistInfo`, when compiling with the `rustls` feature, it will try to check if toolchain is installed, and proceed to install the rustup installer and download the stable toolchain if toolchain is absent


Owen-CH-Leung:
@messense Can you review this PR ? I'm not sure if I'm heading to the right direction so I'd like to get some feedback before proceeding to polish it

src/build_context.rs
```diff
@@ -1132,6 +1148,19 @@ impl BuildContext {
         }
         Ok(wheels)
     }
+    /// Check if user requires to install rust toolchain
+    ///
+    /// Loop over `build-system.requires` defined in pyproject.toml and see if `rust-toolchain` is provided
```
messense, line 1153: This isn't going to work because `pip` will try to install `rust-toolchain` package from `pypi`.

src/upload.rs
```diff
@@ -382,7 +382,7 @@ fn http_agent() -> Result<ureq::Agent, UploadError> {
 
 #[cfg(feature = "rustls")]
 #[allow(clippy::result_large_err)]
-fn http_agent() -> Result<ureq::Agent, UploadError> {
+pub fn http_agent() -> Result<ureq::Agent, UploadError> {
```
messense, line 385: there is a `http_agent` below that also need to be made public.

src/build_context.rs
```diff
@@ -273,6 +273,22 @@ impl BuildContext {
...
+                    target_path.push("bin");
+
+                    fs::create_dir_all(&target_path)
+                        .context("Fail to create directory for installing rust toolchain")?;
+
+                    let target_path_str = target_path
+                        .to_str()
+                        .context("Fail to construct target path for installing rust toolchain")?;
+
+                    download_and_execute_rustup(target_path_str, target_path_str)
```
messense, line 289: Skip download of there is already a working rust toolchain (probably need to validate version with `rust-version` in `Cargo.toml` and `Cargo.toml` of every crate dependencies, I think we can get that from `cargo metadata` output).

src/build_context.rs
messense: Since we don't have `rust-toolchain` on pypi yet, we can try to use `rust-toolchain.toml` / legacy `rust-toolchain` file or `rust-version` in `Cargo.toml` of the binding crate, then fallback to Rust `stable`.

messense:
cc @padix-key since you requested this feature.

src/build_context.rs
Owen-CH-Leung: So it means like if user provides a `rust-toolchain.toml` file in the project directory, that would be indicating that user wants to install rust-toolchain ?

padix-key: To put my two cents in, I think that the rust toolchain should be present for the build, even if `rust-toolchain.toml`is missing in which case the default toolchain would be used. Bonus points if `[toolchain]` can be specified directly in the `pyproject.toml` as e.g `[tool.maturin.toolchain]`.

Owen-CH-Leung: how about an additional command line flag such as `maturin sdist --install-rust-toolchain` ? (Or maybe when env variable `MATURIN_INSTALL_RUST` is set to true? ). This will be backward-compatible while keeping maturin flexible for users .

padix-key: This would only be suitable in the case the user of the package knows what Maturin, Rust etc. is. However, the reason I brought up issue #2120, is that a typical user just wants to use a package and would probably do not like to think about the build backend of this package.

In other words, it would be nice in my opinion, if a user can install a source distribution of a package that uses `maturin` as `build-system` simply via `pip install` without knowing such internals

Owen-CH-Leung: I found that currently maturin did check for the presence of rust toolchain when it attempts to build a project : 

https://github.com/PyO3/maturin/blob/main/maturin/__init__.py#L178-L182

So as an attempt to replicate this error msg, I tried to install a pypi package in an isolated environment without rust: 

`pip install --no-binary orjson orjson`

And I did see this error msg displayed in stdout. So maybe instead of exiting, we can attempt to install the toolchain from there ?

Owen-CH-Leung: @messense @padix-key I've modified `prepare_metadata_for_build_wheel` to remove the check for rust-toolchain, and add logic in `maturin pep517 write-dist-info` command to check for toolchain from there ( and install from rustup if not present). Let me know what you think

padix-key: I have actually barely insight into the `maturin` internals, I only know the user side 😃.

konstin:
(sorry for the further delay, i was on this but got thrown off again by https://github.com/PyO3/maturin/pull/2268)

src/build_context.rs
mbway: I think automatically downloading an installation script and running it for the user and installing to non-standard locations is undesirable. I would not be happy if this took place on my system without any confirmation at least. I could also see it causing problems for CI systems which may accidentally install rustup each build if misconfigured.
An opt-in command line flag seems like a good approach to me. If not passed it could print
```
rust toolchain (cargo) not found.
Please install it manually or pass `maturin sdist --auto-install-toolchain`
```
so discoverability wouldn't be an issue for the 'typical user' described as not being aware of what they need to have installed

mbway:
Why are rustup and cargo being installed to non-standard locations? the code seems to be using `home_dir()` as `CARGO_HOME`. On unix the default is `$HOME/.cargo` ([source](https://doc.rust-lang.org/cargo/reference/environment-variables.html)).

padix-key:
Would it be sensible to install `cargo` and `rustup` in a temporary location and clean it after finishing the build, or do you think the size of them would be too large for repeated install/uninstall?

konstin:
I went with a different implementation in https://github.com/PyO3/maturin/pull/2421, installing rust in python means we can also bootstrap maturin itself.

PR #2206, opened 2024-09-02
Normalize `REPOSITORY_URL`

austin362667:
Normalize URLs before comparison to ensure consistent matching, regardless of trailing slashes.

This approach standardizes the URLs by either consistently adding or removing the trailing slash, allowing for flexible and accurate comparisons irrespective of user input.

For example, 

```
const TEST_REPOSITORY_URL: &'static str = "https://test.pypi.org/legacy/";
```
should be the same as
```
const TEST_REPOSITORY_URL: &'static str = "https://test.pypi.org/legacy";
```
no matter there's a trailing slash at the end of url or not.

### DIFF
--- a/src/upload.rs
+++ b/src/upload.rs
@@ -322,8 +322,8 @@ fn complete_registry(opt: &PublishOpt) -> Result<Registry> {
     {
         let normalized_url = PublishOpt::normalize_url(repository_url);
         let name = match normalized_url {
-            normalized if normalized == PublishOpt::DEFAULT_REPOSITORY_URL => Some("pypi"),
-            normalized if normalized == PublishOpt::TEST_REPOSITORY_URL => Some("testpypi"),
+            PublishOpt::DEFAULT_REPOSITORY_URL => Some("pypi"),
+            PublishOpt::TEST_REPOSITORY_URL => Some("testpypi"),
             _ => None,
         };
         (name, repository_url.to_string())
--- a/src/upload.rs
+++ b/src/upload.rs
@@ -59,8 +59,8 @@ pub struct PublishOpt {
 }
 
 impl PublishOpt {
-    const DEFAULT_REPOSITORY_URL: &'static str = "https://upload.pypi.org/legacy/";
-    const TEST_REPOSITORY_URL: &'static str = "https://test.pypi.org/legacy/";
+    const DEFAULT_REPOSITORY_URL: &'static str = "https://upload.pypi.org/legacy";
+    const TEST_REPOSITORY_URL: &'static str = "https://test.pypi.org/legacy";
 
     /// Set to non interactive mode if we're running on CI
     pub fn non_interactive_on_ci(&mut self) {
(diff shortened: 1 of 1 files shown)


src/upload.rs
```diff
@@ -316,9 +320,18 @@ fn complete_registry(opt: &PublishOpt) -> Result<Registry> {
     let pypirc = load_pypirc();
     let (registry_name, registry_url) = if let Some(repository_url) = opt.repository_url.as_deref()
     {
-        let name = match repository_url {
-            PublishOpt::DEFAULT_REPOSITORY_URL => Some("pypi"),
-            PublishOpt::TEST_REPOSITORY_URL => Some("testpypi"),
+        let normalized_url = PublishOpt::normalize_url(repository_url);
+        let name = match normalized_url {
+            normalized
+                if normalized == PublishOpt::normalize_url(PublishOpt::DEFAULT_REPOSITORY_URL) =>
```
messense, line 326: I'd just remove the end slash in constant value instead.

austin362667: Hi @messense  Thanks for reviewing. That's a solution, too.
Then, what if user input this: `"https://test.pypi.org/legacy/"`?

austin362667: Aha Understood! I got your point, thanks for simplify the logic. Updated the code.

src/upload.rs
```diff
@@ -316,9 +320,10 @@ fn complete_registry(opt: &PublishOpt) -> Result<Registry> {
     let pypirc = load_pypirc();
     let (registry_name, registry_url) = if let Some(repository_url) = opt.repository_url.as_deref()
     {
-        let name = match repository_url {
-            PublishOpt::DEFAULT_REPOSITORY_URL => Some("pypi"),
-            PublishOpt::TEST_REPOSITORY_URL => Some("testpypi"),
+        let normalized_url = PublishOpt::normalize_url(repository_url);
+        let name = match normalized_url {
+            normalized if normalized == PublishOpt::DEFAULT_REPOSITORY_URL => Some("pypi"),
+            normalized if normalized == PublishOpt::TEST_REPOSITORY_URL => Some("testpypi"),
```
messense, line 326: Now the `if`s can be rewriten back to simple `match` arms I think.

src/upload.rs
```diff
@@ -316,9 +320,10 @@ fn complete_registry(opt: &PublishOpt) -> Result<Registry> {
     let pypirc = load_pypirc();
     let (registry_name, registry_url) = if let Some(repository_url) = opt.repository_url.as_deref()
     {
-        let name = match repository_url {
-            PublishOpt::DEFAULT_REPOSITORY_URL => Some("pypi"),
-            PublishOpt::TEST_REPOSITORY_URL => Some("testpypi"),
+        let normalized_url = PublishOpt::normalize_url(repository_url);
```
messense, line 323: Now that there is only one call to `normalize_url` might as well just call `trim_end_matches` on `registry_url` directly.

src/upload.rs
austin362667: Certainly, let's simplify this way.

messense (approved):
Thanks!

PR #2219, opened 2024-09-11
Replace "." with "/" in module_name

### LINKED ISSUE #2218: cannot open shared object file: No such file or directory
### Bug Description

Unable to find shared object when importing libary on Ubuntu.

1.7.1

3.10.12

23.2.1

cffi

### Does `cargo build` work?

- [X] Yes, it works

- [ ] Yes

### Steps to Reproduce

Getting this error when building https://github.com/bempp/ndelement, eg in the yanked pypi releases (https://pypi.org/project/ndelement/0.1.0.dev2/). Can get error locally by running:

```bash
maturin build
pip install target/wheels/ndelement-....whl
python -c "import ndelemement"
```

mscroggs:
If a "." is used in module_name, this should be replaced by joining a path, as it represents a subdirectory.

Fixes #2218.

### DIFF
--- a/src/build_context.rs
+++ b/src/build_context.rs
@@ -480,12 +480,12 @@ impl BuildContext {
         }
 
         // Currently artifact .so file always resides at ${module_name}/${module_name}.so
-        let artifact_dir = Path::new(&self.module_name);
+        let artifact_dir = self.module_name.split(".").collect::<PathBuf>();
         for artifact in artifacts {
             let mut new_rpaths = patchelf::get_rpath(&artifact.path)?;
             // TODO: clean existing rpath entries if it's not pointed to a location within the wheel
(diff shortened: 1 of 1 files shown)


messense:
Just to confirm, do it work for module name that doesn't not contain `.`?

mscroggs:
> Just to confirm, do it work for module name that doesn't not contain `.`?

Good question. I'll make up a local package now and check

mscroggs:
> Just to confirm, do it work for module name that doesn't not contain `.`?

Picking from the examples in the README, I was able to build and install https://github.com/webonnx/wonnx with this branch

mscroggs:
Good news: I locally built [ndelement](https://github.com/bempp/ndelement) with module-name updated to not have a dot, and it built, installed an imported without errors

PR #2220, opened 2024-09-14
feat: copy `.pdb` files to `Editable Installation` and `Wheel` for easier debugging on windows

WSH032:
## What I Did

I added a `--with-debuginfo` option to both `maturin dev` and `maturin build`. Currently, it only works on `msvc`. When this option is specified:

- `maturin dev` will copy the debug file (e.g., `mylib.pdb`) with the same name as the `artifact` (e.g., `mylib.dll`) to the `editable install` directory.
- `maturin build` will include the debug file in the `wheel`.
- For `bin` bindings, `*.pdb` will be distributed along with `*.exe` in the `Scripts` directory.

Even if developers do not generate debug information, there will be no impact as long as they don't manually specify `--with-debuginfo`.

Specifically, with the following configuration:

```toml
# Cargo.toml

[lib]
name = "lib_name"
```

```toml
# pyproject.toml

python-source = "my_project"
module-name = "my_project._lib_name"
```

The resulting directory structure would look like this:

```tree
my-project
├── pyproject.toml
├── Cargo.toml
├── my_project
│   ├── __init__.py
│   ├── bar.py
│   ├── lib_name.pdb
│   └── _lib_name.cp310-win_amd64.pyd
├── README.md
└── src
    └── lib.rs
```

Please note that the debug file is `lib_name.pdb` instead of `_lib_name.pdb`, due to the linker flag `/PDBALTPATH:%_PDB%`.

---

### `Linux`

On Linux, the default `split-debuginfo` is `off`, so the debug information is included in the `.so` file. Therefore, no additional actions are needed by default.

### `macOS`

Currently, the default value for `split-debuginfo` on macOS is `packed`, but it seems recommended to set it to `unpacked`, see <https://internals.rust-lang.org/t/help-test-faster-incremental-debug-macos-builds-on-nightly/14016>.

I tried to reference [python-build-standalone](https://github.com/indygreg/python-build-standalone)'s approach. I downloaded these three distribution packages:

- [aarch64-apple-darwin-install_only](https://github.com/indygreg/python-build-standalone/releases/download/20240909/cpython-3.10.15+20240909-aarch64-apple-darwin-install_only.tar.gz)
- [x86_64-pc-windows-msvc-install_only](https://github.com/indygreg/python-build-standalone/releases/download/20240909/cpython-3.10.15+20240909-x86_64-pc-windows-msvc-install_only.tar.gz)
- [x86_64-unknown-linux-gnu-install_only](https://github.com/indygreg/python-build-standalone/releases/download/20240909/cpython-3.10.15+20240909-x86_64-unknown-linux-gnu-install_only.tar.gz)

I found that only the `x86_64-pc-windows-msvc-install_only` package contained `.pdb` debug information, while the other two platforms did not have any separate debug information. So, I am unsure how to implement this feature for those platforms.

Most importantly, I do not have experience developing on macOS, nor do I have a macOS machine to perform the necessary testing. Therefore, this will need to be left for others to implement.

---

Next, I will work on writing tests and documentation, but I would like to get your feedback on the current implementation first.

### DIFF
--- a/src/build_context.rs
+++ b/src/build_context.rs
@@ -3,8 +3,8 @@ use crate::auditwheel::{PlatformTag, Policy};
 use crate::build_options::CargoOptions;
 use crate::compile::{warn_missing_py_init, CompileTarget};
 use crate::module_writer::{
-    add_data, write_bin, write_bindings_module, write_cffi_module, write_python_part,
-    write_uniffi_module, write_wasm_launcher, WheelWriter,
+    add_data, include_artifact_for_editable_install, write_bin, write_bindings_module,
+    write_cffi_module, write_python_part, write_uniffi_module, write_wasm_launcher, WheelWriter,
 };
 use crate::project_layout::ProjectLayout;
 use crate::python_interpreter::InterpreterKind;
@@ -187,6 +187,11 @@ pub struct BuildContext {
     pub universal2: bool,
     /// Build editable wheels
     pub editable: bool,
+    /// Whether to include debug information(`.pdb` on msvc)
+    /// in the wheels or editable-installs.
+    /// Currently only support `.pdb` files with
+    /// the same name as the binary(`*.exe`/`*.dll`) on `msvc` platform
+    pub with_debuginfo: bool,
     /// Cargo build options
     pub cargo_options: CargoOptions,
 }
@@ -692,6 +697,7 @@ impl BuildContext {
             &self.target,
             self.editable,
             self.pyproject_toml.as_ref(),
+            self.with_debuginfo,
         )
         .context("Failed to add the files to the wheel")?;
 
@@ -770,6 +776,7 @@ impl BuildContext {
             &self.target,
             self.editable,
             self.pyproject_toml.as_ref(),
+            self.with_debuginfo,
         )
         .context("Failed to add the files to the wheel")?;
 
@@ -861,7 +868,11 @@ impl BuildContext {
         let maturin_build = artifact_path.parent().unwrap().join("maturin");
         fs::create_dir_all(&maturin_build)?;
         let new_artifact_path = maturin_build.join(artifact_path.file_name().unwrap());
-        fs::copy(artifact_path, &new_artifact_path)?;
+        include_artifact_for_editable_install(
+            artifact_path,
+            &new_artifact_path,
+            self.with_debuginfo,
+        )?;
         artifact.path = new_artifact_path;
         Ok(artifact)
     }
@@ -893,6 +904,7 @@ impl BuildContext {
             &self.interpreter[0].executable,
             self.editable,
             self.pyproject_toml.as_ref(),
+            self.with_debuginfo,
         )?;
 
         self.add_pth(&mut writer)?;
@@ -959,6 +971,7 @@ impl BuildContext {
             self.target.target_os(),
             self.editable,
             self.pyproject_toml.as_ref(),
+            self.with_debuginfo,
         )?;
 
         self.add_pth(&mut writer)?;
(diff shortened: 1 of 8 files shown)


messense (commented):
Thanks for the PR!

For Linux, when `--with-debuginfo` is presented, we can explicitly pass `-C split-debuginfo=off` to rustc, similarly for macOS we pass `-C split-debuginfo=unpacked`.

(And it'd be better to check for existing `split-debuginfo` rustc flag and error out when its value is incompatible with `--with-debuginfo`.)

src/build_context.rs
```diff
@@ -187,6 +188,11 @@ pub struct BuildContext {
     pub universal2: bool,
     /// Build editable wheels
     pub editable: bool,
+    /// Whether to include debug information(`.pdb` on msvc)
+    /// in the wheels or editable-installs.
+    /// Currently only support `.pdb` files with
+    /// the same name as the binary(`*.exe`/`*.dll`) on `msvc` platform
+    pub with_debuginfo: bool,
```
messense, line 195: This can't be used together with `--strip`:

```suggestion
    #[arg(long, conflicts_with = "strip")]
    pub with_debuginfo: bool,
```

src/build_options.rs
```diff
@@ -202,6 +202,12 @@ pub struct BuildOptions {
     #[arg(long)]
     pub zig: bool,
 
+    /// Include debug information in the Wheel.
+    /// Currently only support `.pdb` files with
+    /// the same name as the binary(`exe/dll`) on `msvc` platform
+    #[arg(long)]
```
messense, line 208: This can't be used together with `--strip`. (Not sure if `clap` supports this kind of setup though)

```suggestion
    #[arg(long, conflicts_with = "strip")]
```

WSH032:
Thank you for the review!

---

> For Linux, when `--with-debuginfo` is presented, we can explicitly pass `-C split-debuginfo=off` to rustc, similarly for macOS we pass `-C split-debuginfo=unpacked`.
>
> (And it'd be better to check for existing `split-debuginfo` rustc flag and error out when its value is incompatible with `--with-debuginfo`.)

I'm concerned that there is no simple and robust way to check the value of `split-debuginfo`. Developers can set it not only through the `RUSTFLAGS` environment variable but also via [`.cargo/config.toml`](https://doc.rust-lang.org/cargo/reference/config.html). I'm unsure if Cargo provides an existing method to retrieve this value.

Considering the difficulty of confirming existing configurations and the uncertainty of whether developers are indeed generating the corresponding debug files, I implemented `--with-debuginfo` in a way that requires developers to manually specify it.

---

This [commit](https://github.com/PyO3/maturin/pull/2220/commits/b9a034f9e62829ed76c4b68968c945668d58edae) introduced a [regression](https://cirrus-ci.com/task/4761335471276032) on FreeBSD, which I couldn’t reproduce on Windows. I’m unsure if it’s a sporadic issue. Could we re-run the CI?

src/build_options.rs
WSH032: On `MSVC`, even if `RUSTFLAGS="-Cstrip=symbols"` is specified, `rustc` will still generate a `.pdb` file, so we can't let it conflict. Not sure how it behaves on `macOS`.

messense: But does the `.pdb` file contains useful information?

messense:
> I'm concerned that there is no simple and robust way to check the value of `split-debuginfo`

It's handled by [`cargo-config2`](https://docs.rs/cargo-config2/latest/cargo_config2/) which is already used in `maturin`.

> Specifically, when using `--with-debuginfo`, I prefer to have developers manually specify `RUSTFLAGS="-Csplit-debuginfo=off"` rather than having maturin automatically set it.

I disagree, it should provide a seamless user experience by default w/o asking user to look up how to set the right `RUSTFLAGS`.

kmk142789 (approved):

PR #2246, opened 2024-10-05
Switch macos x86_64 to macos14 in generate-ci

masklinn:
The macos12 images [are deprecated and will be decommissioned by december](https://github.com/actions/runner-images/issues/10721), jobs have started triggering warnings and brownouts are planned.

In my understanding this means the x86_64 wheels will be cross-compiled from the arm64 runners, but I understand cross compilation works fine. The x86_64 cross compilation is slower than the ARM64 compilation, but on my test project it's faster than the macos13 x86_64 build so that doesn't seem like a massive concern I feel?

### DIFF
--- a/src/ci.rs
+++ b/src/ci.rs
@@ -264,18 +264,13 @@ jobs:\n",
                         target,
                     })
                     .collect(),
-                Platform::Macos => {
-                    vec![
-                        MatrixPlatform {
-                            runner: "macos-12",
-                            target: "x86_64",
-                        },
-                        MatrixPlatform {
-                            runner: "macos-14",
-                            target: "aarch64",
-                        },
-                    ]
-                }
+                Platform::Macos => ["x86_64", "aarch64"]
+                    .into_iter()
+                    .map(|target| MatrixPlatform {
+                        runner: "macos-14",
+                        target,
+                    })
+                    .collect(),
                 Platform::Emscripten => vec![MatrixPlatform {
                     runner: "ubuntu-latest",
                     target: "wasm32-unknown-emscripten",
@@ -767,7 +762,7 @@ mod tests {
                 strategy:
                   matrix:
                     platform:
-                      - runner: macos-12
+                      - runner: macos-14
                         target: x86_64
                       - runner: macos-14
                         target: aarch64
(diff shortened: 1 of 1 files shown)


messense (changes requested):
Thanks, note that `pytest` step also needs adjustment because you can't run tests for x86_64 wheels on aarch64 macOS after this change.

masklinn:
> Thanks, note that `pytest` step also needs adjustment because you can't run tests for x86_64 wheels on aarch64 macOS after this change.

Ah good catch, I assumed either arm64 or x86_64 python would be installed and then that's what would be run, but not so, and it's not even a question of enabling `architecture` for macos as from its docs setup-python installs a universal binary.

From my testing it looks like `arch` will correctly select the architecture, so I could set `env: ARCHPREFERENCE` on the `run` step then prefix every relevant command (pip install to ensure the correct binary arch is selected, and running the test suite) with `arch` and it should work correctly, what do you think?

messense:
That may not works with pypy I think, they don't have universal2 binaries: https://pypy.org/download.html

messense:
> The note about universal binaries is in the readme: https://github.com/actions/python-versions/tree/3.13.0-11228081754?tab=readme-ov-file#building-installation-packages

Right, but note that it's only for Python >= 3.11.

masklinn:
> Right, but note that it's only for Python >= 3.11.

Sure but generate-ci creates a template for 3.x so that's been >= 3.11 since it was created.

Though if we pass the architecture to setup-python for the benefit of pypy and graal it *should* get picked up for cpython < 3.11 and retrieve arch-specific payloads, and like pypy and graal the `arch` is basically just an assertion.

I'll try that out to make sure it works.

masklinn:
Well apparently it doesn't, python < 3.11 x64 seems broken on M1: https://github.com/masklinn/ci-test/actions/runs/11296863554/job/31422564472

I'll have to see with the setup-python people if they know why.

masklinn:
Welp, apparently action/setup-python has decided not to fix the non-universal python images, so not quite sure what to do there, I guess I'll close this.?

messense:
Close for now, we have switched to macos-13.

PR #2254, opened 2024-10-15
feat: add skip attestation option to maturin ci github

moldhouse:
for private, non-enterprise repositories, attestation of github artifacts is not available. this commit adds the option to generate a ci file which skips the attestation step


messense (approved):

messense:
Thanks!

PR #2261, opened 2024-10-17
build_context: improve wheel reproducibility by sorting libs

ycongal-smile:
While tracking a reproducibility issue from maturin's output, we found that the .so file in the output .whl were not ordered the same every time.

Order of the external libraries in the .whl comes down to the order `soname_map` is iterated. But, `std::HashMap` does not provide a stable order and that create an unstable order in the wheel.

Switch to `std::BTreeMap` which keeps keys sorted and is iterable in a stable order.

This can be tested by building current python3-cryptography :
```
SOURCE_DATE_EPOCH=1728915855 maturin build
```
Before this commit, the above give two possible outputs (~50% each), the difference is in the order of libssl and libcrypto.

After this commit, the output is reproducible.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Unreleased]
+
+* Improve wheel reproducibility by sorting external libraries [#2261](https://github.com/PyO3/maturin/pull/2261)
+
 ## [1.7.4]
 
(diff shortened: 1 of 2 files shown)


konstin (approved):
Thank you!

PR #2266, opened 2024-10-18
Remove Python 3.8 from tests

konstin:
Python 3.8 is EOL. We need to add 3.13 in the next step


konstin:
Rolling into https://github.com/PyO3/maturin/pull/2268

PR #2283, opened 2024-10-29
Add Interpreter Argument for pep517 sdist Generation

### LINKED ISSUE #2245: Maturin sdist Generation Fails When Default Python Command Doesn't Exist
### Bug Description

Hello,

I have been working on getting Rust builds with maturin in our company's build system. This system isolates builds in an environment already and then runs `sys.executable -m build --sdist --outdir=<staging dir> --no-isolation`. Maturin then fails with ` Found pyo3 bindings
💥 maturin failed
  Caused by: Invalid python interpreter version`.

I traced this issue to https://github.com/PyO3/maturin/blob/main/src/build_options.rs#L1206. Here, maturin calls `python_ver.split_once('.')...`. However, in a no-isolation situation, for sdist generation, maturin gets the python interpreter from `target.get_python`. This will return `python3` for non-venv linux systems. When we try to `split_once` on that, it fails.

I'm happy to put up a fix for this if I can get some guidance on what the proper fix should be. We could first check for just a major version for example and go from there if that would work downstream.

Our build system is quite complicated, but I can reproduce this locally as well and with multiple python versions.

1.7.4

3.12

23.2.1

pyo3

### Does `cargo build` work?

- [X] Yes, it works

- [ ] Yes

### Steps to Reproduce

1. Create a minimal python/rust library with just PyO3 as a dependency.
2. In a linux system, (no venv), install build and maturin.
3. Run `python3.12 -m build --sdist --no-isolation .`
4. Maturin should crash with the same error.

paddonizio:
- Fixes #2245 
- Adds an interpreter argument to maturin's pep517 sdist generation, causing maturin to skip the interpreter resolution step.
- Adds `--interpreter _get_sys_interpreter()` to the __init__.py command invoking the maturin binary to ensure sdist generation uses the interpreter maturin was invoked with.

### DIFF
--- a/guide/src/distribution.md
+++ b/guide/src/distribution.md
@@ -94,7 +94,7 @@ Options:
           This option is ignored on all non-linux platforms
 
   -i, --interpreter [<INTERPRETER>...]
-          The python versions to build wheels for, given as the executables of interpreters such as `python3.9` or `/usr/bin/python3.8`
+          The python versions to build wheels or an sdist for, given as the executables of interpreters such as `python3.9` or `/usr/bin/python3.8`
 
   -f, --find-interpreter
           Find interpreters from the host machine
--- a/maturin/__init__.py
+++ b/maturin/__init__.py
@@ -121,7 +121,7 @@ def build_wheel(
 
 # noinspection PyUnusedLocal
 def build_sdist(sdist_directory: str, config_settings: Optional[Mapping[str, Any]] = None) -> str:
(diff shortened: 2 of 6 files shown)


paddonizio:
I didn't see any tests covering the areas I was changing, but let me know if I missed something and I can add some.

messense:
> but I guess the proper fix would be skipping python interpreter discovery during sdist build, I wonder how hard that will be, will take a look later when I'm free.

Hi @paddonizio can you try #2292, thanks!

messense:
Superseded by #2292.

Thanks again for the PR.

PR #2284, opened 2024-10-30
Update goblin to 0.9

musicinmybrain:
The breaking change from 0.8.0 appears to be the addition of a value to a pub enum associated with TE (terse executable) support; this does not look like it should affect maturin’s usage of goblin since it doesn’t match exhaustively on this enum.

https://github.com/m4b/goblin/blob/d096260201158ed34d64728fd1ab0ca125e2f956/CHANGELOG.md#092----2024-10-26

I’ve also filed PR’s for both of the crates that are still asking for goblin 0.8.0 in `maturin`’s dependency tree:

- https://github.com/messense/fat-macho-rs/pull/20
- https://github.com/messense/lddtree-rs/pull/12

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -724,11 +724,11 @@ checksum = "9fc0510504f03c51ada170672ac806f1f105a88aa97a5281117e1ddc3368e51a"
 
 [[package]]
 name = "fat-macho"
-version = "0.4.8"
+version = "0.4.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


konstin (approved):
Thanks

messense:
Could you please also update related deps now that I've release new versions? Thanks!

musicinmybrain:
> Could you please also update related deps now that I've release new versions? Thanks!

Yes, I missed that you had already released! Pushing now.

PR #2292, opened 2024-11-06
Don't resolve python interpreter when building sdist only

### LINKED ISSUE #2245: Maturin sdist Generation Fails When Default Python Command Doesn't Exist

messense:
Alternative to #2283

Fixes #2245

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -273,7 +273,13 @@ fn pep517(subcommand: Pep517Command) -> Result<()> {
             strip,
         } => {
             assert_eq!(build_options.interpreter.len(), 1);
-            let context = build_options.into_build_context(true, strip, false)?;
+            let context = build_options
+                .into_build_context()
+                .release(true)
+                .strip(strip)
+                .editable(false)
+                .sdist_only(true)
+                .build()?;
 
             // Since afaik all other PEP 517 backends also return linux tagged wheels, we do so too
             let tags = match context.bridge() {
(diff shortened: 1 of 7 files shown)


messense:

- #2292 ◀

src/main.rs
```diff
@@ -273,7 +273,13 @@ fn pep517(subcommand: Pep517Command) -> Result<()> {
             strip,
         } => {
             assert_eq!(build_options.interpreter.len(), 1);
-            let context = build_options.into_build_context(true, strip, false)?;
+            let context = build_options
+                .into_build_context()
+                .release(true)
+                .strip(strip)
+                .editable(false)
+                .sdist_only(true)
```
messense, line 281:
```suggestion

```

PR #2208, opened 2024-09-02
UniFFI: supports bindings generated from multiple crates

MatMaul:
This changes allow to build a functional wheel of [Matrix Rust SDK](https://github.com/matrix-org/matrix-rust-sdk) FFI bindings.

PR #2305, opened 2024-11-16
Fix __init__ exports when using multiple UniFFI bindings

### LINKED ISSUE #2294: Invalid exports in `__init__.py` when UniFFI project doesn't include .udl files
### Bug Description

Hey folks,

This issue I believe is related to unresolved issue I [voiced awhile back](https://github.com/PyO3/maturin/issues/2090) as well as the multi-crate features added in #2208. 

I currently have a Python project I'm hoping can leverage an external crate's UniFFI .udl to export Python bindings for the crate. Right now I just have the external crate added as a dependency and have the following in my `lib.rs`:

```rust
pub use odict::*;

odict::uniffi_reexport_scaffolding!();
```

Running `maturin develop`, the following files get generated:

- `__init__.py`
- `libuniffi_pyodict.dylib`
- `odict.py`

Inside `__init__.py`, it attempts to export from a nonexistent module:

```python
from .pyodict import *  # NOQA
```

Similarly, `odict.py` attempts to read from `libpyodict.dylib`, which doesn't exist unless I manually override the cdylib name to be `pyodict`, but that could be a separate issue. 

I looked at the source code and saw this `__init__.py` [seems to be hardcoded to use the current project's name](https://github.com/PyO3/maturin/blob/ae059a3dca24ceb9b18f4bcdf627187e79fec00c/src/module_writer.rs#L1139). I think this could be fixed by #1910, as it includes a change to __init__.py that isn't present in #2208. Clarity on the cdylib issue would also be appreciated!

1.7.4

3.13.0

24.3.1

uniffi

### Does `cargo build` work?

- [X] Yes, it works

- [ ] Yes

### Steps to Reproduce

Described above

Nickersoft:
This PR is a continuation of the work started in #1910 and #2208 that updates the generated `__init__.py` to export from all Python files and not just the one it assumes has been created for the local project. 

Closes #2294. 

There is one remaining concern I have, and that is around dylib files. I believe each crate that uses a UniFFI may try to reference its own dylib file, but only a single dylib will exist in Maturin's build directory. So I'm still not sure the expected strategy here or how that should be managed for multi-crate projects.

### DIFF
--- a/src/module_writer.rs
+++ b/src/module_writer.rs
@@ -1120,7 +1120,13 @@ fn generate_uniffi_bindings(
                 .and_then(std::ffi::OsStr::to_str)
                 .map_or(false, |ext| ext == "py")
         })
-        .flat_map(|file| file.file_name().into_string())
+        .map(|file| {
+            file.path()
+                .file_stem()
+                .unwrap()
+                .to_string_lossy()
+                .to_string()
+        })
         .collect_vec();
 
     Ok(UniFfiBindings {
@@ -1176,9 +1182,10 @@ pub fn write_uniffi_module(
 
             for binding_name in binding_names.iter() {
                 let target: PathBuf = base_path.join(binding_name);
-                fs::copy(binding_dir.join(binding_name), &target).with_context(|| {
-                    format!("Failed to copy {:?} to {:?}", binding_dir.display(), target)
-                })?;
+                fs::copy(binding_dir.join(binding_name).with_extension("py"), &target)
+                    .with_context(|| {
+                        format!("Failed to copy {:?} to {:?}", binding_dir.display(), target)
+                    })?;
             }
         }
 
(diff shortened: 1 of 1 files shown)


Nickersoft:
Strange that these tests are failing – they all pass locally for me when running `cargo test`. Does anyone have an idea as to why?

messense:
@Nickersoft You need to have uniffi-bindgen in `PATH`, otherwise it will skip running related tests.

Nickersoft:
Thanks @messense! I got a lot more tests to pass now. There are still some Windows failures though, as it can't find a Python interpreter for some reason. Any idea why? Didn't think I changed anything that would affect it.

messense:
Don't worry about it, it's an existing issue.

messense (approved):
Thanks!

messense:
> There is one remaining concern I have, and that is around dylib files. I believe each crate that uses a UniFFI may try to reference its own dylib file, but only a single dylib will exist in Maturin's build directory. So I'm still not sure the expected strategy here or how that should be managed for multi-crate projects.

Can you open an issue with a repro to track this?

PR #2307, opened 2024-11-17
Fix GNU/Hurd support

sthibaul:
The detected plaform is actually gnu (from platform.system()), not gnu0 (from sys.platform).


sthibaul:
(sorry I missed taking my latest changes in the commit)

messense (approved):

PR #2317, opened 2024-11-25
Replace unmaintained fs2 with fs4

messense:


messense:

- #2317 ◀

messense:
@konstin any idea why cffi venv related tests always fail on Windows?

PR #2326, opened 2024-11-26
support interpreter names like '3.13t

ngoldbaum:
Without this, I get:

```
± maturin build --release --out dist --interpreter 3.13t
📦 Including license file "/Users/goldbaum/Documents/rpds-py/LICENSE"
🔗 Found pyo3 bindings
⚠️  Warning: Failed to determine python platform
[src/build_options.rs:1254:22] interp.display().to_string() = "3.13t"
💥 maturin failed
  Caused by: Invalid python interpreter minor version '13t', expect a digit
  Caused by: invalid digit found in string
```

### DIFF
--- a/src/build_options.rs
+++ b/src/build_options.rs
@@ -1279,8 +1279,17 @@ fn find_interpreter_in_sysconfig(
             .map(|c| c.is_ascii_digit())
             .unwrap_or(false)
         {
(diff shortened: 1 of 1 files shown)


ngoldbaum:
Are there tests for the maturin CLI somewhere?

messense:
Already done in #2324

PR #2332, opened 2024-11-28
Bump metadata version to 2.4 to fix license bug

konstin:
When a project sets `license = { file = "LICENSE" }`, we're currently emitting a `License-File` entry, but Metadata-Version 2.3, which is invalid and now gets rejected by PyPI, breaking publish for projects using `license = { file = "LICENSE" }`. This change bumps the metadata version.

Alternatively, we could only bump the version if license files are used and use the old `License` field for `license = { file = "LICENSE" }` again, but i'm not aware of any index currently rejecting metadata version 2.4.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1341,7 +1341,7 @@ dependencies = [
 
 [[package]]
 name = "maturin"
-version = "1.7.6-beta.1"
+version = "1.7.6"
 dependencies = [
  "anyhow",
  "base64 0.21.7",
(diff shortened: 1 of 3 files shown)


ddelange:
> but Metadata-Version 2.3, which is invalid and now gets rejected by PyPI

by [pinning maturin](https://github.com/oconnor663/blake3-py/commit/1c488ca59b19f99ae80e631a66e7aae08ffe55b4) to 1.7.5, we were [able to push to pypi](https://github.com/oconnor663/blake3-py/actions/runs/12092435938) (twine check within pypa/gh-action-pypi-publish passes with metadata 2.3, and apparently pypi still accepts it).

ddelange:
ref https://github.com/pypa/gh-action-pypi-publish/issues/308

ddelange:
pypa/gh-action-pypi-publish@v1.12.3 has been released with metadata 2.4 support :+1:

PR #2342, opened 2024-11-30
Upgrade pyproject-toml to 0.13.4

messense:


messense:
I'm not sure how to update to use the new `pep508_rs` api in

https://github.com/PyO3/maturin/blob/b39f8cab8d7c0e5695f0cbdd15f513ffad81df29/src/develop.rs#L154-L171

and 

https://github.com/PyO3/maturin/blob/b39f8cab8d7c0e5695f0cbdd15f513ffad81df29/src/metadata.rs#L300-L309

@konstin please help, thanks!

messense:
Oh looks like `pep440_rs` requires Rust 1.80 which is too new for our MSRV.

PR #2343, opened 2024-11-30
Normalize python source directory path

### LINKED ISSUE #2290: Function calls in Pyxel fail when using Maturin 1.7.1 or later
### Bug Description

Hi.
I tried compiling Pyxel with Maturin 1.7.1+.

I changed an option in Pyxel's makefile from `--skip-auditwheel` to `--auditwheel` skip and then the compilation finished.
But when run an application, the following error occurs:

```
    pyxel.init(300, 300, title="hoge")
    ^^^^^^^^^^
AttributeError: module 'pyxel' has no attribute 'init'
```

Maturin is called from Pyxel's makefile. https://github.com/kitao/pyxel/blob/main/Makefile#L102

The directory structure of Pyxel may no longer work correctly with Maturin 1.7.1 or later.
https://github.com/kitao/pyxel/blob/main/python/pyproject.toml#L35-L38

https://github.com/kitao/pyxel/issues/574

1.7.0 → 1.7.0+

Python 3.12.3

pip 24.2

pyo3

### Does `cargo build` work?

- [ ] Yes, it works

- [ ] Yes

### Steps to Reproduce

Just build Pyxel and use it.
- Build with `make`
- `pyxel edit` (which calls some Pyxel functions and invoke the error)

messense:
Fixes #2290

### DIFF
--- a/src/project_layout.rs
+++ b/src/project_layout.rs
@@ -144,7 +144,7 @@ impl ProjectResolver {
             .unwrap_or_default()
             .to_vec();
         let py_root = match pyproject.and_then(|x| x.python_source()) {
-            Some(py_src) => project_root.join(py_src),
+            Some(py_src) => project_root.join(py_src).normalize()?.into_path_buf(),
             None => match pyproject.and_then(|x| x.project_name()) {
                 Some(project_name) => {
                     // Detect src layout


messense:

- #2343 ◀

Issue #2346, opened 2024-11-30
non-editable install with `maturin develop --uv`


mbway:
### Bug Description

if `maturin develop --uv` is used with an environment created with `uv venv`, the package is not installed in editable mode. This is likely because `pip` is not available in the environment and there is no fallback (but I'm not sure why `fix_direct_url()` isn't crashing).

Since [uv 0.4.25](https://github.com/astral-sh/uv/releases/tag/0.4.25) the tool has supported `uv pip show --files` so assuming maturin is OK with requiring at least this version, the fallback to always use pip in `fix_direct_url()` can be removed.

uv version: `0.5.5`

### Your maturin version (`maturin --version`)

maturin 1.7.6

### Your Python version (`python -V`)

Python 3.12.2

### Your pip version (`pip -V`)

pip 24.0

### What bindings you're using

pyo3

### Does `cargo build` work?

- [X] Yes, it works

### If on windows, have you checked that you aren't accidentally using unix path (those with the forward slash `/`)?

- [ ] Yes

### Steps to Reproduce

The following script installs the `foobar` package in non-editable mode when `USE_UV=1` and in editable mode when `USE_UV=0` (the package is also installed in editable mode when using `maturin develop --uv` if the environment is created with venv):
```bash
set -e
OUT=/tmp/minimal
USE_UV=1
rm -rf $OUT

if [ $USE_UV -eq 1 ]; then
    uv venv $OUT/venv
else
    python -m venv $OUT/venv
fi

. $OUT/venv/bin/activate

maturin new --name foobar --bindings pyo3 $OUT/foobar
cd $OUT/foobar

if [ $USE_UV -eq 1 ]; then
    maturin develop --uv
else
    maturin develop
fi

uv pip list
```

messense:
Pull requests are welcome!

PR #2352, opened 2024-12-01
Remove hard coded `pip show` in `fix_direct_url`

mbway:
Before uv 0.4.25, `uv pip show` did not support the `--files` argument which is required by `fix_direct_url`. Now that the flag is supported, falling back to pip is not necessary.

Error handling is also improved. Previously if pip was not present in the environment (which is the case for `uv venv`) no error was reported to the user but `fix_direct_url` will have failed so the package is not installed in editable mode.

fixes https://github.com/PyO3/maturin/issues/2346

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,7 @@ packages = ["maturin"]
 bindings = "bin"
 
 [tool.black]
-target_version = ['py37']
+target-version = ['py37']
 extend-exclude = '''
 # Ignore cargo-generate templates
 ^/src/templates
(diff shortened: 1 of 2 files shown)


mbway:
not sure why the `develop_cffi_mixed` test is failing on windows but passing on other platforms. The other failures seem to be CI flakyness rather than something wrong

pyproject.toml
```diff
@@ -42,7 +42,7 @@ packages = ["maturin"]
 bindings = "bin"
 
 [tool.black]
-target_version = ['py37']
+target-version = ['py37']
```
mbway, line 45: my IDE said that it was the wrong key and in the black repo the key `target-version` is used

https://github.com/psf/black/blob/96ca1b6be3e543618dbcf791d1c143e7119cb139/pyproject.toml#L10

messense (approved):
LGTM, thanks!

PR #2359, opened 2024-12-03
Fix aarch64 Windows cross compilation

messense:


davidhewitt (approved):
Thanks, this looks good to me 👍

Issue #2416, opened 2024-12-28
`project.version` field is required


heinrich5991:
```
💥 maturin failed
  Caused by: `project.version` field is required in pyproject.toml unless it is present in the `project.dynamic` list
Error: command ['maturin', 'pep517', 'write-sdist', '--sdist-directory', '/home/runner/work/libtw2/libtw2/uniffi/huffman/dist'] returned non-zero exit status 1
```

I had a pyproject.toml file without
```toml
[project]
version = "1.0.0"
```
and it failed with the above error in my CI. What are the backward compatibility guarantees of maturin, is something like this expected?

andy-maier:
We also have that error in some of our projects when installing 'pywinpty', since maturin 1.8.0 was released.

The change log for 1.8.0 does not mention this incompatibility: https://github.com/PyO3/maturin/blob/main/Changelog.md#180 - Please mention incompatibilities explicitly in future change logs.

Also, the user guide does not mention the possibility of specifying a dynamic version, and how the version is determined in that case. I think the user guide should be updated to add that.

messense:
> What are the backward compatibility guarantees of maturin, is something like this expected?

This one is expected because it was considered as bugfix as the previous behavior was not spec compliant, quoting my comment in https://github.com/PyO3/maturin/pull/2417#issuecomment-2564977424

> According the spec: https://packaging.python.org/en/latest/specifications/pyproject-toml/#dynamic
> 
> > If the metadata does not list a key in dynamic, then a build back-end **CANNOT** fill in the requisite metadata on behalf of the user (i.e. dynamic is the only way to allow a tool to fill in metadata and the user must opt into the filling in).
> 
> the current behavior is spec compliant, change it to warn only may seem to resolve the error, it can resurface when user try to install a package from sdist (because wheel isn't avaiable for some reason), for eample when using `uv`: #2390

> * The change log for 1.8.0 does not mention this incompatibility

The changelog has a link to #2391, I agree it should be improved, PR welcome.

> * The user guide does not mention the possibility of specifying a dynamic version

We are following [PEP 621](https://peps.python.org/pep-0621/), IMHO it's not reasonable for us to document all of the possible options of PEP 621 in the [user guide](https://www.maturin.rs/metadata#python-project-metadata). As for this specific option, I think the solution is clearly indicated in the error message. 

> * how the version is determined in that case. I think the user guide should be updated to add that.

[...]

Issue #2390, opened 2024-12-18
Build should error if `pyproject.toml` does not contain (dynamic) version


stinodego:
### Bug Description

Currently, maturin will fill in the `version` specifier if it is missing from `pyproject.toml` (it is available in `Cargo.toml`). However, this is prohibited in the official spec:
https://packaging.python.org/en/latest/specifications/pyproject-toml/#dynamic

> If the metadata does not list a key in dynamic, then a build back-end CANNOT fill in the requisite metadata on behalf of the user

So rather than filling in the version, maturin should error and tell the user to add `dynamic = ["version"]` to their `pyproject.toml`.

This came up because `uv` is now enforcing this, and won't install any packages with a missing `version` specifier in the metadata: https://github.com/astral-sh/uv/issues/9989

### Your maturin version (`maturin --version`)

maturin 1.7.4

### Your Python version (`python -V`)

Python 3.12.2

### Your pip version (`pip -V`)

pip 24.3.1

### What bindings you're using

pyo3

### Does `cargo build` work?

- [X] Yes, it works

### If on windows, have you checked that you aren't accidentally using unix path (those with the forward slash `/`)?

- [X] Yes

### Steps to Reproduce

You can just go to the official docs:
https://www.maturin.rs/metadata

If you remove the version specifier in the pyproject.toml metadata, it will take the version from the Cargo.toml. This works, but it shouldn't!

PR #2417, opened 2024-12-29
Make invalid version info in `pyproject.toml` less fatal

mhils:
This PR is a simple fix https://github.com/PyO3/maturin/issues/2416 by making invalid version information in `pyproject.toml` less fatal. 

Thank you for your great work on Maturin! 🍰

### DIFF
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -146,11 +146,11 @@ impl Metadata24 {
             self.name.clone_from(&project.name);
             if let Some(version) = &project.version {
                 if dynamic.contains("version") {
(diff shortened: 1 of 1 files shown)


messense:
According the spec: https://packaging.python.org/en/latest/specifications/pyproject-toml/#dynamic

> If the metadata does not list a key in dynamic, then a build back-end CANNOT fill in the requisite metadata on behalf of the user (i.e. dynamic is the only way to allow a tool to fill in metadata and the user must opt into the filling in).

the current behavior is spec compliant, change it to warn only may seem to resolve the error, it can resurface when user try to install a package from sdist (because wheel isn't avaiable for some reason), for eample when using `uv`: #2390

PR #2433, opened 2025-01-11
Use `uv` automatically when running `maturin develop` inside a uv-created venv

### LINKED ISSUE #2432: Use `uv` automatically when running `maturin develop` inside a uv-created venv
```bash
$ cat .venv/pyvenv.cfg
home = /opt/homebrew/opt/python@3.13/bin
implementation = CPython
uv = 0.5.15
version_info = 3.13.1
include-system-site-packages = false
```

we should be able to detect that `uv` should be used because the venv probably won't have `pip` installed unless it was created with `uv venv --seed`:

```bash
$ cat .venv/pyvenv.cfg
home = /opt/homebrew/opt/python@3.13/bin
implementation = CPython
uv = 0.5.15
version_info = 3.13.1
include-system-site-packages = false
seed = true
```

messense:
Closes #2432

### DIFF
--- a/src/develop.rs
+++ b/src/develop.rs
@@ -162,6 +162,18 @@ fn find_uv_python(python_path: &Path) -> Result<(PathBuf, Vec<&'static str>)> {
     }
 }
 
+/// Check if a virtualenv is created by uv by reading pyvenv.cfg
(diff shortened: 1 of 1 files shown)


messense:

- #2433 ◀

PR #2446, opened 2025-01-20
Fix constant rebuilds

TheZoq2:
On some of my computers, maturin build always rebuilds pyo3 and friends, and looking at the cargo fingerprinting, this is caused by the py03-config file being written regardless of changes. This only writes the file if it has changed since last rebuild, which seems to prevent the problem.


messense (approved):
Thanks!

PR #2449, opened 2025-01-23
fix the name of the .data directory in the generated wheel

ejd:
The basename of the ".data" directory in the wheel now matches the basename of the ".dist-info" directory, as per the wheel spec[1], like so:

     distribution-1.0.dist-info
     distribution-1.0.data

1: https://packaging.python.org/en/latest/specifications/binary-distribution-format/#the-data-directory


ejd:
This PR fixes #2450

konstin (approved):

PR #2457, opened 2025-01-28
Mention zig in readme and introduction page

konstin:
Zig is the easiest way to build a manylinux wheel, much easier than the manylinux docker container, so we should point users to it more prominently.


jayaddison:
@konstin I was trying to understand why Zig is enabled for the build of wheels for `uap-python` (a library that includes some Rust code) in pull request: https://github.com/ua-parser/uap-python/pull/315

(the PR stem from a [bugreport](https://github.com/ua-parser/uap-python/issues/312) I'd opened, before I noticed this `README` text)

What are the features/components that Zig provides that enables the wheel build?

konstin:
Please do not comment on old pull requests, but consult the documentation (it has lots of information on zig), if there are questions remaining, open a new issue.

PR #2513, opened 2025-03-13
fix auditwheel .so relocation for namespace modules

pascalkuthe:
For a project of mine I have a setup as a namespace package like this:

``` toml
[tool.maturin]
python-source = "python/src"
module-name = "my.example.namespace.module"
```

The module has a shared library dependency and currently the shared object generated by audit wheel fails to load as the rpath gets set incorrectly. The wheel has the following contents

```
/my.libs/my_dep.so
/my/example/namespace/module.so
```
 The `RATH`/`RUNPATH` for `/my/example/namespace/module.so` gets set to `$ORIGIN/../my.libs/my_dep.so` which means that `my_dep.so` is not found while trying to load the module. Instead the `RATH`/`RUNPATH` should be: `$ORIGIN/../../../my.libs/my_dep.so`.

I took a stab at fixing this. It looks like the logic for determining the artifact path was wrong and assumed the shared library would endup in `my.example.namespace.module/module.so` but it endsup in `my/example/namespace/module.so`.

### DIFF
--- a/src/build_context.rs
+++ b/src/build_context.rs
@@ -419,6 +419,14 @@ impl BuildContext {
         let artifact_dir = match self.bridge() {
             // cffi bindings that contains '.' in the module name will be split into directories
             BridgeModel::Cffi => self.module_name.split(".").collect::<PathBuf>(),
+            // For namespace packages the modules the modules resides resides at ${module_name}.so
+            // where periods are replaced with slashes so for example my.namespace.module would reside
+            // at my/namespace/module.so
+            _ if self.module_name.contains(".") => {
+                let mut path = self.module_name.split(".").collect::<PathBuf>();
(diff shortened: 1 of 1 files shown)


messense (approved):
Thanks!

pascalkuthe:
Tha ks for the quick response! This issue is a blocker for me at work right now. I am weighing whether to setup a from-source build internally or wait for the next release.

Do you have a rough plan for when you would release a new version or will it be a while?

messense:
I can release a new patch version tonight after work, or you can speed it up by sending a PR to update changelog & version like #2469 :-).

pascalkuthe:
that would be fantistic!. I will do a PR similiar like that to help. Thanks! :)

PR #2514, opened 2025-03-13
Updates Rng legacy methods in tutorial

rafaelbeckel:
## Changes

- Renames `thread_rng()` to `rng()`
- Renames `gen_range()` to `random_range()`


messense:
I guess we'd also need to update rand version here?
https://github.com/PyO3/maturin/blob/30cdcbdf13c7c91b7a5158adf0cfa870adc9f26a/guide/src/tutorial.md#L36

PR #2527, opened 2025-03-19
Parse min/max supported CPython/PyPy versions from pyo3-ffi metadata

messense:


src/bridge.rs
```diff
@@ -89,6 +169,16 @@ impl PyO3 {
         }
     }
 
+    /// Returns the maximum PyPy minor version supported
+    fn maximal_pypy_minor_version(&self) -> usize {
```
messense, line 173:
```suggestion
    fn maximum_pypy_minor_version(&self) -> usize {
```

src/bridge.rs
```diff
@@ -159,6 +253,16 @@ impl BridgeModel {
         }
     }
 
+    /// Returns the maximum PyPy minor version supported
+    pub fn maximum_pypy_minor_version(&self) -> usize {
+        use crate::python_interpreter::MAXIMUM_PYPY_MINOR;
+
+        match self.pyo3() {
+            Some(bindings) => bindings.maximal_pypy_minor_version(),
```
messense, line 261:
```suggestion
            Some(bindings) => bindings.maximum_pypy_minor_version(),
```

PR #2550, opened 2025-04-10
Sort RECORD file in wheel archives to make them deterministic.

3pleX-dev:
While working on reproducible builds in nixpkgs ([#397290](https://github.com/NixOS/nixpkgs/pull/397290), [#384709](https://github.com/NixOS/nixpkgs/issues/384709)), I noticed that the RECORD files generated in wheels had inconsistent ordering between builds.

It looked like this was due to both the WheelWriter and PathWriter implementations writing entries in the order they were processed, without any explicit sorting. This PR adds a simple sort step before writing to ensure deterministic output.

I'm not familiar with Rust yet, so if there's anything I’ve missed or done in a non-idiomatic way, I’d really appreciate any feedback or suggestions. Thanks!

### DIFF
--- a/src/module_writer.rs
+++ b/src/module_writer.rs
@@ -153,7 +153,11 @@ impl PathWriter {
             record_file.display()
         ))?;
 
-        for (filename, hash, len) in self.record {
+        // Sort records for deterministic output
+        let mut sorted_records = self.record.clone();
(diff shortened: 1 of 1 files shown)


messense (approved):

3pleX-dev:
Hi @messense please I'm not sure why the lint check is failing, the error in logs looks like it's not something related to this pr right?

messense:
Dont worry about it, it was fixed in #2551

Issue #2561, opened 2025-04-20
cargo-metadata invoked by maturin resolves dependencies differently from cargo fetch with --target specified


ziyao233:
### Bug Description

With --target argument supplied, cargo-metadata invoked by maturin during building may resolve dependencies differently from cargo-fetch, failing a --frozen build with unresolvable crates that are actually unnecessary on the target.

### Your maturin version (`maturin --version`)

1.8.3

### Your Python version (`python -V`)

3.13.3

### Your pip version (`pip -V`)

25.0.1

### What bindings you're using

pyo3

### Does `cargo build` work?

- [x] Yes, it works

### If on windows, have you checked that you aren't accidentally using unix path (those with the forward slash `/`)?

- [ ] Yes

### Steps to Reproduce

Taking rpds-py as example,

```
export CARGO_HOME=$PWD/cargocache # ensure a clean environment
cargo fetch --locked --target x86_64-unknown-linux-musl

maturin build --frozen --target x86_64-unknown-linux-musl
cargo build --frozen --target x86_64-unknown-linux-musl 
```

Maturin fails to build, with error message like

```
$ maturin build --frozen --target x86_64-unknown-linux-musl
error: failed to download `cc v1.0.90`

Caused by:
  attempting to make an HTTP request, but --frozen was specified
💥 maturin failed
  Caused by: Cargo metadata failed. Does your crate compile with `cargo build`?
  Caused by: `cargo metadata` exited with an error: 
```

but cargo builds the crate smoothly,

```
$ cargo build --target x86_64-unknown-linux-musl --frozen
   Compiling target-lexicon v0.13.2
   Compiling python3-dll-a v0.2.12
   Compiling once_cell v1.19.0
   Compiling proc-macro2 v1.0.86
   Compiling unicode-ident v1.0.12
   Compiling autocfg v1.3.0
   Compiling libc v0.2.155
   Compiling heck v0.5.0
   Compiling memoffset v0.9.1
...
```

Without specifying `--target x86_64-unknown-linux-musl` when invoking `cargo fetch`, both cargo and maturin successfully build the crate. The problematic dependency, `cc`, is only necessary on windows  and shouldn't be pulled in with `--target` specified.

```
[target.'cfg(windows)'.dependencies]
cc = "1.0.73"
```

PR #2562, opened 2025-04-20
Convert --target to --filter-platform and pass it to cargo-metadata

ziyao233:
An explicitly-specified target may simplify the dependency graph, i.e. crates unused on the target may be omitted. Cargo-fetch does this type of simplification and may pull fewer packages with --target argument.

maturin build doesn't pass the specified triple to cargo-metadata currently as --target isn't supported in cargo-metadata. When building reproducibly with cargo-fetch and --frozen argument, this could ultimately lead to a confusing and annoying dependency that cannot be resolved offline, since cargo-metadata has no idea about the target and cannot just drop the unnecessary dependency.

However, cargo-metadata does support --filter-platform since Cargo 1.40, which plays basically the same role as --target during dependency resolving. Let's translate --target to --filter-platform when calling cargo-metadata to handle the case.

Closes: https://github.com/PyO3/maturin/issues/2561

---

This is the first time that I play with Rust, thanks for your patience and guide.


ziyao233:
The failed test seems some random network issues

messense (approved):
LGTM

PR #2572, opened 2025-04-27
Add `--compression-level` option to `build` command

### LINKED ISSUE #2566: build: please add support for specifying the compression level
While the primary use case for PEP517 build backends is to built wheels for redistribution, distribution packagers are also forced to use them to build their distribution packages. This implies packing a wheel that is unpacked and discarded immediately afterwards. As such, compressing such a wheel makes little sense, as it only wastes time and energy on compressing, only to waste more energy decompressing it a minute later.

In Gentoo, we're usually working around the problem by patching the Python `zipfile` module to implicitly override the compression level. However, that approach works only for pure Python backends.

Could you please consider accepting a command-line option to the `build` command that would permit the user to override the zip compression level used? I can attempt implementing it if you're interested — I've been doing pretty well at guesswork in Rust so far.

mgorny:
Add a `--compression-level` hidden option that can be used to adjust deflate compression level used for written wheels. When not provided, the default level of 6 is used (as before). When provided, accepts 0 to store the file uncompressed, and 0-264 otherwise as `zip` crate accepts.

Fixes #2566

### DIFF
--- a/src/build_context.rs
+++ b/src/build_context.rs
@@ -131,6 +131,8 @@ pub struct BuildContext {
     pub editable: bool,
     /// Cargo build options
     pub cargo_options: CargoOptions,
+    /// Zip compression level
+    pub compression_level: u16,
 }
 
 /// The wheel file location and its Python version tag (e.g. `py3`).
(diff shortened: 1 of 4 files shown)


messense (approved):

mgorny:
Thank you!

PR #2575, opened 2025-04-29
Add `tpchgen-cli` python library to examples

kevinjqliu:
[`tpchgen`](https://github.com/clflushopt/tpchgen-rs) is a TPC-H data generator written in rust. Here's more background on its development, https://datafusion.apache.org/blog/2025/04/10/fastest-tpch-generator/

`maturin` was used to distribute the rust binary as python script, using [its `bin` bindings feature](https://www.maturin.rs/bindings#bin), with minimal code change (See https://github.com/clflushopt/tpchgen-rs/pull/121)

The rust binary cli can now be pip installed,
```
pip install tpchgen-cli
tpchgen-cli -h
```

Took me a while to figure out that `maturin` had first-class support for something like this. My first attempt was to export the cli functions manually (See https://github.com/clflushopt/tpchgen-rs/pull/119). Because of this, I think this use case is a great example to add. 

(Following #2529 as example PR)


messense (approved):

messense:
Thanks!

PR #2584, opened 2025-05-07
Fix broken `maturin develop` with latest uv

thomasjo:
This patch fixes `maturin develop` not working with the latest versions of uv due to breaking changes in uv. Previously the `--version` argument could be supplied to any subcommand, but since v0.7 that argument can only be supplied as `uv --version`.

Specifically, the fix changes the `uv pip` command to `uv` when we get the uv version as part of the check if `show --files` is supported.

Without this patch `configure_as_editable` fails due to `check_supports_show_files` failing, which again fails because `InstallBackend::version` fails with uv 0.7.x or newer.

See https://github.com/astral-sh/uv/pull/13108 for more details.

I've not added any new tests since I believe the current test suite covers this change. The change is backwards compatible with older versions of uv. But if a new test is required, please advice exactly what/how you want tested, and I'll add it.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [1.8.4]
+
+* Fix broken `maturin develop` with latest uv in [#2584](https://github.com/PyO3/maturin/pull/2584)
+
 ## [1.8.3]
 
 * Fix relocating shared library for namespace modules in [#2513](https://github.com/PyO3/maturin/pull/2513)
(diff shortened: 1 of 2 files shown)


thomasjo:
I just noticed I've got a typo in the first commit, `maturing ...` instead of `maturin ...`, but it looks like you squash when merging PRs, so I'll ignore it unless you want me to fix it 😅

messense (approved):
Thanks!

PR #2585, opened 2025-05-08
Release 1.8.4

konstin:
Getting #2421 out.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1397,7 +1397,7 @@ dependencies = [
 
 [[package]]
 name = "maturin"
-version = "1.8.3"
+version = "1.8.4"
 dependencies = [
  "anyhow",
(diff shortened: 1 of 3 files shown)


youknowone:
#2572 was a breaking change, which included in this release.

Due to this change, 1.8.4 is not compatible to 1.8.3. This breaks semver expectation.

konstin:
Can you explain what broke?

youknowone:
`BuildOptions` is a public type. User can make an object of it by manually writing every field. The patch added an additional field to `BuildOptions`, which leads the previous code to be broken.

Probably adding `#[non_exhaustive]` will prevent it.

messense:
The `maturin` library has no API stability guarantee currently, only the binary has.

youknowone:
Oh, thank you!

PR #2587, opened 2025-05-09
Remove Readme CI badges

konstin:
Nowadays, GitHub shows CI status inline.


messense:
We don't run CI on readme file changes only, merged manually.

PR #2589, opened 2025-05-09
Type target triple to fix regression.

konstin:
#2562 introduced a regression breaking publishing maturin (https://github.com/PyO3/maturin/actions/runs/14923136319/job/41923286669). This PR moves the target triple type from a String to an enum, forcing all consumers to handle the virtual universal2 target themselves.

I tested this as far as possible on Linux, but it's missing real macOS testing (there are two different error messages for universal2 build before and after the regression and it's now back to the "good" error message).


messense (approved):
LGTM, Thanks!

PR #2597, opened 2025-05-12
fixes #1894 -- when the abi3 feature is set with no explicit version, use the current python interpreter's version

### LINKED ISSUE #1894: Build option to dynamically set abi3 version
Currently the way to set abi3 version is via applying a versioned-feature in `Cargo.toml` (e.g., `abi3-py312`).

I'm requesting a way set that at the CLI. The reason for this is to be able to build wheels targeting multiple abi3 versions from the same source.

We're currently able to do this with `setuptools-rust` with the following command:

```
if [ -n "${{ matrix.PYTHON.ABI_VERSION }}" ]; then
    PY_LIMITED_API="--config-settings=--build-option=--py-limited-api=${{ matrix.PYTHON.ABI_VERSION }} --no-build-isolation"
fi
OPENSSL_DIR="/opt/pyca/cryptography/openssl" \
    OPENSSL_STATIC=1 \
    .venv/bin/python -m pip wheel -v $PY_LIMITED_API cryptograph*.tar.gz -w dist/ && mv dist/cryptography*.whl tmpwheelhouse
```

alex:


messense (commented):
The change looks good to me, I'm just a bit worried that this can be considered as a breaking change which can't be released in a 1.x version.

messense:
I agree, let's move on and release a new version to also fix #2599

alex:
awesome, thanks!

PR #2614, opened 2025-05-20
Upgrade cbindgen to 0.29

youknowone:
It also removes heck 0.4.1 dependency


youknowone:
The failing test is network timeout. Could you trigger it again?

messense (approved):

Issue #2622, opened 2025-05-31
`maturin develop` hardcodes compression level which adds signifinicant latency to the development loop


KarelPeeters:
## Overview

Recently, https://github.com/PyO3/maturin/pull/2572 added an option `--compression-level` to the `build` subcommand, which can be used to lower the compression level to speed up the build flow. The specific reasoning in the original issue https://github.com/PyO3/maturin/issues/2566 was related to distribution packages.

A similar reasoning applies to the `develop` subcommand: the artifacts that are build will only be used for local development, so their size is not that important. During local development iteration speed is a major concern, so disabling compression seems like a worthwhile tradeoff to me. The compression level for `develop` is currently hardcoded to `6` here: https://github.com/PyO3/maturin/blob/fd59bd33510e85fc157002c8101a0e2f2d1c517c/src/develop.rs#L419

## Benchmarks

For my current project changing the compression level to `0` lowers the time that `maturin dev` takes from 2.7s to 1.0s, a significant reduction.

## Solutions

I see a couple possible solutions here:
* Just hardcode the compression level to `0`, under the assumption that no one really needs to configure.
* Add a `--compression-level` option to the `develop` subcommand, similar to what was done for `build`. In this case I suggest that we set the default to `0`, to speed up everyone's workflow by default.

I'd be happy to submit a PR that implements either of these choices, I don't have a strong preference for either one.

KarelPeeters:
Thanks for the quick fix!

PR #2625, opened 2025-06-02
Allow specifying compression method and level, in both `build` and `develop` modes.

corentin-ant:
This is a follow-up to https://github.com/PyO3/maturin/issues/2566 that adds the ability to
- Specify other compression methods
- Control compression in the `develop` mode, not only `build`. Solves #2622

```
--compression-method <COMPRESSION_METHOD>
    Zip compresson method. Only Stored and Deflated are currently compatible with all package managers

    [default: deflated]

    Possible values:
    - deflated: Deflate compression
    - stored:   No compression
    - bzip2:    BZIP2 compression
    - zstd:     Zstandard compression (supported from Python 3.14)

--compression-level <COMPRESSION_LEVEL>
    Zip compresson level

    [default: 6]
```

Note that additional compression methods were added in `uv` 0.7.3 in https://github.com/astral-sh/uv/pull/13285 but adding local `whl` is broken until https://github.com/astral-sh/uv/pull/13781 is merged.

I tested the change locally by setting
```
        compression: maturin::CompressionOptions {
            compression_method: maturin::CompressionMethod::Bzip2,
            compression_level: 6,
        },
```
in the tests, which is supported by `pip`. I can do a follow-up PR to update the tests after `uv` is updated, possibly with a version check.

### DIFF
--- a/src/module_writer.rs
+++ b/src/module_writer.rs
@@ -1,4 +1,5 @@
 //! The wheel format is (mostly) specified in PEP 427
+use crate::compression::{CompressionMethod, CompressionOptions};
 use crate::project_layout::ProjectLayout;
 use crate::target::Os;
 use crate::{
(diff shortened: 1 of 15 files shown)


src/module_writer.rs
```diff
@@ -239,7 +240,26 @@ pub struct WheelWriter {
...
+        let compression_method = if cfg!(feature = "faster-tests") || compression_level == 0 {
+            zip::CompressionMethod::Stored
+        } else {
+            self.compression_method.into()
+        };
+
+        let mut options =
+            zip::write::SimpleFileOptions::default().compression_method(compression_method);
+        if compression_level != 0 {
+            options = options.compression_level(Some(compression_level as i64));
```
messense: I'm a bit worried that the compression level value varies among compression methods: https://docs.rs/zip/latest/zip/write/struct.FileOptions.html#method.compression_level

see also https://github.com/PyO3/maturin/discussions/2608 which could be related.

corentin-ant: One option would be normalizing, but it feels a bit wrong too. We could:
1. Write the level range in the command line help and/or
2. Check the validity of the level/method combination in the CLI, before it causes an error when calling `zip` and/or
3. Ensure that the default level is reasonable for each method (currently all are set to 6).

My suggestion would be to do at least 1. and 3., what do you think?

messense: Sounds good to me!

corentin-ant: Implemented in the update :)

```
$ maturin build --compression-method deflated --compression-level -1
error: Invalid level -1 for compression method Deflated. Use a level in range 0..=9

Usage: maturin [OPTIONS]

For more information, try '--help'.

messense (approved):
Thanks!

PR #2644, opened 2025-06-10
`ZipWriter` requires a compression level of `None` for the `stored` method

jatoben:
I tried using the `stored` compression level in Maturin 1.8.7 with my project, but I hit this error:

```
% maturin develop -b pyo3 -m path/to/Cargo.toml --compression-method stored --profile release
🔗 Found pyo3 bindings
🐍 Found CPython 3.13 at /Users/jatoben/.pyenv/versions/3.13.3/envs/project-3.13.3/bin/python
📡 Using build options features from pyproject.toml
    Finished `release` profile [optimized] target(s) in 0.02s
💥 maturin failed
  Caused by: unsupported Zip archive: Unsupported compression level
```

The same issue also occurred when I ran `cargo test --features faster-tests`.

ZipWriter [requires the compression level to be `None`](https://docs.rs/zip/2.3.0/zip/write/struct.FileOptions.html#method.compression_level) if the method is `CompressionMethod::Stored`. I didn't want to change the existing external interface, so I added a check for this case inside `CompressionOptions::get_file_options()`.

### DIFF
--- a/src/module_writer.rs
+++ b/src/module_writer.rs
@@ -1610,4 +1610,27 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn wheel_writer_no_compression() -> Result<(), Box<dyn std::error::Error>> {
+        let metadata = Metadata24::new("dummy".to_string(), Version::new([1, 0]));
+        let tmp_dir = TempDir::new()?;
+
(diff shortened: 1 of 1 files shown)


messense (approved):
Thanks!

PR #2650, opened 2025-06-13
Add `--compatibility pypi` to avoid building for unsupported architectures

konstin:
This was motivated by https://github.com/astral-sh/uv/issues/14007, where I accidentally assumed PyPI would support all manylinux targets. To prevent such regressions in the future by adding `--compatibility pypi`. This flag ensures we're only building for architectures compatible with PyPI's `legacy.py` validation. It falls use the default compatibility internally, i.e. it uses manylinux on linux.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,13 +1,13 @@
 **/.pytest_cache
 **/.tox
 **/__pycache__
-**/target
 **/venv*
 .dockerignore
 .git
 .idea
 .travis.yml
 .venv
+/target
 Dockerfile*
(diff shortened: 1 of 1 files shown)


guide/src/distribution.md
```diff
@@ -203,6 +159,45 @@ Manifest Options:
           Run without accessing the network
 ```
 
+## Source Distribution
```
konstin, line 162: This block only moved

messense (approved):

Issue #676, opened 2021-11-12
Problem with abiflags check for Python3 under MINGW64


sdbbs:
I use Python3 under MINGW64 (as part of MSYS2) - and I stumbled upon this problem.

I also found a workaround - but it should be noted, that I have never programmed in Rust before, so I have no idea if it is correct.

First, my info:

 * Your python version (`python -V`): Python 3.9.7 (installed via `pacman -S mingw-w64-x86_64-python`)
 * Your pip version (`pip -V`): "pip 21.3 from C:/msys64/mingw64/lib/python3.9/site-packages/pip (python 3.9)" (installed via `pacman -S mingw-w64-x86_64-python-pip`)
   * If this isn't the [latest pip version](https://pip.pypa.io/en/stable/news/), have you checked using the latest pip version? - not applicable, MINGW64 so far can only manage pip from its binary package ( https://stackoverflow.com/questions/56930492/pip-on-mingw64 )
 * The version of the bindings you're using, if any (e.g. `pyo3`, `rust-cpython` or `cffi`): - No idea (do not have these names as binaries)
 * Does `cargo build` work? - Yes (rust installed via `pacman -S mingw-w64-x86_64-rust`)
 * If on windows, have you checked that you aren't accidentally using unix path (those with the forward slash `/`)? - it seems so, but that seems to be acceptable under MINGW64

> Please list the exact steps required to reproduce your error with all command output and if possible with a repository:

Right, so, in trying to install some other pip package for Python3 under MINGW64, I stumbled upon `pywinpty` as a dependency, that failed. So, I tried to build it on its own, and keep the build folders for inspection:

```none
user@MYPC 2021-11-12|11:39:42 MINGW64 ~
$ pip3 install --no-clean pywinpty
Collecting pywinpty
  Using cached pywinpty-1.1.5.tar.gz (48 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing wheel metadata (pyproject.toml) ... error
  ERROR: Command errored out with exit status 1:
   command: C:/msys64/mingw64/bin/python.exe C:/msys64/mingw64/lib/python3.9/site-packages/pip/_vendor/pep517/in_process/_in_process.py prepare_metadata_for_build_wheel C:/Users/user/AppData/Local/Temp/tmp87jrb6is
       cwd: C:/Users/user/AppData/Local/Temp/pip-install-hf_xj6dt/pywinpty_e675807829e349a984464ae599e331ef
[...]

PR #2663, opened 2025-06-24
fix: improve Windows Python detection in abiflags validation

leeguooooo:
## Summary

This PR fixes an issue where Windows Python interpreters could be misidentified as Linux/macOS, causing false "must define abiflags" errors. This particularly affects builds using `uv` on Windows.

## Problem

When building Python packages with Rust extensions on Windows using tools like `uv`, maturin can fail with:

```
maturin failed: A python 3 interpreter on Linux or macOS must define abiflags in its sysconfig ಠ_ಠ
```

This happens even though the build is on Windows, not Linux or macOS.

## Root Cause

The `fun_with_abiflags` function has logic that:
1. Checks if `message.system == "windows"` for Windows-specific handling
2. If not Windows, assumes Unix-like system and expects abiflags
3. Fails if abiflags are None/empty

However, in some cases (particularly with `uv`), `platform.system()` might not correctly identify Windows, leading to this false error.

## Solution

This PR adds a fallback detection:
- If abiflags are empty/None AND the target OS is Windows (via `target.get_python_os()`), handle it as Windows Python
- Improved error messages to include both detected system and target info for debugging

## Changes

1. Added fallback Windows detection when `message.system` is incorrect but `target` is Windows
2. Enhanced error messages with more context (detected system, target, abiflags)
3. Reuses existing Windows abiflags logic for the fallback case

## Testing

The fix handles these scenarios:
- Normal Windows Python with correct detection ✓
- Windows Python misdetected as Linux (the uv issue) ✓ 
- Actual Linux/macOS Python with proper abiflags ✓

## Related Issues

- Reported in: https://github.com/aki66938/xhs-toolkit/issues/7
- Similar to #676 but for different environment (uv instead of MINGW64)

## Impact

This allows Windows users to build packages with `uv` and other tools without encountering false platform detection errors.

### DIFF
--- a/src/python_interpreter/mod.rs
+++ b/src/python_interpreter/mod.rs
@@ -455,7 +455,31 @@ fn fun_with_abiflags(
             Ok(abiflags.to_string())
         }
     } else {
-        bail!("A python 3 interpreter on Linux or macOS must define abiflags in its sysconfig ಠ_ಠ")
+        // Check if this might be a Windows Python that wasn't detected correctly
+        if matches!(message.abiflags.as_deref(), Some("") | None) && target.get_python_os() == "windows" {
+            // Handle as Windows Python
+            if message.minor <= 7 {
+                Ok("m".to_string())
+            } else if message.gil_disabled {
+                ensure!(
+                    message.minor >= 13,
+                    "gil_disabled is only available in python 3.13+ ಠ_ಠ"
+                );
+                Ok("t".to_string())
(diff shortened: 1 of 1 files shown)


messense:
> `platform.system()` might not correctly identify Windows

could you elaborate on this? also cc @konstin for python-build-standalone.

Note that you need to run `cargo fmt` to format the changes.

PR #2694, opened 2025-07-27
Add support for RISC-V architecture in manylinux and update compatibi…

ffgan:
Recently, pypi has just provided support for riscv wheel uploads. For details, please see https://github.com/pypi/warehouse/pull/18390

Here you can see that test.pypi.org has deployed support, https://github.com/pypi/warehouse/deployments/test.pypi.org

I think maturin-action can also provide support for this, mainly by setting manylinux_2_39 as the default and supported version of riscv64, so that riscv64 packages that can be uploaded to pypi can be packaged

Here are the results of my github action run, 

CI result:
[release](https://github.com/ffgan/maturin/actions/runs/16551719275)
[test](https://github.com/ffgan/maturin/actions/runs/16551720338)

The results show that riscv64 runs well

**Other Info**

Co-authored-by: @kotvaer
Co-authored-by: nijincheng@iscas.ac.cn；

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -389,6 +389,11 @@ jobs:
             abi: cp310-cp310
             python: python3.10
             container: ghcr.io/rust-cross/manylinux_2_36-cross:loongarch64
+          - target: riscv64gc-unknown-linux-gnu
+            abi: cp310-cp310
+            python: python3.10
+            container: ghcr.io/rust-cross/manylinux_2_39-cross:riscv64
+            extra-args: --manylinux 2_39
           # PyPy
           - target: aarch64-unknown-linux-gnu
             abi: pp310-pypy310_pp73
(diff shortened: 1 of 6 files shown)


src/target/mod.rs
```diff
@@ -483,7 +483,7 @@ impl Target {
             }
             Arch::Riscv64 => PlatformTag::Manylinux {
                 major: 2,
-                minor: 31,
+                minor: 39, // 2_39 is the minimum version supported uploaded to PyPI
```
konstin: Where does it say that that's the minimum version?

ffgan: this link,  https://github.com/pypi/warehouse/pull/18390
This link shows that 2_39 is used as riscv support. I haven't seen any other versions support riscv yet, so I temporarily assume that he uses 2_39 as the lowest version.

ffgan: this link,  https://github.com/pypi/warehouse/pull/18390
This link shows that 2_39 is used as riscv support. I haven't seen any other versions support riscv yet, so I temporarily assume that he uses 2_39 as the lowest version.

ffgan: What I want to express here is that 2_39 is the lowest version that can be uploaded to pypi, which may be different from the meaning here. Please point out if there is anything wrong.

messense: It's fine to keep it `2_31`, see https://github.com/PyO3/maturin/blob/f18f7760d4851946d421c019178582031d0dce83/src/auditwheel/manylinux-policy.json#L420-L427

[pypi doesn't really care about the actually glibc version in the tags](https://github.com/pypi/warehouse/blob/bf2597a7927425e364cef5f10b9de68cf85c241e/warehouse/forklift/legacy.py#L180-L201), we just need to make sure it has a defined manylinux policy for that glibc version.

.github/workflows/test.yml
```diff
@@ -420,7 +420,45 @@ jobs:
 
           # --find-interpreter
           bin/maturin build --find-interpreter --release --out dist --target ${{ matrix.platform.target }} -m test-crates/pyo3-mixed/Cargo.toml
+  test-riscv64-compile:
```
messense: please merge it with `test-cross-compile`

src/target/mod.rs
messense:
```suggestion
                minor: 31,
```

src/target/pypi_tags.rs
```diff
@@ -96,6 +97,17 @@ fn is_platform_tag_allowed_by_pypi(platform_tag: &str) -> bool {
...
+        if arch == "riscv64" && libc == "many" {
+            let parts: Vec<&str> = platform_tag.split('_').collect();
+            let major: u8 = parts[1].parse().expect("parse major failed");
+            let minor: u8 = parts[2].parse().expect("parse minor failed");
+            // pypi now support manylinux_2_39_riscv64
+            if major == 2 && minor >= 39 {
+                return true;
+            }
+            return false;
+        }
```
messense:
```suggestion
```

.github/workflows/test.yml
ffgan: Since pypi only supports uploading manylinux_2_39 packages, I think it is better to test 2_39 for testing, so I did not merge it into test-cross-compile.

If merged into test-cross-compile, the package will be manylinux_2_34, I am not sure whether it is suitable

ffgan: In this separate test, I specified the compiled manylinux_2_39 package at the end

src/target/pypi_tags.rs
ffgan: According to the name, the function is_platform_tag_allowed_by_pypi should be used to determine whether it is supported by pypi and whether the riscv64 should be retained.

.github/workflows/test.yml
messense: github actions can use `matrix` for such customizations.

src/target/pypi_tags.rs
messense: `riscv64` is already added in `MANYLINUX_ARCHES` right? Please see https://github.com/pypi/warehouse/blob/bf2597a7927425e364cef5f10b9de68cf85c241e/warehouse/forklift/legacy.py#L180-L201, this should be a faithful port of this python function.

messense: BTW, `uv` has successfully uploaded riscv64 2_31 wheel to pypi: https://pypi.org/project/uv/#uv-0.8.3-py3-none-manylinux_2_31_riscv64.whl

PR #2717, opened 2025-08-15
Add use-base-python option to pyproject.toml

### LINKED ISSUE #2715: Allow configuring `MATURIN_PEP517_USE_BASE_PYTHON` via `pyproject.toml`
Maturin's build backend has support for resolving the base Python executable before invoking maturin, preventing rebuilds when switching venvs, or building in an temporary isolated venv. However, this causes problems with some projects (#2133), and so this feature was guarded behind a `MATURIN_PEP517_USE_BASE_PYTHON` environment variable.

It feels to me that whether we should resolve the base version of Python is more a property of the project itself, rather than environment, and so this config option should additional be read from `pyproject.toml`.

I think this is a fairly simple patch, just adding a [or `get_config()["use-base-python`]` to here](https://github.com/PyO3/maturin/blob/c01299deba8a175595933a3dae8b909552752c01/maturin/__init__.py#L53) (plus docs, etc...). I'm happy to implement this, but wanted to check if such a patch was desirable first.

SquidDev:
This adds a new `use-base-python = false` option to Maturin's `pyproject.toml` section, with the same behaviour as `MATURIN_PEP517_USE_BASE_PYTHON`. Fixes #2715.

I've added this entry to the schema manually — running `cargo run --features schemars -- generate-json-schema` introduced a whole bunch of extra changes (seemingly from the changes in cee86ffb4b66b6ba87a0d729c7019812593e7988), so didn't want to make the diff too noisy.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Add a `use-base-python` option to `pyproject.toml` with the same behaviour as `MATURIN_PEP517_USE_BASE_PYTHON`.
+
 ## [1.9.3]
 
 * Fix adding `project.license-files` to source distributions.
--- a/guide/src/config.md
+++ b/guide/src/config.md
@@ -59,6 +59,13 @@ strip = true
 # Source distribution generator,
 # supports cargo (default) and git.
 sdist-generator = "cargo"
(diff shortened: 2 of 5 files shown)


SquidDev:
Thank you for merging!

PR #2723, opened 2025-08-21
Set `PYO3_BUILD_EXTENSION_MODULE` env var when building pyo3 extension modules

alex:
See https://github.com/PyO3/pyo3/issues/5202 for context

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -411,6 +411,11 @@ fn cargo_build_command(
         }
     }
 
+    // Set PYO3_BUILD_EXTENSION_MODULE when building pyo3 extension modules
+    if bridge_model.is_pyo3() && !bridge_model.is_bin() {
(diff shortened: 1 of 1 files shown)


alex:
clippy warnings are not related to this branch

src/compile.rs
```diff
@@ -411,6 +411,11 @@ fn cargo_build_command(
         }
     }
 
+    // Set PYO3_BUILD_EXTENSION_MODULE for all PyO3 builds
+    if bridge_model.is_pyo3() {
```
messense: This will also set it for bin targets that use `pyo3` which are not extension modules, see
https://github.com/PyO3/maturin/blob/8ab42219247277fee513eac753a3e90e76cd46b9/src/bridge.rs#L223-L226

alex: thanks, fixed.

PR #2734, opened 2025-09-02
search_lib_dir: Only recurse into `build*` if it is a directory.

xhochy:
This avoids issues with Python 3.14 and cross-compiling where there is a `build-details.json` file now.


messense (approved):
Thanks!

kylebarron:
Awesome, thank you! Looking forward to a patch release with this to unblock conda-forge packaging for Python 3.14

xhochy:
The patch is applied to the latest maturin build on conda-forge, so no need for workarounds there anymore.

PR #2750, opened 2025-10-03
Only select Windows Python interpreters with matching CPU pointer widths and architectures

### LINKED ISSUE #2692: Invalid interpreter used on `windows-11-arm` runners
### Bug Description

See https://github.com/pydantic/jiter/actions/runs/16471086907/job/46560214162#step:5:160

```
🐍 Found CPython 3.11 at C:\hostedtoolcache\windows\Python\3.11.9\x64\python.exe, CPython 3.12 at C:\hostedtoolcache\windows\Python\3.12.9\x64\python.exe, CPython 3.13 at C:\hostedtoolcache\windows\Python\3.13.5\arm64\python.exe, CPython 3.13t, CPython 3.14, CPython 3.14t
```

It looks like the `windows-11-arm` GitHub runners contain some x64 Python versions on their filesystem, and `maturin` is incorrectly recognising these as usable interpreters when attempting to build for Windows aarch64.

This leads to link failures later in the build.

0.19.1

3.11

n/a

pyo3

### Does `cargo build` work?

- [x] Yes, it works

- [x] Yes

### Steps to Reproduce

Try to build a project on a `windows-11-arm` GitHub runner, with `generate-import-lib` feature enabled.

The cached x64 Python versions should be ignored, and all versions built with generated import libs.

tonybaloney:
Fixes #2692 

Updates the Windows Python discovery to use `sysconfig.get_platform()` then matches correct CPU architecture.

Since the hashmap is (major, minor), once an AMD64 Python is found for that version, the ARM64 version will be ignored.

### DIFF
--- a/src/python_interpreter/mod.rs
+++ b/src/python_interpreter/mod.rs
@@ -29,8 +29,8 @@ pub const MAXIMUM_PYPY_MINOR: usize = 11;
 fn windows_interpreter_no_build(
     major: usize,
     minor: usize,
-    target_width: usize,
-    pointer_width: usize,
+    target: &Target,
+    platform: String,
     min_python_minor: usize,
     requires_python: Option<&VersionSpecifiers>,
 ) -> bool {
@@ -46,11 +46,28 @@ fn windows_interpreter_no_build(
         }
     }
 
+    // Platform can be either win32, win-amd64 or win-arm64
+    let pointer_width = if platform.ends_with("64") {
+        64
+    } else {
+        32
+    };
+
     // There can be 32-bit installations on a 64-bit machine, but we can't link
     // those for 64-bit targets
-    if pointer_width != target_width {
+    if pointer_width != target.pointer_width() {
         eprintln!(
-            "👽 {major}.{minor} is installed as {pointer_width}-bit, while the target is {target_width}-bit. Skipping."
+            "👽 {major}.{minor} is installed as {pointer_width}-bit, while the target is {target.pointer_width()}-bit. Skipping."
+        );
+        return true;
+    }
+
+    // Skip mismatching architectures
+    if (platform.contains("arm") && !target.is_aarch64())
+        || (platform.contains("amd") && !target.is_x86_64())
+    {
+        eprintln!(
+            "👽 {major}.{minor} is installed for {platform}, while the target is {target}. Skipping."
         );
         return true;
     }
@@ -101,6 +118,7 @@ fn find_all_windows(
 ) -> Result<Vec<String>> {
     let min_python_minor = bridge.minimal_python_minor_version();
     let code = "import sys; print(sys.executable or '')";
+    let platform_code = "import sysconfig; print(sysconfig.get_platform() or '')";
     let mut interpreter = vec![];
     let mut versions_found = HashSet::new();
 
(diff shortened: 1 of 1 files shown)


tonybaloney:
The conda and windows store checks also need to be updated.

The way the `windows_python_info` function scrapes the `sys.version` is brittle, using `sysconfig.get_platform()` would be more robust.

tonybaloney:
I've refactored all the methods to use `python3 -c "import sys, sysconfig; print(sysconfig.get_platform(), sys.version_info.major, sys.version_info.minor)"` which is a bit more robust for extracting the major, minor and platform details.

There are only 3 platforms in Windows (for now) and this removes some logic around AMD64.

tonybaloney:
FYI Claude wrote the test, I didn't want to manually test them and there don't seem to be any related tests in the suite right now.

src/python_interpreter/mod.rs
```diff
@@ -48,15 +49,33 @@ fn windows_interpreter_no_build(
...
+    {
+        eprintln!(
+            "👽 {major}.{minor} reports {platform}, which does not match the target architecture. Skipping."
+        );
+        return true;
+    }
+
+    // Skip mismatching architectures
+    if (platform == "win-arm64" && target.target_arch() != Arch::Aarch64)
+        || (platform == "win-amd64" && target.target_arch() != Arch::X86_64)
```
tonybaloney: NB: Check what the current default AMD64/x86_64 architecture label would be on Windows. I assume X86_64.

A safer option might be to change the second condition to 

`platform == "win-amd64" && target.target_arch() == Arch::Aarch64`

tonybaloney:
Requesting review from @davidhewitt

src/python_interpreter/mod.rs
tonybaloney: The test failures match my assumption, so I'll use this instead.

tonybaloney:
FYI @diegorusso this is blocking several wheels. Lmk if you have any comments

src/python_interpreter/mod.rs
```diff
@@ -275,23 +287,24 @@ fn windows_python_info(executable: &Path) -> Result<Option<WindowsPythonInfo>> {
...
-            major,
-            minor,
-            pointer_width: Some(pointer_width),
-        }))
-    } else {
-        Ok(None)
-    }
+
+    // Split into 3 segments: platform, major, minor by spaces
+    let segments: Vec<&str> = version_info.split_whitespace().collect();
```
davidhewitt: Is it possible that the platform will ever have whitespace? I think it might be more reliable to re-order the tuple to have major, minor, platform just in case.

We can limit to split into three parts and guarantee this:

```suggestion
    // Split into 3 segments: major, minor, platform by spaces
    let segments: Vec<&str> = version_info.splitn(3, ' ').collect();
    let [major, minor, platform] = pieces.as_slice() else {
        bail!(
            "Unexpected output for Python version info from {}: '{}'",
            executable.display(),
            version_info
        );
    };
    // can then parse each substring
```

src/python_interpreter/mod.rs
```diff
@@ -247,16 +265,10 @@ fn find_all_windows(
...
-    major: usize,
-    minor: usize,
-    pointer_width: Option<usize>,
-}
-
 fn windows_python_info(executable: &Path) -> Result<Option<WindowsPythonInfo>> {
     let python_info = Command::new(executable)
         .arg("-c")
-        .arg("import sys; print(sys.version)")
+        .arg("import sys, sysconfig; print(sysconfig.get_platform(), sys.version_info.major, sys.version_info.minor)")
```
davidhewitt: If you accept my suggestion that platform might be best at the end

```suggestion
        .arg("import sys, sysconfig; print(sys.version_info.major, sys.version_info.minor, sysconfig.get_platform())")
```

src/python_interpreter/mod.rs
```diff
@@ -48,15 +49,33 @@ fn windows_interpreter_no_build(
...
+    // Skip mismatching architectures
+    if (platform == "win-arm64" && target.target_arch() != Arch::Aarch64)
+        || (platform == "win-amd64" && target.target_arch() == Arch::Aarch64)
+    {
         eprintln!(
-            "👽 {major}.{minor} is installed as {pointer_width}-bit, while the target is {target_width}-bit. Skipping."
+            "👽 {major}.{minor} is installed for {platform}, while the target is not. Skipping."
         );
         return true;
     }
```
davidhewitt: I assume that what we really want is to check that the Python interpreter architecture matches the Rust target, so maybe we can normalise the platform to an architecture first.

Maybe something like the below:

```rust
    let python_arch = match platform.as_str() {
        "win32" => Arch::X86,
        "win-amd64" => Arch::X86_64,
        "win-arm64" => Arch::Aarch64,
        _ => {
            eprintln!("⚠️  Warning: {major}.{minor} reports unknown platform '{platform}'. This may fail to build.");
            // false => build it anyway
            return false;
        }
    };

    let target_arch = target.target_arch();

    if python_arch != target.target_arch() {
        eprintln!(
            "👽 {major}.{minor} reports a platform '{platform}' (architecture '{python_arch}'), while the Rust target is '{target_arch}'. Skipping."
        );
        return true;
    }
```

davidhewitt (commented):
Thanks! I verified this works, I have a few suggestions, once tweaked let's get this shipped and unblock builds.

tonybaloney:
@davidhewitt that's all the tests passing now.

davidhewitt (approved):
Thanks, this looks good to me!

@messense, happy if I merge and release this?

messense:
Sure!

PR #2755, opened 2025-10-05
Trim platform details on Windows detection

### LINKED ISSUE #2754: Windows platform detection logic includes line break
### Bug Description

I noticed whilst testing 1.9.5 against the maturin GitHub action that it's including a line break in the platform name:

```default
⚠️  Warning: 3.14 reports unknown platform 'win-arm64
'. This may fail to build.
```

1.9.5

3.11.0

25.0.0

None

### Does `cargo build` work?

- [ ] Yes, it works

- [ ] Yes

### Steps to Reproduce

Force maturin-action to 1.9.5 and try building multiple Python versions for Windows

tonybaloney:
Fixes #2754

### DIFF
--- a/src/python_interpreter/mod.rs
+++ b/src/python_interpreter/mod.rs
@@ -285,7 +285,7 @@ fn windows_python_info(executable: &Path) -> Result<Option<WindowsPythonInfo>> {
         }
     };
 
-    let version_info = str::from_utf8(&python_info.stdout).unwrap();
+    let version_info = str::from_utf8(&python_info.stdout).unwrap().trim();
 
     // Split into 3 segments: major, minor, platform by spaces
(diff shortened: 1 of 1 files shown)


tonybaloney:
We need to figure out how this passed tests and where a new test needs to go for this.

src/python_interpreter/mod.rs
```diff
@@ -47,7 +47,7 @@ fn windows_interpreter_no_build(
         }
     }
 
-    let python_arch = match platform.as_str() {
+    let python_arch = match platform.as_str().trim() {
```
surister, line 50: This could be simpler.

```suggestion
    let python_arch = match platform.trim() {
```

PR #2763, opened 2025-10-08
Update maximum Python version to 3.14

messense:


messense:
odd, it gets merged even when the required status is not passed...

PR #2765, opened 2025-10-08
add Cygwin port of maturin

Kreijstal:
for people who might need maturin on cygwin


messense:
preliminary Cygwin support was added in #2818.

PR #2769, opened 2025-10-11
[docs] Add full pyproject.toml example to dynamic metadata docs

ddelange:
Hi 👋 

Took me a while to figure out this exhaustive list as part of https://github.com/oconnor663/blake3-py/pull/88. 
- I didn't understand why maturin [fails](https://github.com/oconnor663/blake3-py/actions/runs/18435038379/job/52527366829#step:10:66) if I try to make `name` dynamic too. 
- Also License will be missing from the wheel even if [project] is not defined at all, which is also undocumented, see https://github.com/oconnor663/blake3-py/pull/69. edit: https://github.com/PyO3/maturin/issues/2773
```pytb
💥 maturin failed
  Caused by: pyproject.toml at /Users/runner/work/blake3-py/blake3-py/pyproject.toml is invalid
Error: The process '/Users/runner/work/_temp/a0ba2459-b23c-41be-b302-c3f2eeef0ed3/maturin' failed with exit code 1
  Caused by: TOML parse error at line 1, column 1
    at ExecState._setResult (/Users/runner/work/_actions/PyO3/maturin-action/v1.49.4/dist/index.js:1823:25)
  |
    at ExecState.CheckComplete (/Users/runner/work/_actions/PyO3/maturin-action/v1.49.4/dist/index.js:1806:18)
1 | [project]
    at ChildProcess.<anonymous> (/Users/runner/work/_actions/PyO3/maturin-action/v1.49.4/dist/index.js:1700:27)
  | ^^^^^^^^^
    at ChildProcess.emit (node:events:524:28)
missing field `name`
    at maybeClose (node:internal/child_process:1104:16)

    at Socket.<anonymous> (node:internal/child_process:456:11)
Error: The process '/Users/runner/work/_temp/a0ba2459-b23c-41be-b302-c3f2eeef0ed3/maturin' failed with exit code 1
    at Socket.emit (node:events:524:28)
    at Pipe.<anonymous> (node:net:343:12)
```

PR #2770, opened 2025-10-13
[docs] Mention that project.name is required in pyproject.toml

ddelange:
> I didn't understand why maturin [fails](https://github.com/oconnor663/blake3-py/actions/runs/18435038379/job/52527366829#step:10:66) if I try to make name dynamic too.

docs for one of the issues I mentioned in https://github.com/PyO3/maturin/pull/2769

> Also License will be missing from the wheel even if [project] is not defined at all, which is also undocumented, see https://github.com/oconnor663/blake3-py/pull/69.

this one looks like an actual bug in maturin to me, but I might be mistaken. edit: https://github.com/PyO3/maturin/issues/2773

### DIFF
--- a/guide/src/metadata.md
+++ b/guide/src/metadata.md
@@ -74,9 +74,11 @@ When the `[project]` section is not present, maturin will populate metadata from
 * `license` - From `package.license` in Cargo.toml
 * `project_url` - From various URLs in Cargo.toml (like repository, homepage, etc.)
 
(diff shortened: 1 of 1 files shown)


guide/src/metadata.md
ddelange, line 71: this was confusing to me, so I added another sentence below to clarify the relationship is between Cargo.toml and pyproject.toml, not the MEDADATA file fields mentioned here.

> To use `package.homepage` from `Cargo.toml`, add `urls` to `project.dynamic` and so forth:

PR #2776, opened 2025-10-14
Update base64 dependency from 0.21 to 0.22

musicinmybrain:
This does not appear to require any source-code changes or introduce any test regressions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -118,12 +118,6 @@ dependencies = [
 
 [[package]]
 name = "base64"
-version = "0.21.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9d297deb1925b89f2ccc13d7635fa0714f12c87adce1c75356b39ca9b7178567"
-
-[[package]]
-name = "base64"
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -1284,6 +1278,15 @@ dependencies = [
  "either",
 ]
 
+[[package]]
+name = "itertools"
+version = "0.14.0"
```
messense: unfortunately we'll have 3 versions of `itertools` now.

messense: how about only update base64 in this PR and try to update and reduce copies of itertools in another PR?

musicinmybrain: Ok, I’ll separate these two updates. I’m not sure if there are enough degrees of freedom in the dependency tree to actually remove itertools 0.12 or 0.13 from `Cargo.lock`, but I’ll give it a shot.

musicinmybrain: > how about only update base64 in this PR and try to update and reduce copies of itertools in another PR?

https://github.com/PyO3/maturin/pull/2779

PR #2781, opened 2025-10-16
Add 3.14 and 3.15 for general CI steps

clin1234:


clin1234:
?

messense:
There is no point in adding 3.15 to CI without official pyo3 support first, 3.14 was already added in #2830 .

PR #2784, opened 2025-10-17
Bump pyo3: 0.25.0 --> 0.26.0

vvsagar:


vvsagar:
The `pyo3-no-extension-module` test fails to fail now probably because of this: [https://github.com/PyO3/maturin/pull/2723](https://github.com/PyO3/maturin/pull/2723)

Should I remove or modify that test accordingly?

messense:
> The `pyo3-no-extension-module` test fails to fail now probably because of this: #2723
> 
> Should I remove or modify that test accordingly?

Keep it at 0.25.0 for now?

PR #2790, opened 2025-10-20
Bump MSRV to 1.83.0

messense:


messense:

- #2790 ◀

clippy.toml
```diff
@@ -1,4 +1,4 @@
-msrv = "1.74.0"
+msrv = "1.83.0"
```
davidhewitt: is this still needed now the `rust-version` key exists in `Cargo.toml`?

davidhewitt (approved):

clippy.toml
messense: https://github.com/rust-lang/rust-clippy/blob/973e596eb08afe75e55ab0342029ccd9b0ae80f7/README.md#specifying-the-minimum-supported-rust-version

Looks like it's not needed.

messense:
```suggestion
```

PR #2800, opened 2025-10-22
chore: clean up various logic for rust versions under MSRV

MatthijsKok:
MSRV was bumped to 1.83 in https://github.com/PyO3/maturin/pull/2790
While reading through the codebase working on other changes I encountered a few hard-to-grep logic checks for rust versions below the MSRV of this project. Thus it can be cleaned up.


messense:
MSRV was for building `maturin`, not for compiling bindings, these are still useful for old toolchains.

PR #2817, opened 2025-11-02
update cargo_metadata to 0.23.0

alexanderkjall:


cjwatson:
Looks like this requires an MSRV bump to 1.85.0 due to `cargo-platform` changes?

messense:
Could you rebase? We've bumped MSRV to 1.85.0 on main. Thanks!

messense:
Upgraded to 0.20 instead for now in #2864.

Issue #1858, opened 2023-11-17
the `multipart` crate is unmaintained


decathorpe:
The `multipart` crate (a dependency of the "upload" feature) has not been updated in over two years, the GitHub project has been [archived](https://github.com/abonander/multipart) earlier this year, and it has been officially marked as "unmaintained" in the [RustSec advisory database](https://rustsec.org/advisories/RUSTSEC-2023-0050.html) a few months ago.

As far as I can tell, almost all projects that previously used the `multipart` crate (including the `axum`, `poem`, `rocket`, and `warp` crates - so basically all popular Rust web frameworks) have migrated to `multer`, which is also listed as a maintained alternative to `multipart`.

However, I'm not sure if `multer` provides all the functionality that was available in `multipart` - notably, client-side support seems to be missing. It appears that the `multipart` crate is only needed as a dependency because `ureq` does not have built-in multipart support? In that case, switching to a different HTTP client library (like `reqwest`) that has multipart support built-in might be an alternative.

messense:
At the moment I don't want to introduce `tokio`, `futures` and lots of other dependencies to `maturin`(through `reqwest`), I'd suggest asking `ureq` to add multipart support.

lazka:
There seems to be some recent ureq progress in https://github.com/algesten/ureq/issues/698, though nothing merged yet.

(drive by comment, as this is currently a blocker for the cygwin build, among other things..)

edit 2026: cygwin builds fine now, by disabling the upload feature, so this is no longer a blocker there

konstin:
We'll be removing upload support in maturin 2.0, which will remove this crate too: https://github.com/PyO3/maturin/pull/2875

PR #2819, opened 2025-11-02
Add Cygwin support

### LINKED ISSUE #2818: Cygwin Support
Hey,

I'd like to add Cygwin support to maturin. There is #2765 (thanks @Kreijstal) but I'll try to split it up into smaller chunks and only focus on a minimal maturin build (without upload etc).

We currently have a working downstream build with patches, updates to dependencies, and patches to dependencies (which are out of scope here): https://github.com/msys2/MSYS2-packages/tree/master/maturin

This is a tracking issue.

edit: looks like all dependencies that needed updating for cygwin have been updated in main already since the last release, so nothing to do there :)

lazka:
Add Cygwin to all OS logic and bump `target-lexicon` dependency to 0.13.3 for Cygwin support.

In `fun_with_abiflags()` add a special case for cygwin, since `platform.system()` returns something like `"CYGWIN_NT-10.0-26200"` there, and not just `"Cygwin"` alone like with other systems.

In `get_platform_tag()` default to something like `"cygwin-x86_64"`, which doesn't match `sysconfig.get_platform()` in upstream Cygwin currently, but is proposed here to make pip happy:
https://cygwin.com/pipermail/cygwin-apps/2025-October/044659.html

In `write_bindings_module()` use `".abi3.dll"` for abi3 builds, as `importlib.machinery.EXTENSION_SUFFIXES` looks like
`['.cpython-39-x86_64-cygwin.dll', '.abi3.dll', '.dll']` on Cygwin.

With this, one gets a working maturin when building with `--no-default-features --features cli-completion,scaffolding`. The other features pull in abandoned crates that don't work on Cygwin, like https://github.com/PyO3/maturin/issues/1858

Fixes #2818

Co-authored-by: @Kreijstal

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,7 +71,7 @@ fat-macho = { version = "0.4.8", default-features = false }
 once_cell = "1.7.2"
 rustc_version = "0.4.0"
 semver = "1.0.22"
-target-lexicon = "0.13.0"
+target-lexicon = "0.13.3"
 indexmap = "2.2.3"
 pyproject-toml = { version = "0.13.5", features = ["pep639-glob"] }
 python-pkginfo = "0.6.6"
--- a/src/build_context.rs
+++ b/src/build_context.rs
@@ -662,6 +662,14 @@ impl BuildContext {
             (Os::Wasi, Arch::Wasm32) => {
                 "any".to_string()
             }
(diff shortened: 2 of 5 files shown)


lazka:
Thanks!

PR #2821, opened 2025-11-03
PyO3: do not add `extension-module` feature in template and tutorial

Tpt:
Similar change as https://github.com/PyO3/pyo3/pull/5588

### DIFF
--- a/guide/src/tutorial.md
+++ b/guide/src/tutorial.md
@@ -49,6 +49,13 @@ requires = ["maturin>=1.0,<2.0"]
 build-backend = "maturin"
 ```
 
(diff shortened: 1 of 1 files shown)


guide/src/distribution.md
```diff
@@ -230,7 +230,7 @@ or providing any Windows Python library files.
 
 ```toml
 [dependencies]
-pyo3 = { version = "0.27.0", features = ["extension-module", "generate-import-lib"] }
```
messense, line 233: IMO we should still point out `extension-module` feature is required for `pyo3` < 0.27.

Tpt: I have added it back to the tutorial: 5a2a40f454f52bfbf646a22fb5100c2e359f1c9e Not sure if it's the best place though

PR #2823, opened 2025-11-04
Ensure parent directory exists before writing a file

e-nomem:
I ran into a bug when I moved my `LICENSE` file from the root of the repository to `LICENSES/MIT.txt` and updated the `license-file` directive in my `pyproject.toml` to match. `uv build --sdist .` and `uv build --wheel .` both had no issues, but running `uv lock -U` reported an error.

I tracked the issue down to the fact that both `WheelWriter` and `SdistWriter` don't care about adding directories and just stub out their implementation of `add_directory()`, but `PathWriter` does, and it needs the parent directory to exist before the file is written.

In this PR I ensured that all non-test code that calls `add_bytes_with_permissions()` calls `add_directory()` first. I considered giving `add_bytes_with_permissions()` a default impl that calls `add_directory()` and adding a layer of indirection to the trait but decided against it.

Alternately, given that 2 of the 3 main implementations of `ModuleWriter` don't care about directories, maybe `add_directory()` shouldn't be part of the trait and should be an internal implementation detail for the `PathWriter` implementation of `add_bytes_with_permissions()`? (See https://github.com/PyO3/maturin/pull/2824 for a take at this)

### DIFF
--- a/tests/common/metadata.rs
+++ b/tests/common/metadata.rs
@@ -1,18 +1,27 @@
+use anyhow::Context;
 use anyhow::Result;
 use insta::assert_snapshot;
+use itertools::Itertools;
 use maturin::{write_dist_info, BuildOptions, CargoOptions, ModuleWriter};
 use std::collections::HashMap;
-use std::path::{Path, PathBuf};
+use std::collections::HashSet;
+use std::path::{Path, PathBuf, MAIN_SEPARATOR};
 
 #[derive(Default)]
 struct MockWriter {
+    directories: HashSet<String>,
     files: Vec<String>,
     contents: HashMap<String, String>,
 }
 
 impl ModuleWriter for MockWriter {
     fn add_directory(&mut self, path: impl AsRef<Path>) -> Result<()> {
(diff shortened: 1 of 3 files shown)


tests/common/metadata.rs
```diff
@@ -58,13 +73,16 @@ fn metadata_hello_world_pep639() {
...
     hello_world-0.1.0.dist-info/METADATA
     hello_world-0.1.0.dist-info/WHEEL
-    hello_world-0.1.0.dist-info/licenses
     hello_world-0.1.0.dist-info/licenses/LICENSE
     hello_world-0.1.0.dist-info/licenses/licenses/AUTHORS.txt
     ");
+    assert_snapshot!(writer.directories.iter().sorted().join("\n"), @r"
+    hello_world-0.1.0.dist-info
+    hello_world-0.1.0.dist-info/licenses
+    hello_world-0.1.0.dist-info/licenses/licenses
```
e-nomem, line 86: This extra level of `licenses` nesting being missing from the previous snapshot output is the exact issue I had

PR #2824, opened 2025-11-04
Remove add_directory() from ModuleWriter trait

e-nomem:
In https://github.com/PyO3/maturin/pull/2823 I considered the idea of removing `add_directory()` entirely from the `ModuleWriter` trait since it seemed error-prone to require the programmer to ensure it was called everywhere. I decided to make a PR for that as well to see what it would look like.


messense:
Thanks!

Issue #2810, opened 2025-10-30
Cross compling error for Android


ririv:
### Bug Description

I discovered that using Maturin for cross-compilation on Android resulted in errors, as I sought help regarding issue #2803.

I carefully analyzed the cause of the problem and successfully identified the issue.

During cross-compilation, Maturin first reads the `PYO3_CROSS_LIB_DIR` environment variable.

`PYO3_CROSS_LIB_DIR` is used to locate the sysconfigdata file. Once found, the file path is passed to the MATURIN_PYTHON_SYSCONFIGDATA_DIR environment variable

https://github.com/PyO3/maturin/blob/d51f86d90946e5b28fe84a9495e5b20a7588a1ae/src/build_options.rs#L262-L284

However, `MATURIN_PYTHON_SYSCONFIGDATA_DIR` passes this value to `PYO3_CROSS_LIB_DIR` for Cargo to use.

https://github.com/PyO3/maturin/blob/d51f86d90946e5b28fe84a9495e5b20a7588a1ae/src/compile.rs#L457-L459

At this point, the value of `PYO3_CROSS_LIB_DIR` has changed.

Before：
`PYO3_CROSS_LIB_DIR=python-3.13.9-aarch64-linux-android/prefix/lib`

After
`PYO3_CROSS_LIB_DIR=python-3.13.9-aarch64-linux-android/prefix/lib/python3.13`
`MATURIN_PYTHON_SYSCONFIGDATA_DIR=python-3.13.9-aarch64-linux-android/prefix/lib/python3.13`

This behavior occurs regardless of whether you use a self-built Python 3.13 Android embedded package or the official Python 3.14 Android embedded package provided by python.org, as their directory structures are identical.

The problem arises: rustc cannot find the desired lpython3.13.so file. 

Below is the actual CI scenario and corresponding log:

https://github.com/ririv/android-wheels/actions/runs/18934546393/job/54058135209

And this is part of log
```
   Compiling rpds-py v0.28.0 (/home/runner/work/android-wheels/android-wheels/library-source)
[...]

Issue #2803, opened 2025-10-24
Android Support


ririv:
In [PEP 738](https://peps.python.org/pep-0738/), Python has provided official support since version 3.13. Starting with version 3.14, [python.org](https://www.python.org/downloads/android/) began offering an Android embeddable package.
I would like to know if there are any plans to support Android in the near future.

messense:
I don't use Android so no plan from me, unless someone requires Android support contributes patches. So PR welcome if you are interested.

PR #2825, opened 2025-11-05
Update environment variables for Android cross-compilation support

ririv:
Fix https://github.com/PyO3/maturin/issues/2810
Enhancement https://github.com/PyO3/maturin/issues/2803

In fact, to successfully compile Python extension modules written with PYO3 on Android,

we must not only set `PYO3_CROSS_LIB_DIR` (refer to  [PEP738](https://peps.python.org/pep-0738/#linkage)),

but also configure environment variables for the Android NDK and related utilities like rustup and cargo.
- `rustup target add aarch64-linux-android`
- For Cargo, configure the `linker` variable to point to the Android NDK's clang.
- Additional environment variables like `cc` may also need configuration.

However, this is not an inherent issue with Maturin itself. These environment variables can be configured externally, and Cargo within Maturin will read them.

This fix ensures that after properly configuring the relevant environment variables, we can successfully cross-compile to an Android-targeted wheel.

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -454,7 +454,6 @@ fn cargo_build_command(
         }
     }
 
-
     // Set default macOS deployment target version for non-editable builds
     if !context.editable && target.is_macos() && env::var_os("MACOSX_DEPLOYMENT_TARGET").is_none() {
         use crate::build_context::rustc_macosx_target_version;


src/compile.rs
```diff
@@ -454,7 +454,7 @@ fn cargo_build_command(
         }
     }
 
-    if let Some(lib_dir) = env::var_os("MATURIN_PYTHON_SYSCONFIGDATA_DIR") {
+    if let Some(lib_dir) = env::var_os("PYO3_CROSS_LIB_DIR") {
         build_command.env("PYO3_CROSS_LIB_DIR", lib_dir);
     }
```
messense:
```suggestion
```

I think we can get rid of this and https://github.com/PyO3/maturin/blob/a2d38e1d067fccf1b6e1b6aae1b0740dc9fc84f4/src/build_options.rs#L284-L289

src/python_interpreter/config.rs
```diff
@@ -454,6 +454,10 @@ suppress_build_script_link_lines=false"#,
         if let Some(pointer_width) = self.pointer_width {
             write!(content, "\npointer_width={pointer_width}").unwrap();
         }
+        if let Ok(lib_dir) = std::env::var("PYO3_CROSS_LIB_DIR") {
+            write!(content, "\nlib_dir={}", lib_dir).unwrap();
+            write!(content, "\nlibs=python{}.{}", self.major, self.minor).unwrap();
```
messense: there is no `libs` config and should not just set to `pythonX.Y`, it depends https://github.com/PyO3/pyo3/blob/afefcced7e6e90ddb854bd94aadd27241452bafb/pyo3-build-config/src/impl_.rs#L1717-L1815

just let pyo3-build-config to fill it.

```suggestion
```

ririv: Thanks, `libs` config should be `lib_name`  see https://docs.rs/pyo3-build-config/latest/pyo3_build_config/struct.InterpreterConfig.html#structfield.lib_name

After testing, I found that setting only the `lib_dir` and `PYO3_CROSS_LIB_DIR` environment variables is sufficient to successfully link to `libpythonX.Y.so`.

src/compile.rs
ririv: `PYO3_CROSS_LIB_DIR` and `MATURIN_PYTHON_SYSCONFIGDATA_DIR` actually differ by one parent directory. However, I've found that even setting `PYO3_CROSS_LIB_DIR` correctly here is irrelevant, since the correct `lib_dir` is already properly set in the generated pyo3 config file. Therefore, we can remove it here.

PR #2826, opened 2025-11-06
add `editable-profile` option

davidhewitt:
Implements the suggesed `editable-profile` option in https://github.com/PyO3/maturin/issues/2590#issuecomment-3421007471

The tests introduced seem to be the first Rust tests exercising the pep517 pathway, so I added some machinery for that.

### DIFF
--- a/tests/common/pep517.rs
+++ b/tests/common/pep517.rs
@@ -24,6 +24,20 @@ pub fn test_pep517(
     // Ensure the test doesn't wrongly pass
     check_installed(package, &python).unwrap_err();
 
+    // Install `tomli` into the virtualenv (runtime dependency of `maturin`'s pep517 builds for
+    // Python <3.11)
+    let mut cmd = Command::new(&python);
+    cmd.args(["-m", "pip", "install", "tomli"]);
+    let output = cmd.output()?;
+    if !output.status.success() {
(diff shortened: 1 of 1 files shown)


maturin/__init__.py
```diff
@@ -58,6 +58,22 @@ def _get_sys_executable() -> str:
...
+    """
+
+    # e.g. with `pip install --no-build-isolation <package>` seems like the
+    # path may not contain the maturin from that "pip" installation, so we
+    # try to find it next to sys.executable
+    installed_maturin = os.path.join(os.path.dirname(sys.executable), "maturin")
+    if os.path.exists(installed_maturin):
+        return installed_maturin
+
+    return "maturin"
```
davidhewitt: These new tests appeared to be the first tests of the pep517 build (did I miss something?).

I hit this case when using `--no-build-isolation` to use a locally-built maturin when testing pep517. Might be worth splitting this out into a separate PR (maybe with the `pep517_default_profile` test?).

davidhewitt: NVM, I think I found a better way to build with `--no-build-isolation` that doesn't require changes here.

davidhewitt:
🎉 CI now looks green and I think this is good for review!

PR #2828, opened 2025-11-07
Add support for iOS cross-platform virtual environments

freakboy3742:
Tools such as `cibuildwheel` and [`xbuild`](https://pypi.org/project/xbuild/) manage building wheels for iOS by using a cross-platform virtual environment - a valid virtual environment for the build platform that can "pretend" to be the target/host platform when asked for sysconfigdata. This is done by monkeypatching key values in the sys, sysconfig, platform and os module when the interpreter starts. Many of these environments also set `sys.cross_compiling = True` as an indicator that they aren't "real" environments.

Maturin's concept of "cross-compiling", however, has more to do with whether the interpreter that is running the Maturin build can be used to interrogate details about the Python install for the purpose of configuring a build. To that end, somewhat counterintuitively, a cross-platform venv *isn't* "cross platform" from the perspective of Maturin.

This PR makes 3 changes:
1. It marks iOS builds as *not* being cross-platform. This ensures that the sysconfigdata values from the interpreter running maturin are passed to the build
2. When the build has not been given an explicit `--target`, or `--interpreter`, the build interpreter will be interrogated to see if it is a cross-compilation environment - and if it is, the target implied by the platform is used for the build. 
3. When an explicit `--interpreter` *has* been provided, the target will be modified if the environment is a cross-platform environment.

When combined with #2827, this means it is possible to create an iOS wheel with no more than `maturin build`, provided you're in an iOS cross-platform venv (such as those created by `cibuildwheel` and `xbuild`).

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * Allow iOS cross-platform virtual environments, such as those used by cibuildwheel, to imply an iOS target.
+* Fix iOS wheel naming to be compliant with PEP 730.
 * Fix generated WHEEL Tag metadata to be spec compliant.
 * Remove `add_directory()` from ModuleWriter and make it an implementation detail for the specific impl
 
--- a/src/build_context.rs
+++ b/src/build_context.rs
@@ -620,6 +620,21 @@ impl BuildContext {
                     format!("macosx_{x86_64_tag}_x86_64")
                 }
             }
+            // iOS (simulator and device)
+            (Os::Ios, Arch::X86_64) | (Os::Ios, Arch::Aarch64) => {
+                let arch = if target.target_arch() == Arch::Aarch64 {
+                    "arm64"
+                } else {
+                    "x86_64"
+                };
+                let abi = if self.target.target_triple().ends_with("-sim") {
+                    "iphonesimulator"
+                } else {
+                    "iphoneos"
+                };
+                let (min_sdk_major, min_sdk_minor) = iphoneos_deployment_target(env::var("IPHONEOS_DEPLOYMENT_TARGET").ok().as_deref())?;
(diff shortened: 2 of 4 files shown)


messense:
Thanks for the PR, having to use cibuildwheel or xbuild to compile iOS wheel is probably fine, but I just want to ask if it's possible to do it without them by only using `maturin build` and a working Xcode installation?

freakboy3742:
> Thanks for the PR, having to use cibuildwheel or xbuild to compile iOS wheel is probably fine, but I just want to ask if it's possible to do it without them by only using `maturin build` and a working Xcode installation?

Yes, a build is possible without using cibuildwheel or xbuild; however, you also need to:
1. Explicitly specify the target (i.e., `maturin build --target aarch64-apple-ios`)
2. Configure your project so that it manually adds in the linking flags that are needed to ensure the iOS Python framework is linked into the module.

The benefit of the cross environment approach is that both these details are managed for you.

(FYI - I've made one additional change from the original PR, so the the cross-platform target discovery is also performed when an interpreter is explicitly provided; `python -m build` does this, so it seemed worth including).

messense:
> * Explicitly specify the target (i.e., `maturin build --target aarch64-apple-ios`)

I don't see this as an issue, we always pass it in github actions (maturin-action) when cross-compiling

> * Configure your project so that it manually adds in the linking flags that are needed to ensure the iOS Python framework is linked into the module.

`maturin build` could learn to add the required framework linking flags as long as we can query it from Xcode?

freakboy3742:
> `maturin build` could learn to add the required framework linking flags as long as we can query it from Xcode?

That's the problem though - it's not something Xcode knows about. You need to know the location where the iOS version of Python is installed, which is specific to a given user's install, and the version that they have installed.

PR #2830, opened 2025-11-07
ci: bump to Python 3.14, update runners

davidhewitt:
This bumps the test matrix:
- to use Python 3.14
- to use newer macos GitHub runners

It also fixes an issue with the free-threaded tests where the system `python3` (a 3.10 install) seems to be picked up, see https://github.com/PyO3/maturin/actions/runs/19167952266/job/54793087726?pr=2826#step:19:1242

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,8 +53,10 @@ jobs:
           # TODO: Tests are getting stuck
           - os: ubuntu-latest
             python-version: pypy3.11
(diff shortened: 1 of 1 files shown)


test-crates/pyo3-ffi-pure/src/lib.rs
davidhewitt, line 1: This test needed updating from an _ancient_ pyo3-ffi version 😂

messense (approved):

PR #2831, opened 2025-11-07
Make sdist reproducible

e-nomem:
I compared an sdist uploaded to pypi to one built from source at the same tag and found two cases of metadata making the build not reproducible:
1. The uid/gid data is not cleared out (for both reproducibility and security, these should be 0/root)
2. The file mtime seems to be the build timestamp instead of `SOURCE_DATE_EPOCH` or some other fixed value

This PR combines 2 changes that both, individually, can address both issues above:
1. Ensure that we use `HeaderMode::Deterministic` when creating the sdist tar. This ensures that all files added using `add_file()` have uid/gid set to 0 and an mtime of July 23, 2006
2. Remove the override of the `add_file()` method in the `SdistWriter` implementation of `ModuleWriter`. This forces all writes to go through `add_bytes_with_permissions()`, which never had an issue with the uid/gid and, as of this PR, also sets the mtime to the static value from the tar crate.

This means the fix 1 isn't strictly required as long as no one writes to the archive without going through `add_bytes_with_permissions()` but, if they do, those files should look identical to everything else in the archive.

~~This PR does not honor `SOURCE_DATE_EPOCH` if it is set in the build env, but it probably would not be a lot of work to add it.~~

### DIFF
--- a/src/module_writer.rs
+++ b/src/module_writer.rs
@@ -433,6 +433,7 @@ pub struct SDistWriter {
     path: PathBuf,
     file_tracker: FileTracker,
     excludes: Override,
+    mtime: u64,
 }
 
 impl ModuleWriter for SDistWriter {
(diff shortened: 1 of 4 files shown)


messense:
> This PR does not honor `SOURCE_DATE_EPOCH` if it is set in the build env, but it probably would not be a lot of work to add it.

We already have honored this when building wheels so IMO it's better to also honor it here too.

e-nomem:
@messense I've updated the PR to check the `SOURCE_DATE_EPOCH` env var and added a few tests

PR #2833, opened 2025-11-08
Upgrade goblin to 0.10

messense:


messense:
Requires a MSRV bump to 1.85.0.

MatthijsKok:
> Requires a MSRV bump to 1.85.0.

@messense Does this mean that the `rust-version` in `Cargo.toml` could be bumped to `1.85`?
That would imply that we could also upgrade to `edition = "2024"` :D
I'd like to pick that up if that's the case.

Potentially I would still like to also include parts of https://github.com/PyO3/maturin/pull/2800
I'm still not 100% confident on which parts of that PR are used for compiling bindings.
Can the code in the tests at least be upgraded? 

Appreciate your time :)

PR #2839, opened 2025-11-09
add test case for uniffi with multiple crates

### LINKED ISSUE #1904: Uniffi multiple crates not supported

davidhewitt:
Closes #1904 

Looks like the issue was fixed by #2208 and #2305 - pushing this test as a draft to clean up another day.


messense:
Anything else missing for this to be merged?

davidhewitt:
I think this might need adding to actual test cases in the suite? I might have just run it by hand.

PR #2840, opened 2025-11-09
always use "library" mode to generate uniffi bindings

davidhewitt:
Closes #2617
Closes #2618 

My take is that because we compile a cdylib we should _always_ be using that cdylib to generate the uniffi bindings.

cc @JoFrost @badboy @messense would be interested in your views on this approach. Seems to "just work", reduces complexity, and avoids adding an option.

PR #2841, opened 2025-11-09
release: 1.10.0

davidhewitt:
@messense would be awesome if we could ship this soon so that I can use #2826 without needing a git dependency on `maturin`.

Maybe to merge #2840 first?


messense (approved):

davidhewitt:
Thanks for the approve, I'll run the release process now.

PR #2842, opened 2025-11-10
Split out convenience methods from ModuleWriter trait

e-nomem:
I was thinking that the convenience methods with default implementations available on `ModuleWriter` shouldn't really be overridden and that implementers of `ModuleWriter` should only focus on the one primary method. In order to enforce that separation, this PR moves all the convenience methods into a `ModuleWriterExt` trait that has a blanket implementation for all `ModuleWriter`s.

In addition to that, the following related changes are included:
 - Change the main method to be implemented on `ModuleWriter` to accept an `impl Read` instead of a byte slice
 - Related to using an `impl Read`, add a `StreamSha256` struct that calculates the hash as the data is streamed to the underlying Writer
 - Limit the permissions for the files to executable or non-executable (`0o755` or `0o644`)

The PR is structured so that the individual commits can be reviewed to ensure that there are no behavioral changes.

### DIFF
--- a/src/binding_generator/cffi_binding.rs
+++ b/src/binding_generator/cffi_binding.rs
@@ -89,13 +89,13 @@ pub fn write_cffi_module(
     };
 
     if !editable || project_layout.python_module.is_none() {
-        writer.add_data(
+        writer.add_bytes(
             module.join("__init__.py"),
             None,
             cffi_init_file(&cffi_module_file_name).as_bytes(),
             false,
         )?;
-        writer.add_data(
+        writer.add_bytes(
             module.join("ffi.py"),
             None,
             cffi_declarations.as_bytes(),
--- a/src/binding_generator/pyo3_binding.rs
+++ b/src/binding_generator/pyo3_binding.rs
@@ -126,7 +126,7 @@ pub fn write_bindings_module(
     } else {
         let module = PathBuf::from(ext_name);
         // Reexport the shared library as if it were the top level module
-        writer.add_data(
+        writer.add_bytes(
             module.join("__init__.py"),
             None,
             format!(
--- a/src/binding_generator/uniffi_binding.rs
+++ b/src/binding_generator/uniffi_binding.rs
@@ -248,7 +248,7 @@ pub fn write_uniffi_module(
     };
 
     if !editable || project_layout.python_module.is_none() {
-        writer.add_data(module.join("__init__.py"), None, py_init.as_bytes(), false)?;
+        writer.add_bytes(module.join("__init__.py"), None, py_init.as_bytes(), false)?;
         for binding in binding_names.iter() {
             writer.add_file(
                 module.join(binding).with_extension("py"),
--- a/src/binding_generator/wasm_binding.rs
+++ b/src/binding_generator/wasm_binding.rs
@@ -54,6 +54,6 @@ if __name__ == '__main__':
     let launcher_path = Path::new(&metadata.get_distribution_escaped())
         .join(bin_name.replace('-', "_"))
         .with_extension("py");
-    writer.add_data(&launcher_path, None, entrypoint_script.as_bytes(), true)?;
+    writer.add_bytes(&launcher_path, None, entrypoint_script.as_bytes(), true)?;
     Ok(())
 }
--- a/src/module_writer/mod.rs
+++ b/src/module_writer/mod.rs
@@ -35,7 +35,7 @@ pub trait ModuleWriter {
     /// the appropriate unix permissions
     ///
     /// For generated files, `source` is `None`.
-    fn add_data(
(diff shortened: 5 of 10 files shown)


src/module_writer.rs
```diff
@@ -1508,25 +1580,16 @@ pub fn add_data(
...
                     if file.path_is_symlink() {
                         // Copy the actual file contents, not the link, so that you can create a
                         // data directory by joining different data sources
-                        let source = fs::read_link(file.path())?;
-                        writer.add_file_with_permissions(
-                            relative,
-                            source.parent().unwrap(),
-                            mode,
-                        )?;
+                        writer.add_file_infer_permissions(relative, file.path())?;
```
e-nomem: This is the only change I'm uncertain about. I'm passing `file.path()` here as the source so that the permission can be checked, but previously the code used to pass in the parent directory as the source. It's not clear to me why.

src/module_writer.rs
```diff
@@ -460,18 +468,25 @@ impl ModuleWriter for SDistWriter {
...
+            .with_context(|| format!("Failed to read data into buffer for {}", target.display()))?;
+
         let mut header = tar::Header::new_gnu();
-        header.set_size(bytes.len() as u64);
-        header.set_mode(permissions);
+        header.set_size(buffer.len() as u64);
+        header.set_mode(default_permission(executable));
         header.set_mtime(self.mtime);
-        header.set_cksum();
+
```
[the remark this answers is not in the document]
e-nomem: This is not necessary. `tar.append_data()` takes a `&mut header` because it does internally mutate the header to write the path into it, so the checksum will be recalculated internally within `append_data()` anyway.

src/module_writer.rs
```diff
@@ -50,59 +50,66 @@ const SDIST_DETERMINISTIC_TIMESTAMP: u64 = 1153704088;
...
+        source: impl AsRef<Path>,
+    ) -> Result<()> {
+        let source = source.as_ref();
+
+        #[cfg(unix)]
+        let mode = source.metadata()?.permissions().mode();
+        #[cfg(not(unix))]
+        let mode = 0o644;
+
+        let executable = (0o100 & mode) == 0o100;
```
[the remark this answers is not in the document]
e-nomem: Only the user execute bit should be propagated IMO. This also matches the behavior of the tar crate when `HeaderMode::Deterministic` is used

src/module_writer.rs
```diff
@@ -577,6 +592,62 @@ impl FileTracker {
...
+#[inline]
+fn default_permission(executable: bool) -> u32 {
+    match executable {
+        true => 0o755,
+        false => 0o644,
+    }
+}
+
+#[cfg(not(unix))]
+#[inline]
```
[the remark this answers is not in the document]
e-nomem: I had to think about this one for a while. I think defaulting to `0o644` for `default_permission()` on Windows is actually a change in behavior for the cases where an executable permission was explicitly requested. E.g. when the dylib is written to the zip or tarball, we previously used to set the executable bit, but we would no longer do so if the zip or tarball was built on a Windows host. I've removed the Windows-specific version of `default_permission()` so the behavior remains consistent.

src/module_writer.rs
```diff
@@ -48,59 +48,66 @@ const SDIST_DETERMINISTIC_TIMESTAMP: u64 = 1153704088;
 
 /// Allows writing the module to a wheel or add it directly to the virtualenv
 pub trait ModuleWriter {
-    /// Adds a file with bytes as content in target relative to the module base path.
+    /// Adds a file with data as content in target relative to the module base path while setting
+    /// the appropriate unix permissions
     ///
     /// For generated files, `source` is `None`.
-    fn add_bytes(
+    fn add_data(
```
messense: I'd prefer we keep the name `add_bytes` since `add_data` could mean adding file to [`.data` directory](https://packaging.python.org/en/latest/specifications/binary-distribution-format/#wheel-defines-a-data-directory-should-i-put-all-my-data-there)

src/module_writer.rs
messense: I'm not sure about this either. cc @konstin

src/module_writer.rs
e-nomem: Done

e-nomem:
@messense I can back out the change adding `add_file_infer_permissions()` so that we don't change the symlink behavior.  Does the rest of the PR look good to you?

messense:
> @messense I can back out the change adding `add_file_infer_permissions()` so that we don't change the symlink behavior. Does the rest of the PR look good to you?

Sounds good to me.

e-nomem:
The PR has been updated to remove `add_file_infer_permissions()`

messense:
Thanks!

PR #2849, opened 2025-11-12
Add linux sw_64 architecture support

Verney7:
In this commit, support for a new architecture sw_64 is added to maturin. sw_64 is a new RISC ISA, which is a bit like RISC-V. Though it is a rather new ISA, it has been added to openEuler and openAnolis port since 2020. So supporting this new architecture is meaningful. 
It's PR is to solve compilation failure on sw_64 architecture.

Binutils has identified EM_SW64 is 268, I hope maturin can accept this patch. Thanks.

https://gabi.xinuos.com/elf/a-emachine.html

https://en.wikipedia.org/wiki/Sunway_(processor)

### DIFF
--- a/src/target/mod.rs
+++ b/src/target/mod.rs
@@ -97,7 +97,7 @@ pub enum Arch {
     Sparc64,
     Sparcv9,
     LoongArch64,
-    Sw_64,
+    Sw64,
 }
 
 impl fmt::Display for Arch {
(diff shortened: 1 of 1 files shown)


messense:
I don't think it make sense to add this when rustc has no support for it: https://doc.rust-lang.org/nightly/rustc/platform-support.html

messense:
As you can see the build is failed, thanks for the interest in adding the support anyway. But we can not move forward with this without rustc supporting the Linux sw_64 target.

Verney7:
OK. thanks.

Issue #2334, opened 2024-11-29
Implement PEP 740 in `maturin upload` or drop the `upload` command in favor of `twine`/`uv`


messense:
https://docs.pypi.org/attestations/

IMHO implement it in `maturin` feels like duplicate efforts, what do you think @konstin

konstin:
`uv publish` has grown much better in every dimension: More features, better UX, more standards compliant. Remove `maturin upload` would remove a lot of dependencies. Once `uv publish` moves out of preview, we can deprecate `maturin upload` and eventually remove it in maturin 2.0

davidhewitt:
Nearly 1 year later, shall we move forward with this?

konstin:
Sounds good

airvzxf:
Hi, I'd like to work on this for Hacktoberfest!

messense:
Feel free to work on this, I think there are 3 things we want

1. print a deprecated message in `upload` command
2. change `generate-ci` to use `uv` to upload sdist and wheels to pypi
3. a separate PR to remove the `upload` command entirely for maturin 2.0

konstin:
We should do a new release to get the warning out.

Here's some baseline data from bigquery:

```sql
SELECT COUNT(*), DATE_TRUNC(DATE(upload_time), WEEK) AS week,
FROM `bigquery-public-data.pypi.distribution_metadata`
WHERE uploaded_via LIKE "maturin%" AND DATE(upload_time) >= DATE("2025-09-01")
GROUP BY week
ORDER BY week DESC
```

The ends aren't normalized to full weeks and there's of course the christmas break:

<img width="911" height="585" alt="Image" src="https://github.com/user-attachments/assets/b3f47a77-bea2-4cdb-990a-82de9f0e2280" />

jeteve:
Is the upgrade documented somewhere for maintainers who have already generated their CIs in the past?

PR #2875, opened 2025-11-26
Deprecate 'upload' and 'publish' CLI commands

e-nomem:
Implements half of #2334

### DIFF
--- a/src/ci.rs
+++ b/src/ci.rs
@@ -659,14 +659,17 @@ jobs:\n",
             );
         }
         conf.push_str(
+            r#"      - name: Install uv
+        if: ${{ startsWith(github.ref, 'refs/tags/') }}
+        uses: astral-sh/setup-uv@v7
+"#,
+        );
+        conf.push_str(
             r#"      - name: Publish to PyPI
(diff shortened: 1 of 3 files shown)


src/ci.rs
```diff
@@ -658,15 +658,18 @@ jobs:\n",
...
             r#"      - name: Publish to PyPI
         if: ${{ startsWith(github.ref, 'refs/tags/') }}
-        uses: PyO3/maturin-action@v1
+        run: uv publish 'wheels-*/*'
         env:
-          MATURIN_PYPI_TOKEN: ${{ secrets.PYPI_API_TOKEN }}
-        with:
-          command: upload
-          args: --non-interactive --skip-existing wheels-*/*
+          UV_PUBLISH_TOKEN: ${{ secrets.PYPI_API_TOKEN }}
```
e-nomem, line 672: IMO we should suggest secure defaults and add `--trusted-publishing always` to the default command here, removing the `UV_PUBLISH_TOKEN` env var

messense: Maybe in next version so this version is a smooth upgrade

PR #2887, opened 2025-12-03
Add `VirtualWriter` to track and order archive entries

e-nomem:
As I mentioned previously in https://github.com/PyO3/maturin/pull/2882, this PR adds a `VirtualWriter` that wraps around the existing module writer structs and extracts the common file tracking and exclusion logic. In addition, it tracks all files before writing them at the end allowing us to write the files with a consistent order.

After this PR, it should be possible to have mostly reproducible builds. For sdists this requires using the same version of maturin for reproducibility, and for wheels it requires using the same versions of maturin, the rust toolchain, and external libraries.

### DIFF
--- a/src/module_writer/virtual_writer.rs
+++ b/src/module_writer/virtual_writer.rs
@@ -4,13 +4,17 @@ use std::collections::hash_map::Entry;
 use std::collections::hash_map::VacantEntry;
 use std::path::Path;
 use std::path::PathBuf;
+use std::rc::Rc;
 
 use anyhow::Result;
 use anyhow::bail;
 use ignore::overrides::Override;
 #[cfg(test)]
 use indexmap::IndexMap;
+use once_cell::unsync::OnceCell;
 use same_file::is_same_file;
+use tempfile::TempDir;
+use tempfile::tempdir;
 use tracing::debug;
 
 use crate::Metadata24;
@@ -39,6 +43,7 @@ pub struct VirtualWriter<W> {
     tracker: HashMap<PathBuf, ArchiveSource>,
     excludes: Override,
     target_exclusion_warning_emitted: bool,
+    temp_dir: OnceCell<Rc<TempDir>>,
 }
 
 impl<W: ModuleWriterInternal> VirtualWriter<W> {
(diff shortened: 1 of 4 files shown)


src/module_writer/mod.rs
```diff
@@ -264,11 +266,11 @@ pub fn add_data(
 
 /// Creates the .dist-info directory and fills it with all metadata files except RECORD
 pub fn write_dist_info(
-    writer: &mut impl ModuleWriter,
+    writer: &mut VirtualWriter<impl ModuleWriterInternal>,
```
e-nomem, line 269: Most of the function signatures for `{add,write}_<whatever>` changed to point at `VirtualWriter<SomeConcreteType>` to force the usage of the virtual writer and because they're only intended for use specifically with either `WheelWriter` or `SDistWriter`. `write_dist_info` here is the only exception because it's used with both `WheelWriter` and `PathWriter`

src/module_writer/virtual_writer.rs
```diff
@@ -0,0 +1,294 @@
...
+                        }
+                    }
+                }
+            }
+        };
+
+        Ok(entry)
+    }
+
+    pub(crate) fn add_entry(
```
e-nomem, line 149: Note that `VirtualWriter` has an `add_entry()` function but does not implement `ModuleWriterInternal` intentionally. Because `VirtualWriter` is generic over any `ModuleWriterInternal` this specific setup prevents creating nested `VirtualWriter<VirtualWriter<..>>` situations.

src/module_writer/mod.rs
```diff
@@ -315,6 +317,37 @@ pub fn write_dist_info(
...
+                    "python dir path `{}` does not exist or is invalid",
+                    project_layout.python_dir.display()
+                )
+            })?
+            .into_path_buf();
+        if let Some(python_path) = absolute_path.to_str() {
+            let name = metadata24.get_distribution_escaped();
+            let target = format!("{name}.pth");
+            debug!("Adding {} from {}", target, python_path);
+            writer.add_bytes(target, None, python_path, false)?;
```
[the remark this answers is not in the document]
e-nomem: This isn't necessary. `&str` already implements `Into<Vec<u8>>`

messense (approved):

messense:
There is a test failure on Alpine Linux. and upgrade Cirrcus CI freebsd version could fix FreeBSD test I think.

e-nomem:
The failure was caused by the TempDir in `add_external_libs()` going out of scope before the virtual writer started committing the files to the archive. I've added a TempDir to the virtual writer itself that callers can use if they need it and have updated `add_external_libs()` as well as the `CffiBindingGenerator` and `PyO3BindingGenerator` structs to use the shared temp dir.

e-nomem:
I rebased this branch onto https://github.com/PyO3/maturin/pull/2888 so that you can just merge either PR instead of both

PR #2888, opened 2025-12-03
Update FreeBSD image to 14.3

e-nomem:
FreeBSD 14.2 hit EOL yesterday, which is likely why CI started failing


e-nomem:
I rebased https://github.com/PyO3/maturin/pull/2887 onto this branch so you can just merge either PR instead of both

PR #2889, opened 2025-12-03
v2: Remove the 'upload' and 'publish' commands

### LINKED ISSUE #2334: Implement PEP 740 in `maturin upload` or drop the `upload` command in favor of `twine`/`uv`

e-nomem:
This is a breaking change intended for maturin v2

Closes #2334

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -160,12 +160,6 @@ static = ["xz2/static"]
 # Internal feature to speed up the tests significantly
 faster-tests = []
 
-# Deprecated features, keep them now for compatibility
-human-panic = []
-log = []
-password-storage = []
-upload = []
-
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -206,6 +163,8 @@ faster-tests = []
 # Deprecated features, keep them now for compatibility
 human-panic = []
 log = []
+password-storage = []
+upload = []
```
e-nomem: Should we use the v2 breaking changes to also clean these up?

messense: I think we should.

PR #2895, opened 2025-12-16
Replace rustls-pemfile with rustls-pki-types

e-nomem:
The recommended replacement for `rustls-pemfile` seems to be using `rustls-pki-types`, so I did that here

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -81,8 +81,6 @@ ignore = [
     "RUSTSEC-2024-0436",
     # num_prefix unmaintained
     "RUSTSEC-2025-0119",
-    # rustls-pemfile unmaintained
-    "RUSTSEC-2025-0134",
     #{ id = "RUSTSEC-0000-0000", reason = "you can specify a reason the advisory is ignored" },
     #"a-crate-that-is-yanked@0.1.1", # you can also ignore yanked crate versions if you wish
     #{ crate = "a-crate-that-is-yanked@0.1.1", reason = "you can specify why you are ignoring the yanked crate" },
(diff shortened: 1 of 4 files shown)


src/upload.rs
```diff
@@ -393,11 +420,9 @@ fn http_agent() -> Result<ureq::Agent, UploadError> {
...
-    if let Some(ca_bundle) = tls_ca_bundle() {
-        let mut reader = io::BufReader::new(File::open(ca_bundle)?);
-        for cert in rustls_pemfile::certs(&mut reader) {
-            let cert = cert?;
-            tls_builder.add_root_certificate(native_tls::Certificate::from_pem(&cert)?);
+    if let Some(root_certs) = ca_certificates()? {
+        for cert in root_certs {
+            tls_builder.add_root_certificate(native_tls::Certificate::from_der(&cert)?);
         }
     }
```
e-nomem, line 427: I think we should really be calling `tls_builder.disable_built_in_roots(true)` in this block. That would match the behavior on the `rustls` side where we replace the root certs instead of appending to it.

messense: I guess we can do it in v2.0 to avoid potential breakage.

deny.toml
```diff
@@ -81,8 +81,6 @@ ignore = [
     "RUSTSEC-2024-0436",
     # num_prefix unmaintained
     "RUSTSEC-2025-0119",
-    # rustls-pemfile unmaintained
-    "RUSTSEC-2025-0134",
```
messense, line 85: we need to keep this for now because of MSRV limit we can't upgrade cargo-xwin yet.

messense:
Done in `main`.

PR #2902, opened 2025-12-28
Make PEP 517 profile tests more resilient to cargo profiles

konstin:
A trick for faster rust compilation from the official book is to avoid generating debug information by default (https://doc.rust-lang.org/cargo/guide/build-performance.html#reduce-amount-of-generated-debug-information).

To apply this for globally, or to use this for projects that already have a `.cargo/config.toml`, we can add the following to `~/.cargo/config.toml`:
```toml
[profile.dev]
debug = "line-tables-only"

[profile.dev.package."*"]
debug = false

[profile.debugging]
inherits = "dev"
debug = true
```

or simply:

```toml
[profile.dev]
debug = false
```

This effect is larger for more complex projects with more dependencies, but also noticeable for maturin. On my machine, a fresh debug build goes from 14.6s to 9s, `hyperfine --prepare "touch src/archive_source.rs" "cargo build"` goes from 960ms to 870ms.

This breaks the PEP 517 profile test because the log message looks different:

```
Finished `dev` profile [unoptimized] target(s) in 8.84s
```

To make the test more resilient to all sorts of global profile changes, we instead check in the target directory which profiles were built.

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -222,13 +222,3 @@ pub fn create_conda_env(name: &str, major: usize, minor: usize) -> Result<(PathB
 pub fn test_python_path() -> Option<String> {
     env::var("MATURIN_TEST_PYTHON").ok()
 }
-
-/// Find index of subslice in a larger slice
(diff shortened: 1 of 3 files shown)


konstin:
The emscripten failure is unrelated.

PR #2903, opened 2025-12-28
Remove dtolnay/rust-toolchain GitHub Actions usage outside docker

konstin:
Rustup and stable rust are nowadays shipped on GitHub Actions, we don't need an action to install and/or manage them. This PR trims that dependency and simplifies CI. The action remains used inside custom docker containers that don't have rustup in the same way that the GitHub runner images have.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -401,7 +401,10 @@ jobs:
           name: maturin-build
           path: bin
       - run: chmod +x bin/maturin
-      - run: rustup target add ${{ matrix.platform.target }}
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets:
+            ${{ matrix.platform.target }}
       - name: Build wheels
         run: |
           set -ex


.github/workflows/release.yml
```diff
@@ -183,10 +180,12 @@ jobs:
       # powerpc64le-unknown-linux-musl doesn't have official std library release
```
konstin: Is this still true? https://doc.rust-lang.org/beta/releases.html#version-1850-2025-02-20 has promoted this target.

.github/workflows/release.yml
```diff
@@ -183,10 +180,12 @@ jobs:
...
       - run: rustup target add --toolchain stable ${{ matrix.platform.target }}
         if: ${{ !contains(fromJson('["powerpc64le-unknown-linux-musl", "s390x-unknown-linux-gnu", "loongarch64-unknown-linux-gnu", "riscv64gc-unknown-linux-gnu"]'), matrix.platform.target) }}
-      - uses: dtolnay/rust-toolchain@stable
+      - name: Install Rust stable toolchain
         if: contains(fromJson('["s390x-unknown-linux-gnu", "loongarch64-unknown-linux-gnu", "riscv64gc-unknown-linux-gnu"]'), matrix.platform.target)
-        with:
-          targets: ${{ matrix.platform.target }}
+        run: |
+          rustup toolchain install stable --profile minimal
```
konstin: This is the case I'm least sure about - Do we have a dry run for the release pipeline?

.github/workflows/release.yml
messense: No longer true.

.github/workflows/release.yml
messense: It has workflow_dispatch to manual test release pipeline.

PR #2906, opened 2025-12-29
Update pyodide version to fix emscripten CI

konstin:
Apparently a change in the Rust compiler around exception handling broke the pyodide CI (https://github.com/rust-lang/compiler-team/issues/920). We can fix this by updating pyodide in the test.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -56,7 +56,7 @@ def setup_pyodide(session: nox.Session):
                 "pyodide.asm.js",
                 external=True,
             )
-            with open("repodata.json") as f:
+            with open("pyodide-lock.json") as f:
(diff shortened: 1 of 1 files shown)


noxfile.py
```diff
@@ -56,7 +56,7 @@ def setup_pyodide(session: nox.Session):
                 "pyodide.asm.js",
                 external=True,
             )
-            with open("repodata.json") as f:
+            with open("pyodide-lock.json") as f:
```
konstin, line 59: See https://github.com/pyodide/pyodide/blob/a3b4979568a2b37223b0d345189c2c7d4607dec9/docs/project/changelog.md?plain=1#L1077-L1079

Issue #2777, opened 2025-10-15
Unreadable directories in project root cause build failures


lucascool12:
### Bug Description

When directories exist in the project root that the user is not able to read, maturin build immediately gives up.

### Your maturin version (`maturin --version`)

1.8.3

### Your Python version (`python -V`)

python 3.11

### Your pip version (`pip -V`)

25.0.1

### What bindings you're using

pyo3

### Does `cargo build` work?

- [x] Yes, it works

### If on windows, have you checked that you aren't accidentally using unix path (those with the forward slash `/`)?

- [ ] Yes

### Steps to Reproduce

For example if in any project you run the following commands:

```bash
mkdir bogus_dir
chmod 700 bogus_dir
sudo chown root bogus_dir
```

Then running `maturin build` in this project fails with

```
💥 maturin failed
  Caused by: Failed to add the files to the wheel
  Caused by: Failed to add the python module to the package
  Caused by: /$PROJECT_ROOT/bogus_dir: IO error for operation on /$PROJECT_ROOT/bogus_dir: Permission 
```

This seems to be caused by this question mark [here](https://github.com/PyO3/maturin/blob/50baca02012cf64b6a5a08efb718fc61cf372ac2/src/module_writer.rs#L1418).

I found this from a ci run with sccache were somehow someway some subdirectories in the cache are not readable for the build user.
Here is the full log if you are interested.

<details>

<summary>ci log</summary>

[0KRunning with gitlab-runner 18.4.0~pre.115.gb2218bab (b2218bab)[0;m
[0K  on blue-5.saas-linux-small-amd64.runners-manager.gitlab.com/default -AzERasQr, system ID: s_4cb09cee29e2[0;m
[0K[36;1mResolving secrets[0;m[0;m
section_start:1760535276:prepare_executor
[0K[0K[36;1mPreparing the "docker+machine" executor[0;m[0;m
[0KUsing Docker executor with image registry.gitlab.com/sli-lib/sli/ci:latest ...[0;m
[0KUsing effective pull policy of [always] for container registry.gitlab.com/sli-lib/sli/ci:latest[0;m
[0KAuthenticating with credentials from job payload (GitLab Registry)[0;m
[0KPulling docker image registry.gitlab.com/sli-lib/sli/ci:latest ...[0;m
[...]

PR #2916, opened 2026-01-04
Ignore unreadable excluded directories

konstin:
Fixes https://github.com/PyO3/maturin/issues/2777

### DIFF
--- a/tests/common/other.rs
+++ b/tests/common/other.rs
@@ -428,6 +428,11 @@ pub fn abi3_without_version() -> Result<()> {
 pub fn test_unreadable_dir() -> Result<()> {
     use std::os::unix::fs::PermissionsExt;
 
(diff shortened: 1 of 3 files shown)


tests/common/other.rs
```diff
@@ -400,3 +420,50 @@ pub fn abi3_without_version() -> Result<()> {
...
+        temp_dir.path().join("target").to_str().unwrap(),
+        "--out",
+        temp_dir.path().join("wheels").to_str().unwrap(),
+    ])?;
+
+    let build_context = options
+        .into_build_context()
+        .strip(cfg!(feature = "faster-tests"))
+        .editable(false)
+        .build()?;
```
konstin, line 467: I don't like that this requires a wheel build but the source dist path has a different behavior we have to capture separately.

PR #2922, opened 2026-01-06
Fix manylinux2014 compliance check

tuxu:
PR #2901 updated the manylinux policies, which now include `__cxa_thread_atexit_impl`. This symbol is [linked weakly](https://github.com/rust-lang/rust/blob/4fa80a5e733e2202d7ca4c203c2fdfda41cfe7dc/library/std/src/sys/thread_local/destructors/linux_like.rs#L4) by the Rust stdlib.

Rather than requiring a bump of the minimum manylinux platform tag for users, ignore weak symbols in the check. This is in line with changes to PyPA's auditwheel (https://github.com/pypa/auditwheel/pull/665).

### DIFF
--- a/src/auditwheel/audit.rs
+++ b/src/auditwheel/audit.rs
@@ -5,7 +5,7 @@ use crate::compile::BuildArtifact;
 use crate::target::Target;
 use anyhow::{Context, Result, bail};
 use fs_err::File;
-use goblin::elf::{Elf, sym::STT_FUNC};
+use goblin::elf::{Elf, sym::STB_WEAK, sym::STT_FUNC};
 use lddtree::Library;
 use once_cell::sync::Lazy;
 use regex::Regex;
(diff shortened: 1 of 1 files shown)


src/build_options.rs
```diff
@@ -749,10 +749,9 @@ impl BuildContextBuilder {
                         // Zig bundles musl 1.2
                         Some(PlatformTag::Musllinux { major: 1, minor: 2 })
                     } else {
-                        // With zig we can compile to any glibc version that we want, but
-                        // there are some black-listed libc symbols in auditwheel for manylinux <= 2.17/2024.
-                        // Thus we let auditwheel decide the compatibility tag.
-                        None
+                        // With zig we can compile to any glibc version that we want, so we pick the lowest
+                        // one supported by the rust compiler
+                        Some(target.get_minimum_manylinux_tag())
```
konstin, line 754: The comment is outdated now, but doesn't it still make sense to use auditwheel to compute the real tag?

messense: IMHO for future proof we should still let auditwheel compute the tag.

tuxu: From my understanding, the original comment and implementation was correct. At this point in the code, we need to decide on a specific manylinux tag, as this will be used in the package build to pass the right glibc version and target triple to zig. By returning `None`, we would target a later glibc version.

konstin: Oh you're right - The default gets us 2.31, with the explicit minimum version we get 2.17.

PR #2925, opened 2026-01-08
Use uv instead of sudo pip and pin release bootstrapping version

konstin:
Remove "sudo pip" style commands, they are a bad practice and we shouldn't have them in our CI.

This change relies on uv covering the same platforms as maturin in CI.

Also pin maturin==1.10.1 for cross builds for now, see https://github.com/PyO3/maturin/actions/runs/20812085659/job/59778558155 and https://github.com/PyO3/maturin/pull/2922

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -204,7 +204,9 @@ jobs:
           # Make psm compile, see https://github.com/rust-lang/stacker/issues/79
           CFLAGS_s390x_unknown_linux_gnu: "-march=z10"
         run: |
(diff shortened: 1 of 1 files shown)


reneleonhardt:
v1.11.3 specifically contains PR2922 😄
Is there more work to be done or can the pin be removed now after v1.11.4?

Interestingly enough, "fastly" doesn't seem to be that fast, the badge is still showing v1.11.3 40 minutes after the upload 😆
https://pypi.org/project/maturin/

PR #2932, opened 2026-01-11
Replace patchelf with arwen

konstin:
Avoid the external patchelf library, use arwen instead.

This creates a split situation of using both goblin and arwen for binary parsing. It does less reparsing of ELF files than previously.


konstin:
Blocked on https://github.com/gimli-rs/object/issues/842

PR #2938, opened 2026-01-13
Fix editable install for binary projects with Python modules

### LINKED ISSUE #2933: Sudden "No such file or directory" with maturin >=1.11.2
Hello,

starting from maturin version 1.11.2, we have been running into issues with "No such file or directory" problems.

Context: as part of the Magika (https://github.com/google/magika) project, we have a self-contained client written in Rust, a python `Magika` module, and a python package (`magika`) that ships both the Rust binary and the python module.

Up to version 1.10.2, we could do something like this:
```
$ git clone https://github.com/google/magika.git
$ cd magika/python
$ uv sync
$ uv run magika -V
magika 1.0.2-dev standard_v3_3
```

However, from 1.11.2, we get the following:
```
$ uv run magika -V
error: Failed to spawn: `magika`
```

We noticed as our build/test workflows, which worked fine before January 5th, started breaking with the error above.

Our temporary fix is to constraint the version of maturin to <=1.10.2. See https://github.com/google/magika/pull/1264/ for more context.

We actually don't know whether this should be considered a maturin bug... or whether we have abused some old bug so far... in which case we would appreciate some guidance on how to deal with this :-)

Note also that the new version of maturin does compile the magika binary, and it creates a file somewhere in `./python/src/magika-1.0.3.dev0.data/scripts/magika`... which is of course not in the PATH and can't be found.

Happy to provide additional details if needed; thanks for looking into this (and for the great project)!

/cc @ia0

messense:
Fixes #2933. Previously, editable installs incorrectly tried to copy binaries to the Python source directory (base_path + data/scripts/...), resulting in broken installations where the binary couldn't be found.

This introduces an ArtifactTarget enum to distinguish between Binary and ExtensionModule artifacts. In editable mode with a Python module:
- Extension modules are copied to the source directory (for immediate code changes)
- Binaries are added to the wheel archive (so pip installs them to the environment's scripts directory where they're in PATH)

### DIFF
--- a/test-crates/bin-with-python-module/check_installed/check_installed.py
+++ b/test-crates/bin-with-python-module/check_installed/check_installed.py
@@ -6,6 +6,7 @@ This tests the fix for https://github.com/PyO3/maturin/issues/2933
 import os
 import shutil
 import subprocess
+import sys
 
 
 def main():


messense:
cc @e-nomem

konstin:
Can we add a test case for this? I don't think we cover binary plus python files yet.

test-crates/bin-with-python-module/check_installed/check_installed.py
```diff
@@ -0,0 +1,37 @@
...
+def main():
+    # Test 1: Check that the Python module works
+    from bin_with_python_module import get_version
+
+    version = get_version()
+    if version != "0.1.0":
+        raise Exception(f"Expected version '0.1.0', got '{version}'")
+
+    # Test 2: Check that the binary is installed and works
+    # The binary should be in PATH after editable install
```
konstin: We should avoid accepting a global binary for the test:

```suggestion
    # The binary should be in PATH in the venv bin dir after editable install
    bin = which("bin-with-python-module")
    if not os.path.realpath(bin).startswith(os.path.realpath(sys.prefix)):
        raise Exception(f"Binary '{os.path.realpath(bin)}' not in venv at '{os.path.realpath(sys.prefix)}'")
```

konstin (approved):
Thanks!

messense:
CI failure is unrelated.

PR #2940, opened 2026-01-16
Added automatic type stub generation with pyo3-introspection

varchasgopalaswamy:
This PR adds a basic mechanism for type stub generation with `pyo3-introspection`. 

Perhaps there is a better way to do it, but the most obvious way to get the stubs into the wheel seemed to be by 

1. Generating the stubs right after `compile`.
2. Writing the stubs out in a folder right next to where the `.so`/`.dll` file ends up
3. Stuffing the paths to the stubs into `BuildArtifact`
4. Having `binding_generator` append the stubs into the `additional_files` field that it carries around to put python files into the wheel

I've currently just implemented the last step for `pyo3_binding` but I guess there's no reason why it couldnt be applied elsewhere. I suppose the stubs could also be generated during the binding generation rather than during compile, but I think it's nice to be able to see the generated stubs and be able to check that they look good before they get stuffed into a wheel.

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -97,9 +97,10 @@ pub fn compile(
 
         if let Err(e) = generate_stubs_for_module(&stub_dir, module_name, module_path) {
             eprintln!(
-                "⚠️  Warning: Failed to generate type stubs for module '{}' due to error: {}",
+                "⚠️  Error! Failed to generate type stubs for module '{}' due to error: {}",
                 module_name, e
             );
+            return Err(e);
         } else {
             artifact.stub_dir = Some(stub_dir.to_path_buf());
             eprintln!(


Tpt:
[for context I am the main person working on introspection in Pyo3]

Thank you for this! I am not sure pyo3-introspection in its released version is ready to be part of a build pipeline, there are still to many issues. I am scared it will move people out of the feature even if it's quite promising. A good first step might a new `maturin generate-stubs` command or something like this that would allow people to have a look at the generated output and maybe fix it and commit it?

For context, I also have a draft commit adding a `--introspect-stubs` option to the build pipeline, it hooks into maturin in the `generate_binding` function: 92b90be7adf295f88c456fc2d8141afd579262e8. Not sure which approach is best. But anyway, I would prefer to make crystal clear the current state of the stub introspection code is not ready for use at the moment.

varchasgopalaswamy:
Yes, I think feature gating it, maybe behind a CLI arg might be a good idea to let people know it's experimental and let them opt in. But my two cents is that exposing it even just as a preview feature through maturin would make it a lot easier for folks to give feedback! Since pyo3-introspection is lib only, it requires a bit of work to get it to start generating stubs. 

That said, it's working pretty well so far!

Tpt:
> would make it a lot easier for folks to give feedback! 

Yes! But I am afraid the missing features will be frustrating to the people trying and I guess most of the feedback will be elements still on the TODO list or already fixed in the unreleased version. Waiting a few month for at least the next pyo3 release seems a bit better to me.

varchasgopalaswamy:
> > would make it a lot easier for folks to give feedback!
> 
> Yes! But I am afraid the missing features will be frustrating to the people trying and I guess most of the feedback will be elements still on the TODO list or already fixed in the unreleased version. Waiting a few month for at least the next pyo3 release seems a bit better to me.

Fair point. I guess we can leave this PR open so people can test-drive the introspection changes, and re-visit this once more of the TODOs are dealt with!

messense:
Done in #3105

PR #2942, opened 2026-01-18
Allow build armv6l for musllinux

andypost:
Release [1.11.4](https://github.com/PyO3/maturin/releases/tag/v1.11.4) added support for the `armv6l` arch but upgrading the package for Alpinelinux from [1.9.6 to 1.11.5](https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/95959) caused build failure  for the arch so I have to patch

The tail of log 
```
    📦 Including files matching "pyo3_mixed/assets/*"
    📦 Built source distribution to /tmp/.tmpfacQKA/dist/pyo3_mixed-2.1.5.tar.gz
    🔗 Found pyo3 bindings
    🐍 Found CPython 3.12 at /usr/bin/python3
    Cause: Your library is not musllinux_1_2 compliant because it has unsupported architecture: armv6l
    Cause: Error ensuring musllinux_1_2 compliance
    thread 'unreadable_dir' (21096) panicked at tests/run.rs:610:5:
    Error ensuring musllinux_1_2 compliance
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
        PASS [  42.327s] (123/126) maturin::run pyo3_source_date_epoch
        PASS [  42.596s] (124/126) maturin::run pep517_editable_profile
        PASS [  43.551s] (125/126) maturin::run integration_workspace_inverted_order
        PASS [  50.646s] (126/126) maturin::run pep517_default_profile
────────────
     Summary [ 383.739s] 126 tests run: 125 passed (17 slow), 1 failed, 12 skipped
        FAIL [  40.011s] (122/126) maturin::run unreadable_dir
error: test run failed
```

### DIFF
--- a/src/target/legacy_py.rs
+++ b/src/target/legacy_py.rs
@@ -60,5 +60,6 @@ pub(super) static MANYLINUX_ARCHES: &[&str] = &[
     "x86_64", "i686", "aarch64", "armv7l", "ppc64le", "s390x", "ppc64", "riscv64",
 ];
 
-pub(super) static MUSLLINUX_ARCHES: &[&str] =
-    &["x86_64", "i686", "aarch64", "armv6l", "armv7l", "ppc64le", "s390x"];
+pub(super) static MUSLLINUX_ARCHES: &[&str] = &[
+    "x86_64", "i686", "aarch64", "armv6l", "armv7l", "ppc64le", "s390x",
+];


src/auditwheel/musllinux-policy.json
```diff
@@ -43,6 +43,8 @@
       },
       "s390x": {
       },
+      "armv6l": {
+      },
```
konstin, line 47: This file is vendored from https://github.com/pypa/auditwheel/blob/main/src/auditwheel/policy/musllinux-policy.json, we shouldn't diverge from it. Maybe we need to update it upstream?

konstin:
armv6l is kind of an odd platform cause pypi supports it as `linux_armv6l`. We need to fix the arm-unknown-linux-musleabihf target (sorry for breaking it), but I think we need to do something different about tagging it.

src/auditwheel/musllinux-policy.json
andypost: Looks like it was skipped on initial compatibility https://github.com/pypa/auditwheel/pull/315#discussion_r676923741

PR #2950, opened 2026-01-27
Use mmap for faster warn_missing_py_init

orlp:
One of my colleagues noticed that `warn_missing_py_init` is quite slow, reading the full file into memory (for context: the Polars debug .so is ~4GB).

I think it's quite likely that `goblin` only needs to parse a portion of the file, so I think using `mmap` here makes sense.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -111,9 +111,7 @@ jobs:
       - name: Archive binary (windows)
         if: matrix.os == 'windows-latest'
         run: |
-          cd target/maturin
-          7z a ../../${{ matrix.name }} ${{ github.event.repository.name }}.exe
-          cd -
+          7z a ${{ matrix.name }} target/maturin/${{ github.event.repository.name }}.exe
 
       - name: Archive binary (linux)
         if: matrix.os == 'ubuntu-latest'
--- a/src/build_context.rs
+++ b/src/build_context.rs
@@ -1072,6 +1072,9 @@ impl BuildContext {
         fs::create_dir_all(&maturin_build)?;
         let artifact_path = &artifact.path;
         let new_artifact_path = maturin_build.join(artifact_path.file_name().unwrap());
+        // Remove any stale file at the destination so that `fs::rename` succeeds
+        // on Windows (where rename fails if the destination already exists).
+        let _ = fs::remove_file(&new_artifact_path);
         if fs::rename(artifact_path, &new_artifact_path).is_err() {
             // Rename fails across filesystem boundaries, fall back to copy
             fs::copy(artifact_path, &new_artifact_path)?;
--- a/tests/common/pep517.rs
+++ b/tests/common/pep517.rs
@@ -94,19 +94,10 @@ fn insert_path(env_var: &str, new_path: &Path) -> String {
         .expect("PATH is not valid utf8")
 }
 
-/// Whether cargo built the shared library for the specified cargo profile in the test target
-/// directory.
+/// Whether cargo built for the specified cargo profile in the test target directory.
 pub fn target_has_profile(unique_name: &str, profile: &str) -> bool {
-    let shared_library = if cfg!(windows) {
-        "pyo3_pure.dll"
-    } else if cfg!(target_os = "macos") {
-        "libpyo3_pure.dylib"
-    } else {
-        "libpyo3_pure.so"
-    };
-    PathBuf::from(target_dir(unique_name))
-        .join(profile)
-        .join("deps")
-        .join(shared_library)
-        .is_file()
+    let profile_dir = PathBuf::from(target_dir(unique_name)).join(profile);
+    // Check for cargo's .fingerprint directory which is always created for the
+    // profile that was used, and is not affected by maturin's artifact staging.
+    profile_dir.join(".fingerprint").is_dir()
 }


src/compile.rs
```diff
@@ -605,11 +605,10 @@ fn compile_target(
 #[instrument(skip_all)]
 pub fn warn_missing_py_init(artifact: &Path, module_name: &str) -> Result<()> {
     let py_init = format!("PyInit_{module_name}");
-    let mut fd = File::open(artifact)?;
-    let mut buffer = Vec::new();
-    fd.read_to_end(&mut buffer)?;
+    let fd = File::open(artifact)?;
+    let mmap = unsafe { memmap2::Mmap::map(&fd).context("mmap failed")? };
```
konstin, line 760: Is that sound, does it guarantee non of the rust invariants are being violated? If so, can you write a safety comment on why it it safe to do?

orlp: @konstin Please see https://docs.rs/memmap2/0.9.9/memmap2/struct.Mmap.html#safety.

I'm not exactly sure what kind of safety comment you'd be looking for. It is technically inherently 'unsound' by the Rust abstract model if someone edits the file in the file system at the same time. But `mmap` is very widely used regardless, including in the Rust project itself, e.g. https://github.com/rust-lang/rust/blob/94a0cd15f5976fa35e5e6784e621c04e9f958e57/src/tools/rust-analyzer/crates/proc-macro-srv/src/dylib.rs#L110.

konstin: I'm afraid I'm not comfortable with using mmap like that in maturin. For context, the only production usage of `unsafe` we have in maturin currently is `env::set_var` and that also only because it used to be safe in a previous edition.

> Applications must consider the risk and take appropriate precautions when using file-backed maps. Solutions such as file permissions, locks or process-private (e.g. unlinked) files exist but are platform specific and limited.

As long as we can't guarantee that the artifact is not actually modified by some means, we risk UB, and that's a risk that's not worth it for this performance win.

konstin:
> for context: the Polars debug .so is ~4GB

While I agree that we should speed up maturin for large binaries, have you considered reducing the debug info, such as https://doc.rust-lang.org/cargo/guide/build-performance.html#reduce-amount-of-generated-debug-information? I found those changes to make building a lot faster, producing a 4GB file is a large task for rustc and the linker too.

src/compile.rs
orlp: @konstin Which risks are you worried about? 'UB' is not a risk in of itself, it's a binary property of a program that states the abstract model of Rust was violated, a risk needs an actual attacker model, consequence and probability of exploitation.

Mmap is already widely used in the Rust compiler: <https://github.com/search?q=repo%3Arust-lang%2Frust%20mmap&type=code>. I'd really like to know which risks in particular you're worried about that are unacceptable for maturin but are apparently acceptable for the Rust compiler.

konstin: I don't know which measure or soundness considerations the rust-lang/rust repo has around mmap, but the documentation you linked at https://docs.rs/memmap2/0.9.9/memmap2/struct.Mmap.html#safety is very clear that unless we guarantee that the underlying file doesn't change, this is unsound. I assume those mmap calls ensure this, but here I don't see that we've ensured that (yet).

With rust being safe by default, using `unsafe` inverts the language contract: We need to proof and document at the site where we use the `unsafe` that we're keeping up the contract, and should also document why the unsafe is necessary in the first place, usually with a `// SAFETY:` comment.

orlp: @konstin "Measure of soundness" is the wrong measure. The API is **fundamentally** unsound. You can not prevent someone from mutating the file. There also is no alternative without re-writing the entire parser to be `Read` based rather than `&[u8]` based.

I'm perfectly aware of how Rust works with unsafe. You are however holding me to an impossible standard. There is no safety comment I could possibly write containing the contract we uphold, because we can not possibly guarantee that the contract is upheld. **This is no different in the Rust compiler.**

However, would you be satisfied with the following logic?

```rust
// SAFETY: technically this is unsound if the file gets modified while we hold the
// immutable &[u8] slice. However, considering we're reading a dynamic library
// with executable code intended to be run, should a malicious agent have write
// access to it, arbitrary code execution is already trivially achieved. So there
// is no reasonable attacker model where this would lead to exploits that are
// otherwise impossible.
let mmap = unsafe { memmap2::Mmap::map(&fd).context("mmap failed")? };
```

konstin: I meant more in the sense of, can we ensure that this is not a file that changes underneath the mmap if you do another `cargo build` while maturin is running? As I understand it, there's a lot of ecosystem usage, where generally those can either assume that the file doesn't change underneath them (they hold a lock, it's a temp dir, they are the only process in that docker container, etc.) or the file changing isn't catastrophic. I know this is unlikely with cargo being slow and how this check is timed, but the uv issue tracker has scared me that every timing and non-determinism bug that can happen, will happen.

orlp: > can we ensure that this is not a file that changes underneath the mmap if you do another cargo build while maturin is running?

No, you can not in general. And should this be a problem (it's not, really), you're already in the woods because you call

https://github.com/PyO3/maturin/blob/e7922abab54a412cf7c527c41cbcfb170dafc940/src/auditwheel/repair.rs#L15

which does... you guessed it, `mmap`:

https://github.com/messense/lddtree-rs/blob/e3ab19edf39b4d2d9cc90058e53a7ec966c271ac/src/lib.rs#L133

I must re-iterate that I feel you're holding me to an impossibly high standard for this PR, considering your tool already does the exact same strategy I proposed elsewhere (as well as your dependencies in the Rust compiler, probably others).

orlp:
The result would still be an 1.7 GB binary, so still significant overhead to fully read in `warn_missing_py_init`. Plus having to wait several minutes for a complete recompile with debug symbols should a debugger be needed is rather undesirable.

orlp:
What's the status on this?

messense:
I'm generally in favor of use mmap here, maybe to be safer we can lock the `target/$profile/.cargo-lock` file lock so no other cargo build can proceed at the same time?

konstin:
We should move the artifacts to a temp file before we process them, but if we're already using mmap we have that risk already now.

messense:
The artifact is a cdylib that can be multi-GB in debug builds, a 4GB ~~copy~~move would be just as slow as the original `read_to_end` if target path is not on the same disk. The cargo lock coordination we have now is both faster (zero-copy) and provides stronger guarantees than a temp file copy, which itself would be subject to the same race (the source could change mid-copy, producing a corrupt temp file).

messense:
About the `lddtree` issue, its analysis does not need the same cargo lock protection, because:

* lddtree runs on a staged copy — `copy_artifact_for_repair()` copies the artifact to a maturin-owned staging directory before `get_policy_and_libs()` / `find_external_libs()` runs. lddtree mmaps this copy, not the original cargo output.
* lddtree analyzes system libraries — Beyond the main artifact, lddtree resolves and reads shared libraries from the system sysroot (e.g., /lib/, /usr/lib/). These are not modified by cargo builds.

And we don't always copy the artifact, only when doing auditwheel check/repair.

messense:
so moving the artifact is a viable option if we keep it in `target` directory which should usually be the same disk so we get atomic `mv`, let me also try it.

messense:
Implemented move in https://github.com/PyO3/maturin/pull/2950/changes/b607b34df368a864f1d858dcbb703548029008f2, but I think it will break any project that relies on the artifact in the standard cargo target location like in maturin release pipeline when packaging binaries for github release which is unfortunate.

messense:
> Implemented move in [b607b34](https://github.com/PyO3/maturin/commit/b607b34df368a864f1d858dcbb703548029008f2), but I think it will break any project that relies on the artifact in the standard cargo target location like in maturin release pipeline when packaging binaries for github release which is unfortunate.

@konstin this could affect `uv` release pipeline so FYI. The fix should be similar to https://github.com/PyO3/maturin/pull/2950/commits/038aaade0efe88180de35c4370b3dabdf7180907

konstin:
That looks like a breaking change to me, at least I expect this will break a lot of build pielines.

PR #2965, opened 2026-02-06
Normalize wheel `RECORD` paths (on Windows)

texodus:
#2870 refactored `module_writer.rs`, and in doing so [removed a line](https://github.com/PyO3/maturin/commit/d5936050f63248216bdeef18d0c5057aab2cd1e7#diff-a5f91562326af952084dcfb75851be91314679f06545de0c7ea8e9057ebe6ad5L55) which reformatted `PathBuf` rendered platform-specific `String`s to exclusively us `/`:

```python
// The zip standard mandates using unix style paths
let target = target.to_str().unwrap().replace('\\', "/");
```

While it seems this is [technically allowed](https://packaging.python.org/en/latest/specifications/recording-installed-packages/#the-record-file), some tools choke on this. In my case, the [`wheel`](https://pypi.org/project/wheel/) package does not like windows-style paths in the `RECORD` file (even on Windows itself).

I've also added a unit test to validate this - though because the implementation uses `PathBuf::to_string_lossy`, it will only actually fail on Windows.

### DIFF
--- a/src/module_writer/wheel_writer.rs
+++ b/src/module_writer/wheel_writer.rs
@@ -8,6 +8,7 @@ use std::path::PathBuf;
 use anyhow::Context as _;
 use anyhow::Result;
 use fs_err::File;
+use path_slash::PathBufExt as _;
 use tracing::debug;
 use zip::ZipWriter;
 use zip::write::SimpleFileOptions;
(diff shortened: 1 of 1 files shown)


messense:
Thanks for the PR, Windows CI is failing, we need to get it fixed to merge this.

Issue #2773, opened 2025-10-14
Missing LICENSE file from wheel even when [project] is not defined in pyproject.toml


ddelange:
### Bug Description

The LICENSE file will be missing from wheels even if `[project]` is not defined in `pyproject.toml`. See https://github.com/oconnor663/blake3-py/pull/69 to work around this bug and add LICENSE file to the wheel, but removed numerous entries from METADATA, and then https://github.com/oconnor663/blake3-py/pull/88 to fix the latter.

### Your maturin version (`maturin --version`)

logs no longer available

### Your Python version (`python -V`)

all affected

### Your pip version (`pip -V`)

--

### What bindings you're using

pyo3

### Does `cargo build` work?

- [x] Yes, it works

### If on windows, have you checked that you aren't accidentally using unix path (those with the forward slash `/`)?

- [x] Yes

### Steps to Reproduce

see PR linked above. without that PR, METADATA entries are present but the actual LICENSE file is missing from dist-info in the wheels.

Issue #2743, opened 2025-09-19
Maturin fails to use workspace license file


anuradhawick:
### Bug Description

I have a workspace with multiple rust crates. The following fails

```toml
[package]
name = "pip"
version.workspace = true
edition.workspace = true
authors.workspace = true
description.workspace = true
readme.workspace = true
license-file.workspace = true
```

with error

```
💥 maturin failed
  Caused by: license file `/home/anuvini/Desktop/kmertools/LICENSE` exists outside of the project root `/home/anuvini/Desktop/kmertools/pip`
  Caused by: prefix not found
```

If I remove `license-file.workspace = true` and add a symlink to license in root folder, it succeeds.

Only started seeing this issue since I updated from 1.7.8 to 1.9.4

### Your maturin version (`maturin --version`)

1.9.4

### Your Python version (`python -V`)

3.10.9

### Your pip version (`pip -V`)

pip 23.0

### What bindings you're using

pyo3

### Does `cargo build` work?

- [x] Yes, it works

### If on windows, have you checked that you aren't accidentally using unix path (those with the forward slash `/`)?

- [ ] Yes

### Steps to Reproduce

Create a rust workspace

Add a crate that uses maturin

Add license path as follows in the Cargo.toml

```toml
license-file.workspace = true
```

vemonet:
We are seeing the same issue, it is quite problematic because it requires either to:

- Duplicate the license file (which is horrible design)
- Create a symlinks (which can't be trusted, they are often a source of bug and else)

Is this a voluntary design or a bug?

Most maturin projects are mono repos with multiple crates/packages in it (that's the nature of the systems build with it), it would be really nice if maturin could take this into account and enable license file in parent folders

anuradhawick:
Exactly my problem. Project contains multiple crates and I intend to share the license. 

Using symlink is more of a hack when the option to use workspace license is clearly there. 

Seems to me like a bug.

vemonet:
Cool, now `cargo deny` fails due to the symbolic link (as expected.)

```
error[unlicensed]: nanopub-sign = 0.1.4 is unlicensed
  ┌─ path+file:///Users/vemonet/dev/nanopub/nanopub-rs/python#nanopub-sign@0.1.4:2:9
  │
2 │ name = "nanopub-sign"
  │         ━━━━━━━━━━━━ a valid license expression could not be retrieved for the crate
3 │ version = "0.1.4"
4 │ license = ""
  │            ─ license expression was not specified
[...]

Issue #2664, opened 2025-06-25
License file being deleted in Cargo workspace


krokosik:
### Bug Description

Maturin v1.9.0 introduced a new license handling mechanism in #2571 

This caused our workspace to be impossible to build now, as the tool now deletes our license file and then fails due to lack of license file.

```shell
💥 maturin failed
  Caused by: Failed to read C:\fake-workspace-path\host\..\LICENSE
  Caused by: failed to open file `C:\fake-workspace-path\host\..\LICENSE`: The system cannot find the file specified. (os error 2)
```

I tried working around it by manually setting the path in `pyproject.toml`, but this also fails, because parent directory operator `..` is not allowed.

### Your maturin version (`maturin --version`)

1.9.0

### Your Python version (`python -V`)

3.11.10

### Your pip version (`pip -V`)

23.0.1

### What bindings you're using

pyo3

### Does `cargo build` work?

- [ ] Yes, it works

### If on windows, have you checked that you aren't accidentally using unix path (those with the forward slash `/`)?

- [x] Yes

### Steps to Reproduce

1. Create a cargo workspace with a crate that uses maturin
2. Create a LICENSE file in the root folder and specify it in the root Cargo.toml
3. In the crate Cargo.toml, use `license-file.workspace = true`
4. Try executing `maturin develop`

messense:
Sorry I'm not sure what you meant by `deleted`, #2571 is not merged, the change in #2647 does not delete anything.

Please create a minimal repro in a public git repository, otherwise it's going to take more time for this to be resolved.

krokosik:
@messense sorry, my bad. I linked the wrong issue. Here is the minimal repo:
https://github.com/krokosik/maturin-2664

messense:
I'm unable to reproduce this on macOS.

```bash
maturin-2664/host on  main is 📦 v0.1.5 via 🐍 v3.13.5 via 🦀 v1.87.0
❯ uv run maturin develop
Uninstalled 1 package in 1ms
Installed 1 package in 2ms
🔗 Found pyo3 bindings with abi3 support
🐍 Not using a specific python interpreter
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.00s
📦 Built wheel for abi3 Python ≥ 3.9 to /var/folders/wv/3487fygn67qfcw9j29b2gyyh0000gn/T/.tmpl5wjb8/rustpill_clients-0.1.5-cp39-abi3-macosx_11_0_arm64.whl
[...]

PR #2970, opened 2026-02-08
Fix license file handling for workspace-level license files

messense:
Fix three related bugs where license files in Cargo workspaces were not handled correctly:

- #2743: `license-file.workspace = true` fails with 'exists outside of the project root' because `strip_prefix(pyproject_dir)` errors when the license file is at the workspace root, not the crate directory.

- #2664: License file being deleted in Cargo workspace due to path normalization producing '..' paths that break downstream handling.

- #2773: Missing LICENSE file from wheel when `[project]` is not defined in pyproject.toml because license normalization only ran inside the `if-let-Some(project)` block.

Changes:
- Move license file normalization out of the `[project]` guard so it runs unconditionally, even when pyproject.toml has no `[project]` table.
- When `strip_prefix` fails (file outside `pyproject_dir`), fall back to using just the filename and record the absolute source path in a new `license_file_sources` map.
- Update `write_dist_info` to use the source map for copying license files that live outside the pyproject directory.
- Add `rewrite_cargo_toml_license_file` (mirroring `rewrite_cargo_toml_readme`) to rewrite `package.license-file` in sdist Cargo.toml so building from the sdist works correctly.
- Copy workspace-level license files next to Cargo.toml in the sdist, following the same pattern used for readme files.

### DIFF
--- a/src/source_distribution.rs
+++ b/src/source_distribution.rs
@@ -462,20 +462,29 @@ fn rewrite_pyproject_toml(
     Ok(data.to_string())
 }
 
+/// Describes the role of a crate being added to the source distribution.
+enum CrateRole<'a> {
+    /// The main Python binding crate: rewrite workspace deps in Cargo.toml,
+    /// skip pyproject.toml from `cargo package --list` (handled separately).
+    Root {
+        known_path_deps: &'a HashMap<String, PathDependency>,
+    },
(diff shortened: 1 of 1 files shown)


krokosik:
Wonderful, thank you so much!

PR #2980, opened 2026-02-10
feat: Add PEP 770 SBOM support

### LINKED ISSUE #2554: PEP 770 (SBOM) support
https://peps.python.org/pep-0770/

I think there's two elements:

1. Generating an SBOM for rust crates that are built in a wheel
2. Providing a way for users to specify additional SBOMs to be included

messense:
Implements Software Bill of Materials (SBOM) generation according to PEP 770.

Changes:
- Add optional on-by-default 'sbom' feature with cargo-cyclonedx integration
- Add `SbomConfig` to pyproject.toml parsing
- Create sbom module for SBOM generation using cargo-cyclonedx library
- Update wheel writers to include SBOMs in `.dist-info/sboms` directory
- Update all wheel building methods (PyO3, CFFI, UniFfi, Bin) to generate SBOMs

The implementation uses cargo-cyclonedx as a library dependency, eliminating the need for users to install it separately. When the sbom feature is enabled and configured, maturin automatically generates CycloneDX SBOMs for the Rust dependencies and includes them in the wheel.

Closes #2554

### DIFF
--- a/tests/common/other.rs
+++ b/tests/common/other.rs
@@ -345,7 +345,7 @@ pub fn check_wheel_files_with_sbom(package: impl AsRef<Path>, unique_name: &str)
 
     let sbom_files: Vec<String> = wheel
         .file_names()
-        .filter(|f| f.contains("sboms/"))
+        .filter(|f| f.contains(".dist-info/sboms/"))
         .map(|f| f.to_string())
         .collect();
     assert!(


tests/common/other.rs
```diff
@@ -339,6 +339,30 @@ pub fn check_wheel_files(
     Ok(())
 }
 
+#[cfg(feature = "sbom")]
+pub fn check_wheel_files_with_sbom(package: impl AsRef<Path>, unique_name: &str) -> Result<()> {
+    let wheel = build_wheel_files(&package, unique_name)?;
+
+    let sbom_files: Vec<String> = wheel
+        .file_names()
+        .filter(|f| f.contains("sboms/"))
```
messense:
```suggestion
        .filter(|f| f.contains(".dist-info/sboms/"))
```

PR #2982, opened 2026-02-10
fix: exclude duplicate python source files from sdist for workspace members

### LINKED ISSUE #2383: Source distribution (`sdist`) contains source files and the whole project directory itself as a nested folder
### Bug Description

Hello!

When analyzing the source distribution of my Rust/Python package, named `differt-core`, I noticed that the content of `.tar.gz` was as follows:

```bash
> tree -L dist/differt_core-0.0.30/
dist/differt_core-0.0.30/
├── Cargo.lock
├── Cargo.toml
├── differt-core  # <-- I want to exclude this, as this is duplicate (and undesired) contents
│   ├── benches
│   ├── Cargo.toml
│   ├── LICENSE.md
│   ├── python
│   ├── README.md
│   ├── src
│   └── tests
├── LICENSE.md
├── PKG-INFO
├── pyproject.toml
├── python
│   └── differt_core
├── README.md
└── src
    ├── geometry
    ├── lib.rs
    ├── rt
    └── scene
```

As you can see, the `python` folder (with Python source files) is included both `./` and `./differt-core`, and I don't know how to exclude this (because this is not needed to build from source): I tried various exclude patterns, without success.

I have looked at other projects, and found that `py-polars` had the same "*issue*" in that its source distribution also contains the project folder itself.

However, projects like `uv` do not have this issue.

**The main observed difference** between `uv` and my project (as well as `py-polars`) is that `uv` is defined in the root directory, whereas my project (and `py-polars`) are defined in a subdirectory.

Do you have any idea how to fix this?

Thanks in advance for your help!

Relevant links:

- My project (`differt-core` subdirectory): https://github.com/jeertmans/DiffeRT/tree/v0.0.30/differt-core (and [PyPI release files](https://pypi.org/project/differt-core/0.0.30/#files))
- `py-polars`: https://github.com/pola-rs/polars/tree/py-1.17.1/py-polars (and [PyPI release files](https://pypi.org/project/polars/1.17.1/#files))
- `uv`: https://github.com/astral-sh/uv/tree/0.5.8 (and [PyPI release files](https://pypi.org/project/uv/0.5.8/#files))

1.7.8

N/A

N/A

pyo3

### Does `cargo build` work?

- [X] Yes, it works

- [X] Yes

### Steps to Reproduce

You can either download the source distribution from the above links, or build from source with either:

```bash
maturin sdist --out dist --manifest-path differt-core/Cargo.toml
```

or with uv:

```bash
uv build --package differt-core --sdist -o dist
```

Next, open the archive.

messense:
When a project is in a subdirectory of a Cargo workspace, `cargo package --list` returns python source files that are also added by the explicit python source loop. This results in duplicate files in the sdist: once under the crate subdirectory prefix (e.g. `my-crate/python/...`) and once at the top level (e.g. `python/...`).

Skip files from `cargo package --list` that fall under python source directories when the crate is a workspace member in a subdirectory, since those files will be added separately by the python source loop.

Fixes #2383

### DIFF
--- a/src/source_distribution.rs
+++ b/src/source_distribution.rs
@@ -483,9 +483,6 @@ enum CrateRole<'a> {
 /// and rewriting path entries in Cargo.toml
 ///
 /// Runs `cargo package --list --allow-dirty` to obtain a list of files to package.
-///
-/// `skip_prefixes` is a list of path prefixes (relative to the manifest directory) whose files
-/// should be skipped because they are added separately (e.g. python source directories).
 fn add_crate_to_source_distribution(
     writer: &mut VirtualWriter<SDistWriter>,
     manifest_path: impl AsRef<Path>,


src/source_distribution.rs
```diff
@@ -478,6 +483,9 @@ enum CrateRole<'a> {
 /// and rewriting path entries in Cargo.toml
 ///
 /// Runs `cargo package --list --allow-dirty` to obtain a list of files to package.
+///
+/// `skip_prefixes` is a list of path prefixes (relative to the manifest directory) whose files
+/// should be skipped because they are added separately (e.g. python source directories).
```
messense:
```suggestion
```

PR #2990, opened 2026-02-11
Relax ziglang dependency version requirement

### LINKED ISSUE #2989: Allow more recent version of ziglang
Currently ziglang optional dependency is restricted to:
```
zig = ["ziglang>=0.10.0,<0.13.0"]
```

Does maturin not work with more recent versions of ziglang, or is this an artifical limitation?

Currnt version restriction seems to be introduces in commit: 
https://github.com/PyO3/maturin/commit/b853aec3d76ab52c82289106e50ac6e8881fbf79

If it also works with more recent versions (e.g. 0.15.2), the version restriction in `maturin-action` repo should be fixed too.

messense:
This is possible now that we upgraded `cargo-zigbuild` to 0.21.

Closes #2989

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -111,7 +111,6 @@ jobs:
           python_version=$(echo ${{ matrix.python-version }} | sed -e s/-dev//)
           echo "PYTHON_VERSION=$python_version" >> "${GITHUB_ENV}"
       - uses: mlugg/setup-zig@v2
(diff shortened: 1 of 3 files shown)


messense:
N.B. there is still a bug in `cargo-zigbuild` when filtering `-exported_symbols_list`

>  "-Wl,-exported_symbols_list" "-Wl,C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\rustcZKquwX\\list"

Issue #2945, opened 2026-01-22
Maturin does not work in Python 3.13 on Android unless I edit Maturin


robertkirkman:
### Bug Description

using `pip install meshage` does not work in Python 3.13 on Android because this error occurs:

```
💥 maturin failed
Caused by: Failed to get information from the python interpreter at /data/data/com.termux/files/usr/bin/python3.13
Caused by: platform.system() in python, android, and the rust target, Target { os: Linux, arch: X86, env: Android, triple: "i686-linux-android", cross_compiling: false, rustc_version: VersionMeta { semver: Version { major: 1, minor: 92, patch: 0 }, commit_hash: Some("ded5c06cf21d2b93bffd5d884aa6e96934ee4234"), commit_date: Some("2025-12-08"), build_date: None, channel: Stable, host: "i686-linux-android", short_version_string: "rustc 1.92.0 (ded5c06cf 2025-12-08) (built from a source tarball)", llvm_version: Some(LlvmVersion { major: 21, minor: 1 }) }, user_specified: false }, don't match ಠ_ಠ
Error running maturin: Command '['maturin', 'pep517', 'write-dist-info', '--metadata-directory', '/data/data/com.termux/files/usr/tmp/pip-modern-metadata-a3waxdbn', '--interpreter', '/data/data/com.termux/files/usr/bin/python3.13']' returned non-zero exit status 1.
Checking for Rust toolchain....
Running `maturin pep517 write-dist-info --metadata-directory /data/data/com.termux/files/usr/tmp/pip-modern-metadata-a3waxdbn --interpreter /data/data/com.termux/files/usr/bin/python3.13`
```

However, when I edit Maturin and use several other commands like this, `meshage` installs successfully in Python 3.13 on Android.

```bash
git clone https://github.com/PyO3/maturin.git
cd maturin
git apply -v << 'EOF'
--- a/src/target/mod.rs
+++ b/src/target/mod.rs
@@ -456,6 +456,9 @@ impl Target {
 
     /// Returns the name python uses in `sys.platform` for this os
     pub fn get_python_os(&self) -> &str {
+        if (self.is_bionic_libc()) {
+            return "android";
+        }
         match self.os {
             Os::Windows => "windows",
             Os::Linux => "linux",
@@ -682,6 +685,15 @@ impl Target {
         )
     }
 
+    /// Returns true if the current platform's target env is Android
+    #[inline]
+    pub fn is_bionic_libc(&self) -> bool {
+        matches!(
[...]

PR #2992, opened 2026-02-13
fix: platform.system() on Android for Python 3.13+

thunder-coding:
Python from 3.13 onwards supports Android as an official platform. As a part of this official support, there were a bunch of changes including `platform.system()` returning "Android" and `sys.platform` being "android" instead of "Linux" and "linux" on earlier versions. This is a breaking change.

Also try to be backwards compatible while running on Android for older python versions in environments like Termux

This maturin bug was found as a part of testing of builds of python packages for Termux as part of Python 3.13 migration. Downstream Python distribution update PR: https://github.com/termux/termux-packages/pull/27739

Ideally we would like to ensure that get_python_os() returns "linux" for Python versions <= 3.12, but that'd require additional work in supplying python version to get_python_os(), but that'd complicate some things. This works well enough, and shouldn't be necessary in future when Python >= 3.13 becomes the standard target by most (which should ideally happen soon since upstream Python is supporting Android from this version).

My analysis of maturin reveal that get_python_os() returning "android" shouldn't cause any side effects for Python <= 3.12 on Android.

Issue also reported to maturin in https://github.com/PyO3/maturin/issues/2945

Partially based of off work by @robertkirkman in the linked issue. Further fixup was done to ensure suitability for upstream acceptance.

Ref: https://github.com/python/cpython/pull/116215

CC @mhsmith Android support for Python

### DIFF
--- a/src/python_interpreter/mod.rs
+++ b/src/python_interpreter/mod.rs
@@ -295,8 +295,8 @@ fn fun_with_abiflags(
         && !(target.get_python_os() == "cygwin"
             && message.system.to_lowercase().starts_with("cygwin"))
     {
-        // Python versions <= 3.12 used to report sys.platform as "linux". Only on Python versions
-        // >= 3.13, sys.platform reports as "android". So maintain backwards compatibility with
+        // Python versions <= 3.12 used to report platform.system() as "linux". Only on Python versions
+        // >= 3.13, platform.system() reports as "android". So maintain backwards compatibility with
         // Python 3.12 when compiling on Android environment (for e.g. Termux)
         let is_android_compat = target.get_python_os() == "android"
             && message.system == "linux"


robertkirkman:
> My analysis of maturin reveal that get_python_os() returning "android" shouldn't cause any effects on Android targets that do need to be patched.

Do you mean "don't need to be patched" rather than "do need to be patched" here, or did you really mean to say "do need to be patched"?

thunder-coding:
> Do you mean "don't need to be patched" rather than "do need to be patched" here, or did you really mean to say "do need to be patched"?

I meant to say that.  `get_python_os()` returning "android" instead of "linux" for Python <= 3.12 should not be an issue in maturin's codebase as far as I can see. Apologies for the confusing sentence, English is a funny language which can be difficult to grasp depending on context :)

robertkirkman:
> get_python_os() returning "android" instead of "linux" for Python <= 3.12 should not be an issue in maturin's codebase as far as I can see

what's confusing me is that Maturin doesn't need to be patched when running on Python 3.12 on Android, so what did you mean is the thing being patched?

thunder-coding:
> what's confusing me is that Maturin doesn't need to be patched when running on Python 3.12 on Android, so what did you mean is the thing being patched?

Yes Maturin doesn't need to be patched for Python 3.12 on Android. But this change does cause `get_python_os` to behave differently on android than the code comments above. on Android it will not return the value of `platform.system()` for Python <= 3.12. This value is compared against the Python runtime detected when maturin is run, and it tries to check if it matches `message.system`, which doesn't for Python <= 3.12. I was refering to this difference over in my comment. Seems like I should have worded that a bit more clearly

robertkirkman:
Thanks for explaining, that makes sense. If I understand now, this means that you could say '`get_python_os()` returning "android" shouldn't cause any effects on Android targets that do need to be granted an exception in `fun_with_abiflags()` along with these changes' instead of '`get_python_os()` returning "android" shouldn't cause any effects on Android targets that do need to be patched'?

Edit: new description in the PR is good

robertkirkman:
@thunder-coding unfortunately, after this PR, this command

```bash
scripts/run-docker.sh ./build-package.sh -I -f ruff
```

has this error for me:

```
📦 Built wheel to /home/builder/.termux-build/ruff/src/target/wheels/ruff-0.15.2-py3-none-android_6_12_47_gentoo_dist_x86_64.whl
mv: cannot stat 'target/wheels/ruff-0.15.2-py3-none-android_24_arm64_v8a.whl': No such file or directory
```

it looks like the result from the `uname -r` command, which can vary greatly by what computer is running Maturin, is getting into the name of the `.whl` file, but I'm not sure how.

I'm sorry that I don't know how to fix this other than to bisect it to this PR, but I just thought you would want to know.

PR #3008, opened 2026-02-16
docs(config): minor fixes

mkniewallner:
Noticed that the replacement for `skip-auditwheel` mentions `auditwheel = "audit"` instead of `auditwheel = "skip"`, and found a few other fixes in the "Configuration" documentation.


guide/src/config.md
```diff
@@ -2,7 +2,7 @@
 
 ## Configuration format
 
-You can configure maturin in `tool.maturin` section of [`pyproject.toml`](https://peps.python.org/pep-0518/#tool-table).
+You can configure maturin in `tool.maturin` section of [`pyproject.toml`](https://packaging.python.org/en/latest/specifications/pyproject-toml/#arbitrary-tool-configuration-the-tool-table).
```
mkniewallner, line 5: The current link is still valid, but PEPs usually recommend linking to living documentations if possible. I can drop this commit if you prefer the original link though.

Issue #3007, opened 2026-02-16
maturin 1.12.0+ regression, wheel incorrectly linked against musl instead of glibc


patryk4815:
### Bug Description

The maturin 1.12.2 RISC-V wheel (`maturin-1.12.2-py3-none-manylinux_2_31_riscv64.musllinux_1_1_riscv64.whl`) is built against musl libc instead of glibc, making it incompatible with standard glibc-based Linux distributions like Debian. This is a regression from version 1.11.5, which correctly uses glibc.

## Environment

- **Platform:** RISC-V 64-bit
- **OS:** Debian 13 (glibc-based)
- **Affected version:** maturin 1.12.2
- **Working version:** maturin 1.11.5

## Steps to Reproduce

1. Download the maturin 1.12.2 RISC-V wheel:
```bash
wget https://files.pythonhosted.org/packages/6f/ed/5680efbb1becb4f47da3ada8ea4eb6844d2fd91ae558287e1dd0871cb603/maturin-1.12.2-py3-none-manylinux_2_31_riscv64.musllinux_1_1_riscv64.whl
7z x maturin-1.12.2-py3-none-manylinux_2_31_riscv64.musllinux_1_1_riscv64.whl
```

2. Inspect the binary:
```bash
llvm-objdump -p ./maturin-1.12.2.data/scripts/maturin
file ./maturin-1.12.2.data/scripts/maturin
```

## Actual Behavior

**maturin 1.12.2** is linked against musl:

```
Dynamic Section:
  RPATH        $ORIGIN/../maturin.libs
  NEEDED       libgcc_s-998f13c4.so.1
  NEEDED       libc.so

$ file ./maturin-1.12.2.data/scripts/maturin
./maturin-1.12.2.data/scripts/maturin: ELF 64-bit LSB pie executable, UCB RISC-V, RVC, double-float ABI, version 1 (SYSV), dynamically linked, interpreter /lib/ld-musl-riscv64.so.1, not stripped
```

The binary expects the musl dynamic linker (`/lib/ld-musl-riscv64.so.1`), which is not available on glibc-based systems like Debian.

**Additional issue:** The wheel filename suggests it's tagged for both manylinux AND musllinux (`manylinux_2_31_riscv64.musllinux_1_1_riscv64`), which appears to be an invalid/malformed platform tag.

## Expected Behavior

**maturin 1.11.5** correctly links against glibc:

```
Dynamic Section:
  NEEDED          libgcc_s.so.1
[...]

PR #3009, opened 2026-02-16
ci: set crt-static for riscv64 and loongarch64 musl targets

messense:
Closes #3007

PR #3010, opened 2026-02-17
fix: enable static linking for loongarch64/riscv64 musl targets

wojiushixiaobai:
- Fix: #3007 

This ensures the generated binaries are fully statically linked.


wojiushixiaobai:
Fixed by #3009, closing this PR.

PR #3012, opened 2026-02-17
ci: use dev drive for Windows CI runners

### LINKED ISSUE #2274: Use dev drive for windows actions
Windows is significantly slower to compile and test than other platforms due to its slow io. We should use https://github.com/marketplace/actions/setup-dev-drive to speed up windows CI.

messense:
Use `samypr100/setup-dev-drive@v3` to create a ReFS dev drive on Windows runners and move `CARGO_HOME` and `RUSTUP_HOME` there for faster IO.

Closes #2274


messense:
Looks like it's much slower actually.

Issue #2564, opened 2025-04-23
Please support building wheels via source distribution in `maturin build` workflow


mgorny:
Currently, the `maturin build --sdist` workflow builds both source distribution and wheels directly from the project repository (usually a git repository). Unfortunately, sometimes this results in people building and publishing a source distribution that is effectively broken and cannot be used to build the project from source (for example, https://github.com/huggingface/xet-core/issues/255).

Could you please consider either modifying the `maturin build --sdist` workflow or providing an additional option that would follow the workflow of building the source distribution, then unpacking it and using it to build wheels? This way, Maturin would verify that the produced source distribution is actually usable, and reduce the likeliness of a broken source distributions being published.

messense:
I think this is better to be handled by a build frontend instead?

messense:
> To ensure that wheels from different sources are built the same way, frontends may call build_sdist first, and then call build_wheel in the unpacked sdist. But if the backend indicates that it is missing some requirements for creating an sdist (see below), the frontend will fall back to calling build_wheel in the source directory.

According to https://peps.python.org/pep-0517/#build-wheel

mgorny:
Well, `maturin build` command is a frontend command, isn't it? I mean, that's what people are using. Do you think a campaign to tell people to stop building source distributions and wheels via `maturin build` (the "obvious" way) and use, say, `build` instead would actually work?

tiran:
@mgorny and I have been discussing the issue in a private conversation. I was going to file a similar bug for `PyO3/maturin-action`. I don't have a strong opinion about `maturin build`. But I would like to see `PyO3/maturin-action` default to VCS -> sdist -> wheel workflow.

messense:
> But I would like to see `PyO3/maturin-action` default to VCS -> sdist -> wheel workflow.

That's my thought too, most builds are using `maturin-action`, we could have `maturin build` to do it directly, but it maybe simpler to do it in `maturin-action` first.

PR #3014, opened 2026-02-17
feat: build wheels from sdist with `--sdist` flag

messense:
Change `maturin build --sdist` to build the source distribution first, unpack it into a temporary directory, then build wheels from the unpacked sdist. This verifies that the produced sdist is complete and usable.

The same sdist-first workflow is also applied to `maturin publish` when `--no-sdist` is not specified.

Closes PyO3/maturin#2564

### DIFF
--- a/src/build_context.rs
+++ b/src/build_context.rs
@@ -40,6 +40,60 @@ use std::path::{Path, PathBuf};
 use tracing::instrument;
 use zip::DateTime;
 
+/// Unpacks an sdist tarball into a temporary directory and returns the path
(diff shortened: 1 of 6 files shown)


konstin:
I both really want this and wonder how much we reimplement `uv build` this way (which does that by default), but I'm not sure if there's even a way to expose the same options as we now have in maturin from `uv build` in a way that's user friendly.

messense:
Yeah I was also hesitate to implement this in maturin, but the actual code changes are small and provides real benefits, so don't hurt to have it.

PR #3023, opened 2026-02-20
fix: platform tag detection for Android targets

thunder-coding:
Follow up of a23a6b5824439580d5132a1524a891619b387886 (#2992)

It is recommended to go through the commit message linked above before trying to decode this. Since earlier Android was detected as Os::Linux, this extra condition was required, this should no longer be the case. So clean it up

A workaround is to set `_PYTHON_HOST_PLATFORM=android_{ANDROID_API_LEVEL}_{ANDROID_ARCH}` for build scripts that expect that wheels are named in a certain pattern

Issue originally reported in https://github.com/PyO3/maturin/pull/2992#issuecomment-3931492889

CC @robertkirkman

Additionally, I'd appreciate if we can get a patch release with this commit for maturin, as this issue is causing our CI to fail for Python packages for Termux

### DIFF
--- a/src/build_context.rs
+++ b/src/build_context.rs
@@ -655,31 +655,32 @@ impl BuildContext {
             (Os::Windows, Arch::X86) => "win32".to_string(),
             (Os::Windows, Arch::X86_64) => "win_amd64".to_string(),
             (Os::Windows, Arch::Aarch64) => "win_arm64".to_string(),
+            // Android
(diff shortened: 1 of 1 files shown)


robertkirkman:
Thanks, I can confirm that when this commit is in a release of Maturin, `scripts/run-docker.sh ./build-package.sh -I -f ruff` will work again.

Issue #3028, opened 2026-02-22
wrong abi3 tag for "conditional cargo features" enabled pyo3 abi3 feature


trim21:
### Bug Description

this work, build `-cp311-abi3-win_amd64.whl`

```toml
features = [
    # we need PyBuffer which is stable since 3.11
    "pyo3/abi3-py311"
    # { feature = "pyo3/abi3-py311", python-version = ">=3.11" },
]
```

```
❯❯ ~/proj/bencode-rs (master ~4) maturin build --locked --release -i '3.11'
📦 Including license file `LICENSE`
🍹 Building a mixed python/rust project
🔗 Found pyo3 bindings
⚠️  Warning: Failed to determine python platform
⚠️  Warning: Failed to determine python platform
🐍 Found CPython 3.11
📡 Using build options features from pyproject.toml
    Finished `release` profile [optimized] target(s) in 0.15s
📦 Built wheel for CPython 3.11 to C:\Users\Trim21\proj\bencode-rs\target\wheels\bencode_rs-0.0.8-cp311-cp311-win_amd64.whl
```

this doesn't:

```toml
features = [
    # we need PyBuffer which is stable since 3.11
    # "pyo3/abi3-py311"
    { feature = "pyo3/abi3-py311", python-version = ">=3.11" },
]
```

```text
❯❯ ~/proj/bencode-rs (master ~4) maturin build --locked --release -i '3.11'
📦 Including license file `LICENSE`
🍹 Building a mixed python/rust project
🔗 Found pyo3 bindings with abi3 support
⚠️  Warning: Failed to determine python platform
⚠️  Warning: Failed to determine python platform
🐍 Not using a specific python interpreter (automatically generating windows import library)
📡 Using build options features from pyproject.toml
    Finished `release` profile [optimized] target(s) in 0.08s
📦 Built wheel for abi3 Python ≥ 3.11 to C:\Users\Trim21\proj\bencode-rs\target\wheels\bencode_rs-0.0.8-cp311-abi3-win_amd64.whl
```

### Your maturin version (`maturin --version`)

maturin 1.12.4

### Your Python version (`python -V`)

no system python, with `pyo3/generate-import-lib` enabled

### Your pip version (`pip -V`)

no pip installed

### What bindings you're using

pyo3

### Does `cargo build` work?

- [ ] Yes, it works

### If on windows, have you checked that you aren't accidentally using unix path (those with the forward slash `/`)?

- [x] Yes

### Steps to Reproduce

[...]

PR #3029, opened 2026-02-22
Fix wrong abi3 tag for conditional cargo features enabled pyo3 abi3 feature

messense:
Detect abi3 from conditional features in pyproject.toml when not found in cargo metadata. Also filter abi3 interpreters by version >= abi3 minimum version so interpreters below the minimum get regular version-specific wheels.

Fixes https://github.com/PyO3/maturin/issues/3028


trim21:
should we resolve the final pyo3 features before we resolve if abi3 enabled?

PR #3054, opened 2026-03-01
Keep cargo build artifact at original path after staging

messense:
Previously `stage_artifact()` used `fs::rename` to move the artifact from the cargo output directory (e.g. `target/release/`) to a private staging directory (`target/maturin/`). This meant users could no longer find the built .so/.dylib/.dll at the standard cargo output location.

Now `stage_artifact()` still uses an atomic `fs::rename` into the staging directory (protecting against concurrent modification by cargo or rust-analyzer), then copies the staged file back to the original location via reflink (copy-on-write) when the filesystem supports it, falling back to a regular `fs::copy` otherwise. The copy-back is best-effort and does not fail the build. When fs::rename itself fails (e.g. cross-device), we fall back to reflink-or-copy directly.

On Linux, ioctl_ficlone does not preserve file permissions, so we explicitly copy them after the reflink. On macOS, clonefile preserves all metadata natively. This approach is adapted from uv's reflink_with_permissions implementation.

Refs: https://github.com/astral-sh/uv/blob/main/crates/uv-fs/src/link.rs
Refs: https://github.com/astral-sh/uv/issues/18181

Fixes https://github.com/PyO3/maturin/pull/2950#issuecomment-3979694037

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1692,6 +1692,7 @@ dependencies = [
  "pretty_assertions",
  "pyproject-toml",
  "python-pkginfo",
+ "reflink-copy",
  "regex",
  "rstest",
  "rustc_version",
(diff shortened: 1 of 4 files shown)


konstin:
Thanks for fixing this!

PR #3085, opened 2026-03-14
feat: add Profile-Guided Optimization (PGO) support

### LINKED ISSUE #1840: Supporting PGO builds from `maturin`
In `pydantic-core` we have been building with PGO for x86_64 Windows and Ubuntu for some time, and we're now also looking at GitHub's new M1 runners to run PGO optimization for arm macOS. https://github.com/pydantic/pydantic-core/pull/1063

This is in a sort of grey area where we can make it work with `maturin` by passing in the PGO `RUSTFLAGS` directly and then running normal `maturin` build steps. We could call it an orthogonal problem which `maturin` doesn't need to solve.

There are two reasons why I think it would be nice for `maturin` to have firsthand support for PGO builds:
- We can then make it easy for all Rust extension authors to PGO optimize their builds
- We can only pass one `--interpreter` to build with PGO, which makes our build matrix very large (M platforms, N Python versions). It would be nice to be able to pass `--interpreter 3.7 3.8 ...` with all versions to make use of better build caching and reduce the number of jobs.

To have support for PGO, I think `maturin` would need to:
1. build the PGO-instrumented wheel for a given interpreter
2. install that wheel into a temporary virtual environment and run some step to generate PGO samples (we'd need configuration for this)
3. build the PGO-optimized wheel using the samples
... and repeat the above for each interpreter in turn.

I think all the configuration that would be needed is the set of packages to install into the temporary virtual environment and then a script to run to generate the PGO sample. The script could even be responsible for installing the dependencies, reducing to a single configuration entry, but then it probably needs to be a shell script and it gets a little harder to be cross-platform.

There is maybe an argument to leave all of this functionality to `maturin-action`. A nice thing about putting this directly in `maturin` is that the PGO build could then also be done locally, e.g. to benchmark it.

cc @konstin I took a quick look at the `ruff` workflow and it looks like you don't make use of PGO builds, is that something you tried and found it wasn't worth it, or just not ever looked into? In `pydantic-core` we won about 20-30% speedup from PGO, so potentially there is a nice `ruff` performance win to celebrate sitting on the table 😄

messense:
Add `--pgo` CLI flag and `pgo-command` config in `[tool.maturin]` to enable PGO builds via a three-phase pipeline:

1. Build instrumented wheel with `-Cprofile-generate`
2. Install in temp venv and run the user's pgo-command to collect `.profraw` files
3. Merge profiles with `llvm-profdata` and rebuild with `-Cprofile-use`

For non-abi3 PyO3 builds, each interpreter gets its own instrument-train-optimize cycle via `build_single_pyo3_wheel()`, since different Python versions produce different compiled code through PyO3's build-time configuration. For abi3/cffi/uniffi/bin builds, a single PGO pass suffices.

Key components:
- `PgoContext` in `src/pgo.rs`: manages temp dirs, `llvm-profdata` resolution, venv creation, dependency installation, and profile merging
- `PgoPhase` enum injected into `BuildContext.pgo_phase`, read in `compile.rs` to add rustc flags
- `clone_for_pgo()` helper to reduce context-setup boilerplate
- `build_single_pyo3_wheel()` extracted from `build_pyo3_wheels()` to share the compile-audit-write pipeline with the PGO path

Closes #1840

### DIFF
--- a/src/build_context/builder.rs
+++ b/src/build_context/builder.rs
@@ -29,6 +29,7 @@ pub struct BuildContextBuilder {
     editable: bool,
     sdist_only: bool,
     pyproject_toml_path: Option<PathBuf>,
+    pgo: bool,
 }
 
 impl BuildContextBuilder {
@@ -39,6 +40,7 @@ impl BuildContextBuilder {
             editable: false,
             sdist_only: false,
             pyproject_toml_path: None,
+            pgo: false,
         }
     }
 
@@ -62,6 +64,11 @@ impl BuildContextBuilder {
         self
     }
 
+    pub fn pgo(mut self, pgo: bool) -> Self {
+        self.pgo = pgo;
+        self
+    }
+
     #[instrument(skip_all)]
     pub fn build(self) -> Result<BuildContext> {
         let Self {
@@ -70,6 +77,7 @@ impl BuildContextBuilder {
             editable,
             sdist_only,
             pyproject_toml_path: explicit_pyproject_path,
+            pgo,
         } = self;
         build_options.compression.validate();
         let ProjectResolver {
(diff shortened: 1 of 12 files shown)


ppmpreetham:
Any clue on how to use this with Github Actions?
Thanks!

PR #3113, opened 2026-03-31
Support pyo3 abi3t features on Python3.15 and PyO3 0.29

ngoldbaum:
Adds support for the new abi3t ABI in Python 3.15. See https://pyo3.rs/main/building-and-distribution.html?highlight=abi3t#py_limited_apiabi3abi3t for more info from the pyo3 side of things.


ngoldbaum:
@messense do you have any opinions about how tests for abi3t support should look? My first instinct would be to follow and generalize the existing tests for abi3. Does that sound like a good plan, or do you have large test refactorings in mind separate from this?

messense:
Sorry missed the reply in the issue, follow and generalize the existing tests for abi3 sounds good to me.

ngoldbaum:
> Sorry missed the reply in the issue, follow and generalize the existing tests for abi3 sounds good to me.

Sorry for the delay over here. I just added some new tests. Since abi3t support requires Python 3.15 or newer, I needed to also add some utilities to introspect the Python used by the test runner.

ngoldbaum:
I don't think there are any tests for Python 3.15 yet. @messense what do you think about replacing `3.14` and `3.14t` in `test.yml` with `3.15-dev` and `3.15t-dev` in this PR?

ngoldbaum:
I should also note that the new test crates depend on my PR branch for https://github.com/pyo3/pyo3/pull/5807 and I'll need to update them after that PyO3 PR is merged and PyO3 does a release. Not quite sure how to sequence landing everything.

ngoldbaum:
>  @messense what do you think about replacing 3.14 and 3.14t in test.yml with 3.15-dev and 3.15t-dev in this PR?

Never mind, I tried and it will require some more work to get that stable. See the CI run associated with https://github.com/PyO3/maturin/pull/3113/commits/ee98a088c8c185c07c7ea2f326a256e43e7e5fa5.

src/bridge/mod.rs
```diff
@@ -369,6 +390,31 @@ impl BridgeModel {
...
     }
 
+    /// Check whether an abi3 build should be enabled for a specific interpreter.
+    ///
+    /// Returns `true` only when the bridge model has stable abi support **and**
+    /// the given interpreter supports the stable ABI **and** meets the abi3
+    /// minimum version. Version‑specific fallback builds (e.g. Python 3.10 when
+    /// abi3 targets ≥ 3.11) return `false` so that `Py_LIMITED_API` is not
+    /// defined and interpreter‑specific linker names are used.
+    pub fn is_abi3_for_interpreter(&self, interpreter: &PythonInterpreter) -> bool {
```
davidhewitt: Where does this function end up being needed over `is_stable_abi_for_interpreter`?

ngoldbaum: Good call - I think I needed it in an earlier version of this PR but I can delete it now and simplify things a lot.

ngoldbaum: Oh no, that's wrong. It's load-bearing in the implementation of `configure_pyo3_env`, which ends up generating a pyo3 configuration file using the `abi3` key. I probably need to detect the pyo3 version and generate an appropriate config file with a `target_abi` key if I find a new-enough pyo3.

ngoldbaum: I added a new `use_target_abi` argument to `configure_pyo3_env` and set up maturin to generate a configuration file with an `abi3` or `target_abi` argument based on the detected PyO3 version. This let me delete `is_abi3_for_interpreter`.

ngoldbaum:
OK, I think things are mostly stable here running the tests on 3.15 and 3.15t. The Windows 11 ARM failure is due to permissions issues on the runner during the CFFI build, which will go away in the future when CFFI uploads 3.15 wheels.

EDIT: spoke too soon, those Windows pypy failures look real.

ngoldbaum:
The maturin CI is still hitting some issues with cross-compilation and pypy builds so I'll need to fix that and add more tests for pyo3-build-config for those cases as well.

ngoldbaum:
OK, I think CI is stabilized. Not sure what's up with the remaining Mac pypy3.11 failures, but I think they're caused by a change in PyO3 unrelated to abi3t support (https://github.com/PyO3/pyo3/pull/6041).

ngoldbaum:
Also I should say that I had to drop 3.13t from the interpreters considered for interpreter discovery, since PyO3 dropped 3.13t support.

ngoldbaum:
https://github.com/PyO3/pyo3/pull/6078 fixes the PyPy linker issue causing the remaining test failures.

ngoldbaum:
Sorry for all the noise the past few days and the substantial growth of this PR. I realized that I wasn't actually running the integration tests locally on Python 3.15, but once I realized that the maturin CI uses `cargo nextest` and figured out I need to edit the test crates to point at my PyO3 PR, I discovered a number of additional issues that needed fixing in PyO3 and in Maturin.

That said, I think the implementation is ready to review now and is only waiting for PyO3 to do a 0.29 release include abi3t support.

There probably need to be docs as well, I will look at that next week I think.

messense:
I'm holding off review this PR until it's out of draft state, ping me if I missed any notifications, thanks!

ngoldbaum:
I think I understand the cause of the random Windows failures, see https://github.com/pypa/pip/issues/9034. In short, pip's cache isn't safe to concurrently write to on Windows like maturin's CI does. I pushed a commit disabling pip caching, which hopefully doesn't negatively impact CI times too much.

ngoldbaum:
This is still not ready to merge because PyO3 hasn't yet done a release and I'll need to update the test crates one final time. That said, I just added some documentation and marked this ready for review because it's otherwise totally ready.

IMO it'd be a good idea to coordinate merging this and quickly cutting a release with the PyO3 0.29 release.

messense (approved):
LGTM, thanks, let's wait for pyo3 0.29 release then update and merge this.

ngoldbaum:
Updated to use PyO3 0.29 in the test crates. Everything passes locally but let's see what CI says.

PR #3131, opened 2026-04-10
fix: redirect tracing output to stderr to avoid breaking PEP 517

chrisburr:
Since v1.13.0, `pip install --no-build-isolation` has been broken when `RUST_LOG` is set. The tracing subscriber added in v1.13.0 uses `fmt::layer()`, which defaults to stdout. PEP 517 uses stdout as a structured channel (pip reads the last line to find metadata directory names) so tracing output corrupts the path.

This PR fixes the issue by:

* adding `.with_writer(std::io::stderr)` to the tracing subscriber setup
* adding a regression test verifying tracing output stays on stderr

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -320,6 +320,10 @@ fn setup_logging(verbose: u8) -> Result<()> {
         .context("Invalid RUST_LOG directives")?;
 
     let logger = tracing_subscriber::fmt::layer()
+        // fmt::layer() defaults to stdout; redirect to stderr so that
+        // structured stdout output (e.g. PEP 517 directory names) is not
+        // polluted by tracing spans.
+        .with_writer(std::io::stderr)
         // Avoid showing all the details from the spans
         .compact()
         // Log the timing of each span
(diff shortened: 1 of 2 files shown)


messense:
Thanks for the PR, I thought tracing subscriber was added years ago?
https://github.com/PyO3/maturin/blame/a025a1f0046314589add2b20e158cb303646867c/src/main.rs#L322-L332

messense (approved):

chrisburr:
Thanks for the quick review.

> I thought tracing subscriber was added years ago?

Maybe I've misdiagnosed exactly what the origin of the break is. The symptom which triggered this was that this CI job started failing with the latest maturin release: https://github.com/prefix-dev/pixi/actions/runs/24230829818/job/70744566427

Regardless, I think this is still the correct fix.

messense:
It would be great if you can test this so we can verify it actually fixes the issue (maybe dropping `RUST_LOG` makes it pass is a good sign?). But this PR can be merged regardless of the that. Thanks!

chrisburr:
I'm fairly confident this fixes the issue as dropping `RUST_LOG` fixes the pixi test suite where I originally saw this issue.

PR #3138, opened 2026-04-14
docs: add manifest-path config for src layout

plusls:


MukundaKatta:
+1 to documenting this. I used manifest-path yesterday for a workspace
where the bindings crate sits under crates/foo-py/ and the workspace root
is the maturin-managed Python project. Worked first try once I knew the
field existed. The previous failure mode was a confusing "no Cargo.toml"
error from the workspace root, which a one-line note in the docs would
solve.

PR #3143, opened 2026-04-17
fix: honor python-version when inferring abi3 from conditional features

### LINKED ISSUE #3142: 1.13: conditional features python-version predicate ignored, abi3 applied unconditionally
### Bug Description

Since maturin 1.13.0, the `python-version` predicate in conditional `[tool.maturin].features` is ignored. The abi3 feature is applied even when the target interpreter does not satisfy the condition, causing a build failure.

**pyproject.toml:**
```toml
[tool.maturin]
module-name = "bencode_rs._bencode"
python-source = "py"
features = [
  { feature = "pyo3/abi3-py311", python-version = ">=3.11" },
]
```

**Cargo.toml:**
```toml
[features]
default = []

[dependencies]
pyo3 = { version = "0.28.3", features = ['generate-import-lib'] }
```

### Expected Behavior

When running `maturin develop` or `maturin build --interpreter '3.10'` with Python 3.10, the conditional feature `pyo3/abi3-py311` should **not** be enabled (since `3.10 < 3.11`), and a normal `cp310-cp310` wheel should be produced.

This worked correctly in maturin **1.12.6**.

### Actual Behavior (maturin 1.13.1)

```
$ maturin develop --uv --no-default-features
📦 Including license file `LICENSE`
🔗 Found pyo3 bindings with abi3 support
🐍 Found CPython 3.10 at .venv/bin/python
📡 Using build options features from pyproject.toml
💥 maturin failed
  Caused by: None of the found Python interpreters (CPython 3.10) are compatible with the abi3 minimum version (>= 3.11). Please install a compatible Python interpreter.
```

The `python-version = ">=3.11"` condition is being ignored — maturin applies `pyo3/abi3-py311` unconditionally, then rightfully rejects Python 3.10 as incompatible with abi3-py311. This is a regression in the interpreter resolution refactor (#3032).

### Reproducer

1. Use the `pyproject.toml` / `Cargo.toml` shown above
2. Create a virtualenv with Python 3.10
3. Run `maturin develop` or `maturin build --interpreter '3.10'`

Works with maturin 1.12.6, fails with 1.13.0 and 1.13.1.

### Environment

- maturin 1.13.1
- pyo3 0.28.3
- Python 3.10
- Ubuntu 24.04 (GitHub Actions) and Windows

### Related

- #3028 (previous issue about conditional features + abi3 tags, fixed in 1.12.4)
- #3027 (the PR that implemented the fix for #3028)
- #3032 (interpreter resolution refactor in 1.13.0, likely regression source)

lawrence3699:
Fixes #3142.

## Before
When `[tool.maturin].features` contained a conditional `pyo3/abi3-py311` entry, bridge detection treated it as active before checking whether the selected interpreter actually matched `python-version`.

That made `maturin build -i python3.10` and `maturin develop` incorrectly go down the abi3 path and fail with:

`None of the found Python interpreters ... are compatible with the abi3 minimum version (>= 3.11)`

## After
Conditional `pyo3`/`pyo3-ffi` features only contribute to abi3 bridge inference when they match a known interpreter environment.

This keeps the positive case working for Python 3.11, but no longer forces abi3 for Python 3.10.

## Validation
- `cargo fmt --all --check`
- `cargo test --test run conditional_abi3_python_version_predicate -- --nocapture`
- `cargo test --test run abi3_python_interpreter_args -- --nocapture`
- `cargo test --lib test_find_bridge_ -- --nocapture`
- manual repro from the issue pattern now builds a `cp310-cp310` wheel instead of failing


messense:
Thanks for the PR, the change is a bit over-engineered IMO, I took a slightly different approach in #3144.

PR #3152, opened 2026-04-23
Musllinux oci image

### LINKED ISSUE #3151: musllinux variant of the oci image
I build my manylinux wheels using the `ghcr.io/pyo3/maturin` oci image. Would be cool if there was a variant of the image that supports musllinux. I naively tried to supply `--compatibility musllinux_1_1` as argument which resulted in broken wheels.
It would be even better if the same image could be used for manylinux and musllinux, but i don't think that's possible.

Joshix-1:
Fixes: #3151 

Ran at: https://github.com/Joshix-1/pyo3-maturin/pkgs/container/pyo3-maturin


Joshix-1:
Built some wheels using the oci image. auditwheel adds libgcc_s-a04fdf82.so.1 to them, but they seem to work.

Joshix-1:
merged the two Dockerfiles into one

messense (approved):
Thanks!

PR #3161, opened 2026-04-26
develop: fail loudly when pip leaves a stale `~` install behind (#1922)

### LINKED ISSUE #1922: When the .pyd file is in use and Maturin develop mentions the Pip warning, that needs to be much more obvious that that's really bad.
### Bug Description

If you get the following warning: ```
⚠️ Warning: pip raised a warning running ["-m", "pip", "--disable-pip-version-check", "install", "--no-deps", "--force-reinstall", "C:\\blahblah.whl"]:
WARNING: Failed to remove contents in a temporary directory 'C:\blahblah\~ackage.
  You can safely remove it manually.
```
This should be more clear that you MUST remove it manually before running your python process again. To me this reads as "hey, pip warned us, but it's not a big deal, just whenever you get to it go clean up these old temp folders".

But really it is more like "if you do not remove this folder before running your python process it will not use your new Rust code but the old cached code from the previous version. Go remove it right now"

I think it should be an Error honestly, not a warning. The build has essentially failed to be useful in at least one big use case.

1.4.0

3.9.10

23.3.2

pyo3

### Does `cargo build` work?

- [X] Yes, it works

- [X] Yes

### Steps to Reproduce

1. Make a simple pyo3 project with an infinite loop. Maturin develop --release and start the python code calling the pyo3 code.
2. While the python code is still running, change the Rust to remove the infinite loop (or add a new print or something).
3. With the code still running, do maturin develop --release again. You'll see the warning above.
4. Kill and restart your python code. In at least some cases (may not be every time?) your new rust code will not be executed, but your original instead.

### PR DESCRIPTION
**Pull request by pratyush618**
## Summary

Fixes #1922.

When a Python process holds the installed `.pyd` open on Windows, pip "salvages" the install by renaming the old install directory to a sibling whose basename is prefixed with `~` and writes the new wheel files in place. If pip can't later remove that sibling, it emits a soft warning of the form

```
WARNING: Failed to remove contents in a temporary directory 'C:\…\~ackage'.
  You can safely remove it manually.
```

and exits with status 0. The leftover keeps the **old** compiled code, so a subsequent `import` can silently load stale code instead of the freshly built module. The user in #1922 was surfaced this as a generic `⚠️ Warning`, which made it look benign — meanwhile their next Python run was still executing the previous build.

This PR detects exactly that case in `install_wheel` and converts it into a hard error with a three-step recovery (stop Python, delete the leftover path, re-run `maturin develop`). Other pip stderr keeps the prior warning behavior.

## Notes

- Detection is gated on the *basename* starting with `~`, so unrelated `Failed to remove contents in a temporary directory '/tmp/pip-build-xyz'` cleanup hiccups still surface as a warning rather than failing the install.
- Same `static Lazy<Regex>` idiom already used by `parse_direct_url_path` two functions below.
- Linux/macOS hit this code path too if pip ever produces the same warning shape, but in practice the `~` rename only happens on Windows.

## Test plan

- [x] `cargo build` clean
- [x] `cargo clippy --lib --no-deps -- -D warnings` clean
- [x] `cargo fmt --check src/develop/mod.rs` clean
- [x] New unit tests in `develop::tests` pass:
  - Windows-style backslash path with `~`-prefixed basename → detected
  - POSIX path with `~`-prefixed basename → detected
  - Same warning text without `~`-prefixed basename → **not** detected (stays a warning)
  - Unrelated pip warning (`pip version 20.0.0; …`) → not detected

### DIFF
--- a/src/develop/mod.rs
+++ b/src/develop/mod.rs
@@ -207,17 +207,58 @@ fn install_wheel(
         );
     }
     if !output.stderr.is_empty() && install_backend.stderr_indicates_problem() {
+        let stderr = String::from_utf8_lossy(&output.stderr);
+        let stderr = stderr.trim();
+        if let Some(leftover) = detect_pip_in_use_leftover(stderr) {
+            bail!(
+                "{backend} could not overwrite the installed extension module because it is in use \
+                 (most often by a Python process that has already imported the package).\n\
+                 {backend} salvaged the install by renaming the old location to a leftover \
+                 directory, but it could not delete that leftover — and the leftover still holds \
+                 the OLD compiled code. If a Python process imports the package before the \
+                 leftover is removed, the import system can resolve the stale copy and silently \
+                 run pre-build code.\n\n\
+                 To recover:\n  \
+                 1. Stop every Python process that imported this package (this releases the file handle).\n  \
+                 2. Delete the leftover directory: {leftover}\n  \
+                 3. Re-run `maturin develop`.\n\n\
+                 Original {backend} output:\n{stderr}",
+                backend = install_backend.name(),
+            );
+        }
         eprintln!(
             "⚠️ Warning: {} raised a warning running {:?}:\n{}",
             install_backend.name(),
             &cmd.get_args().collect::<Vec<_>>(),
-            String::from_utf8_lossy(&output.stderr).trim(),
+            stderr,
         );
     }
     if let Err(err) = configure_as_editable(build_context, python, install_backend) {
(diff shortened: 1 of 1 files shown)


pratyush618:
Good catch — fixed in 5adf26d6. `detect_pip_in_use_leftover` now runs regardless of backend, so `maturin develop --uv` also fails loudly on the stale leftover. The generic warning print stays gated on `stderr_indicates_problem()` since uv writes regular logs to stderr.

messense (approved):

PR #3162, opened 2026-04-26
Add SLSA generic generator workflow

arvkvk7-crypto:
This workflow generates SLSA provenance files for projects, satisfying level 3 requirements. It includes steps for building artifacts and generating provenance subjects.
.github/workflows/generator-generic-ossf-slsa3-publish.yml


messense:
We are already using GitHub Artifact Attestations.

PR #3165, opened 2026-04-28
fix: support pixi-managed virtualenvs in maturin develop

### LINKED ISSUE #2713: Add support for pixi venvs
Pixi, the modern alternative to conda, is written in Rust, with uv as a library, but not an installed package.
So, unlike in conda's venvs, pixi's venvs aren't currently compatible with maturin:

[asciicast](https://asciinema.org/a/XGwHlSLR3vOBu2HHSUt1mBs9Q)

eyupcanakman:
Fixes #2713.

Pixi sets `CONDA_PREFIX` and creates an environment without `pyvenv.cfg`, so `is_uv_venv` returned false. `maturin develop` then fell back to pip. Pixi does not preinstall pip, so users saw "Failed to find pip" on every invocation.

`is_pixi_venv` checks for the `conda-meta/pixi_env_prefix` marker file (written by pixi 0.43+) and routes detected envs through the same uv backend path used for uv-created venvs.

### DIFF
--- a/src/develop/mod.rs
+++ b/src/develop/mod.rs
@@ -13,7 +13,9 @@ use crate::{
 use anyhow::{Context, Result, anyhow, bail, ensure};
 use cargo_options::heading;
 use fs_err as fs;
-use install_backend::{InstallBackend, check_pip_exists, find_uv_bin, find_uv_python, is_uv_venv};
+use install_backend::{
+    InstallBackend, check_pip_exists, find_uv_bin, find_uv_python, is_pixi_venv, is_uv_venv,
+};
 use once_cell::sync::Lazy;
 use regex::Regex;
 use std::path::Path;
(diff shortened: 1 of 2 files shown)


src/develop/mod.rs
```diff
@@ -404,15 +363,15 @@ pub fn develop(develop_options: DevelopOptions, venv_dir: &Path) -> Result<()> {
                 anyhow!("Expected `python` to be a python interpreter inside a virtualenv ಠ_ಠ")
             })?;
 
-    let uv_venv = is_uv_venv(venv_dir);
+    let uv_venv = is_uv_venv(venv_dir) || is_pixi_venv(venv_dir);
```
messense, line 366: I don't use pixi so I'm not sure this is correct or not, does pixi env ships `uv` in `$PATH` by default?

eyupcanakman: No, pixi itself doesn't install `uv` into the env, but it also doesn't sanitize the inherited `$PATH`, so `find_uv_bin()` succeeds when the user has a global `uv`. Without that, the fall-through to `check_pip_exists` errors with a `--uv` hint.

messense: So user has to install uv separately to make pixi venv work? How can we make it work out of the box?

messense (approved):
LGTM. Detection via the `conda-meta/pixi_env_prefix` marker is robust and reuses the existing uv backend path cleanly. Tests cover the right boundaries (positive, plain venv, vanilla conda).

One tiny nit (non-blocking): the variable `uv_venv` in `develop()` now also covers pixi; renaming to something like `uv_compatible_venv` would read better, but feel free to ignore.

src/develop/mod.rs
messense: Did some research and I think it won't work out of the box.

PR #3168, opened 2026-05-01
Restore symlinked Python source traversal

Bben01:
## Summary

Restore traversal of symlinked directories when maturin collects Python package sources for wheels and sdists.

I dont know if it was removed intentional or not. If it was, I could add some bool to the pyproject.toml that could toggle this behavior (and set the current default to false)

## Why

In our team we package multiple Python packages that share some common code. We use symlinks for those shared directories and files so the source is maintained in one place instead of duplicated across package trees.

This worked previously (from what i remember - 1.7.0), but recent behavior no longer followed symlinked directories while building a pyo3 Rust + Python layout. Symlinked files were still included, but files inside symlinked directories were omitted from the wheel.

## Fix

Use `follow_links(true)` when walking Python package sources for both wheel and sdist generation.

## Validation

- Confirmed the new regression test fails before the fix: `linked_dir/nested.py` is missing from the wheel.
- Confirmed it passes after the fix with `cargo test --test run -- pyo3_mixed_py_subdir_includes_symlinked_python_files`.
- Ran `cargo fmt --all` and `git diff --check`.

## AI disclosure

Fix was written by codex 5.5

### DIFF
--- a/src/module_writer/mod.rs
+++ b/src/module_writer/mod.rs
@@ -10,7 +10,7 @@ use ignore::WalkBuilder;
 use indexmap::IndexMap;
 use itertools::Itertools as _;
 use normpath::PathExt as _;
-use tracing::{debug, trace};
+use tracing::{debug, warn};
 
 use crate::Metadata24;
 use crate::PyProjectToml;
@@ -19,6 +19,7 @@ use crate::archive_source::FileSourceData;
 use crate::archive_source::GeneratedSourceData;
 use crate::project_layout::ProjectLayout;
 use crate::pyproject_toml::Format;
+use crate::util::is_symlink_loop_error;
 
 pub(crate) mod glob;
 #[cfg(test)]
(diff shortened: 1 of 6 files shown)


messense:
@Bben01 Could you clarify where your shared Python sources live relative to the project checkout?

The current PR follows symlinked Python package directories, but there is a safety tradeoff: following symlinks can package files outside the project if a link points somewhere broad or unexpected. One possible rule is to allow symlink targets that stay inside the same git repository/worktree, but reject targets outside it.

Would that match your setup, or are your shared Python sources outside the git repository that contains the maturin project?

Bben01:
Yes, the sources are inside a big monorepo, and we build from that multiple packages, so it could work.

I currently don't have access to a computer but will do it as soon as possible.

If you prefer I could just make it explicit using a config flag in the pyproject.toml

PR #3206, opened 2026-05-28
Add support for finding free-threaded interpreters for --find-interpreters

ngoldbaum:
This is partially extracted from #3113.

Currently, free-threaded interpreters are not always considered for `--find-interpreters`. It is included on Unix but not on Windows. This means that users need to explicitly do another build to produce 3.14t wheels. This also contributes to https://github.com/PyO3/maturin-action/issues/368 and I'm planning to do a followup over there to fix that once this is merged.

With this PR, `--find-interpreters` always picks up 3.14t and newer. I intentionally excluded 3.13t because PyO3 0.29 will refuse to build wheels for 3.13t. This also follows manylinux, which recently dropped 3.13t from the images.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-* `--find-interpreter` now discovers free-threaded CPython interpreters (`python3.14t` and newer). The experimental 3.13t is no longer built by default; build it explicitly with `-i python3.13t`.
+* `--find-interpreter` now discovers free-threaded CPython interpreters (`python3.14t` and newer) on all platforms. The experimental 3.13t is no longer built by default; build it explicitly with `-i python3.13t`.
 
 ## 1.13.3
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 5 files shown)


Changelog.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+* `--find-interpreter` now discovers free-threaded CPython interpreters (`python3.14t` and newer). The experimental 3.13t is no longer built by default; build it explicitly with `-i python3.13t`.
```
ngoldbaum:
```suggestion
* `--find-interpreter` now discovers free-threaded CPython interpreters (`python3.14t` and newer) on all platforms. The experimental 3.13t is no longer built by default; build it explicitly with `-i python3.13t`.
```

ngoldbaum (commented):
Spotted a minor improvement for the changelog entry

messense (approved):

ngoldbaum:
Looks like something broke unrelated to this change yesterday. I'll try to dig in later to unbreak CI...

ngoldbaum:
It looks like conda updated pip from 26.0.1 to 26.1.1 and 26.1.0 dropped support for Python 3.9.

PR #3208, opened 2026-05-29
Fix python versions used in integration_pyo3_mixed_conda

ngoldbaum:
The default anaconda channel shipped a pip version that is broken on Python 3.9, so I bumped the minimum version on Python 3.10.

I also bumped the maximum version as well.

This fixes the CI failures seen in https://github.com/PyO3/maturin/pull/3206 when @messense tried to merge it.


ngoldbaum:
The failures look like unrelated network flakiness or test flakiness.

messense (approved):

PR #3213, opened 2026-06-03
Don't depend on CFFI on PyPy

ngoldbaum:
Per [the CFFI docs](https://cffi.readthedocs.io/en/latest/installation.html#installation-and-status), PyPy vendors CFFI and installing it and requiring it is unnecessary. This is particularly annoying because cffi doesn't ship PyPy wheels, so it triggers an unnecessary C build at install time.

Also see, for example, the way [cryptography handles this](https://github.com/pyca/cryptography/blob/ba936e527bc4e3948c338a7bde54c3bd81957c61/pyproject.toml#L8).

I noticed this was an issue in the CI for #3113, which saw failures related to the unnecessary CFFI builds failing.


konstin (approved):
Good catch!

PR #3226, opened 2026-06-15
Fix issues around crates enabling abi3 and abi3t features

ngoldbaum:
This fixes issues reported by @2bndy5 over at https://github.com/PyO3/maturin-action/issues/368#issuecomment-4696265095.

Unfortunately, I completely skipped adding tests here for crates that mix abi3 and abi3t features and introduced some pretty serious bugs.

This PR makes it structurally impossible to generate abi3.abi3t wheels targeting ABIs before Python 3.15.  Stable ABI selection now happens after interpreter resolution and chooses at most one stable ABI family per build: abi3t only when a compatible CPython 3.15+ interpreter is present, otherwise abi3 when available. Interpreters that do not support the selected stable ABI family fall back to version-specific wheels, e.g. CPython 3.14t builds cp314-cp314t instead of cp314-abi3.abi3t.

Updates the docs I added for abi3t support to describe the actual behavior of Maturin. I considered making it possible to build two stable ABI wheels in the same call using a GIL-enabled interpreter but decided that required too many internal code changes to change the `stable_abi` member of the PyO3 bridge struct from an Option wrapping a single value into a vec! that might contain multiple values.

The bulk of the PR is new tests, including new unit tests and integration tests. The current python versions used in CI capture interesting behaviors in this test: Python 3.9 should produce a py38-abi3 wheel, Python 3.15 and 3.15t should produce an abi3t wheel.


messense:
PyPy tests are failing.

ngoldbaum:
Looks like pypy is fixed in the latest commit.

guide/src/bindings.md
```diff
@@ -14,25 +14,41 @@ maturin automatically detects pyo3 bindings when it's added as a dependency in `
...
+abi3t = ["pyo3/abi3t-py315"]
+```
+
+A single maturin build selects one stable ABI family. If you want to publish
+both a GIL-enabled `abi3` wheel and an `abi3t` wheel, run two wheel builds
+explicitly, with different Cargo features and compatible interpreters:
+
+```console
+maturin build --no-default-features --features abi3 --interpreter python3.10
+maturin build --no-default-features --features abi3t --interpreter python3.15t
```
ngoldbaum: Reading over this again, I think I can just revert this guidance and tell people to add both features.

guide/src/bindings.md
```diff
@@ -14,25 +14,36 @@ maturin automatically detects pyo3 bindings when it's added as a dependency in `
...
-An `abi3-py310` wheel supports all GIL-enabled Python
-versions from Python 3.10 to Python 3.14 and the `abi3.abi3t` wheel supports
-Python 3.15 and all newer versions of CPython.
+A single maturin build selects one stable ABI family. If you want to publish
+both a GIL-enabled `abi3` wheel and an `abi3t` wheel, run separate wheel builds
+explicitly, with compatible interpreters:
+
+```console
+maturin build --interpreter python3.10
+maturin build --interpreter python3.15t
```
2bndy5, line 31: I had problems with this `--interpreter` value/format. The example in question simply does not work for Windows builds. I found it worked better (cross-platform -- `windows-latest`, `ubuntu-latest`, `macos-15-intel`, `macos-latest`) when I used that same version syntax passed to the `actions/setup-python` action:

```yml
strategy:
  matrix:
    # note, '3.11' is min available build on `windows-11-arm` runner
    include:
      - python-version: '3.10'
        pyo3-feature: 'pyo3/abi3-py310'
      - python-version: '3.15t'
        pyo3-feature: 'pyo3/abi3t-py315'
steps:
  - uses: actions/checkout@v6
  - uses: actions/setup-python@v7
    with:
      python-version: ${{ matrix.python-version }}
      # 3.15 is not stable yet, so ...
      allow-prereleases: ${{ startsWith(matrix.python-version, '3.15') }}
  - uses: pyo3/maturin-action@v1
    with:
      args: >-
        --release
        --out dist
        --interpreter ${{ matrix.python-version }}
        --features ${{ matrix.pyo3-feature }}
```


FYI, passing the `${{ steps.python-setup.outputs.python-path }}` is a _bad_ idea for Linux builds (if pyo3/maturin-action `manylinux` input is not `'off'`).

ngoldbaum: Did you have those problems using the version of Maturin from this PR or the released version?

2bndy5: I checked the CI logs...
> Installing 'maturin' from tag 'v1.14.0'

ngoldbaum: So then I think this PR will fix things for you, there is Windows CI running on this PR and the invocation in the docs is what the integration test I added here does.

2bndy5: But the [windows CI logs](https://github.com/PyO3/maturin/actions/runs/27628589769/job/81697643613?pr=3226) show:
```sh
python_version=$(echo 3.15t-dev | sed -e s/-dev//)
echo "PYTHON_VERSION=$python_version" >> "${GITHUB_ENV}"
```
where that output is
```sh
python_version=3.15t
echo PYTHON_VERSION=3.15t
```
and then it is passed to `maturin build`:
```sh
cargo run build -i $PYTHON_VERSION -m test-crates/pyo3-pure/Cargo.toml
```

This coincides with my observation and seems contrary to the example in question.

ngoldbaum: Ah, I see what you're getting at now. Yes, the CI over here doesn't set `allow-prereleases` but I don't think that matters for the question you have? The maturin integration tests don't know about setup-python or its naming conventions.

So the place to handle the problem you're having is probably over in maturin-action. Maybe in the PR I have open over there?

https://github.com/PyO3/maturin-action/pull/453

Also with this PR merged you won't need to explicitly pass in abi3 and abi3t features at build time anymore. You can just enable both features, like in the test crates in this PR.

I'm going to bed now so I won't deal with subsequent replies here until tomorrow.

2bndy5: Ah, I didn't think there was any pre-processing in `pyo3/maturin-action`'s `args` input. I'll look at that PR too.

> Also with this PR merged you won't need to explicitly pass in abi3 and abi3t features at build time anymore. You can just enable both features, like in the test crates in this PR.

The way this patch reads, I think I still need to specify each feature separately if I want a abi3 wheel and a abi3t wheel. My understanding is that it prevents building abi3t wheels for any version of python prior to 3.15t.

Anyway, I bid you well-deserved sweet dreams! Thanks again.

ngoldbaum: > The way this patch reads, I think I still need to specify each feature separately if I want a abi3 wheel and a abi3t wheel. My understanding is that it prevents building abi3t wheels for any version of python prior to 3.15t.

One last comment:

PyO3 is perfectly happy to let you enable both features.


If you enable, say abi3-py310 and abi3t-py315, the wheel you get out of maturin depends on the Python version used to build the wheel:

* 3.14 or older Gil-enabled build -> abi3 wheel targeting the Python 3.10 limited API
* 3.14t -> cp314t wheel
* 3.15 or newer, both builds -> abi3.abi3t wheel targeting the 3.15 limited API

The idea is you get a working build no matter what, with abi3t  preferred and falling back to abi3 on the GIL-enabled build where abi3t can't work and a version-specific wheel on 3.14t whhere neither stable ABI is supported.

Please try it yourself by checking out this PR locally, building and installing it in your Python environment, and trying it with a test crate you come up with or one of the test crates I added in this PR.

messense (approved):

ngoldbaum:
Thanks for merging! Out of curiosity, when are you planning a release? I'm working on wheels for cryptography and needed to pin to maturin's git repo: https://github.com/pyca/cryptography/pull/15063

messense:
Feel free to send release PRs, release process is mostly automated except updating changelog currently requires manual `git cliff`.

PR #3276, opened 2026-07-23
make --find-interpreter skip prerelease Python unless --allow-prereleases

### LINKED ISSUE #1975: Make `--find-interpreter` require opt-in to prereleases
Followup to https://github.com/PyO3/maturin/issues/1960#issuecomment-1969780779 

I think that probably when using `--find-interpreter` maturin should ignore prereleases unless an additional `--allow-prereleases` flag is passed. I think this would help avoid users accidentally uploading wheels for prerelease Python versions before they are ready to deliberately be testing them in CI.

(Maybe it can print that it found them and give a hint to add `--allow-prereleases`, but not build for them.)

tientnc:
## What does this PR do?

- Makes `--find-interpreter` skip prerelease (alpha/beta/rc) Python interpreters, by default, with a `--allow-prereleases` flag to opt back in
- The interpreter probe script now collects `sys.version_info.releaselevel`; each discovered interpreter is tagged as prerelease or not, and the resolver drops prereleases (printing a notice for each) unless `--allow-prereleases` is passed
- Interpreters named explicitly with `-i` are never filtered, since naming one is already an opt-in

**Note:** this changes the default behavior of `--find-interpreter`: prereleases are now skipped unless opted in (which is the behavior asked for in the issue).

## Why was this PR needed?

`--find-interpreter` builds a wheel for every discovered interpreter with no distinction between a final release and a prerelease, which risks accidentally publishing prerelease wheels to PyPI (e.g. in CI). Root cause: the probe script (`get_interpreter_metadata.py`) never collected `releaselevel`, so `InterpreterMetadataMessage` had no field for it and nothing in the Rust discovery path could act on it.  Therefore, the fix starts at collection and threads the value through to the resolver, where the skip decision has access to the user's flags.

## What are the relevant issue numbers?

Closes #1975

## Does this PR meet the acceptance criteria?

- [x] Tests added for new/changed behavior (`test_interpreter_releaselevel_marks_prerelease`)
- [x] All tests passing (`python_interpreter` + `build_options` suites; `cargo fmt --check` + `clippy` clean)
- [x] Follows project style guide
- [x] No breaking changes introduced (additive field/flag; default of `--find-interpreter` now skips prereleases, as requested in the issue)

### DIFF
--- a/src/build_context/builder.rs
+++ b/src/build_context/builder.rs
@@ -340,6 +340,7 @@ impl BuildContextBuilder {
             metadata24.requires_python.as_ref(),
             &user_interpreters,
             build_options.python.find_interpreter,
+            build_options.python.allow_prereleases,
             has_import_lib_support,
         );
         resolver.resolve()
--- a/src/build_options.rs
+++ b/src/build_options.rs
@@ -22,6 +22,11 @@ pub struct PythonOptions {
     #[arg(short = 'f', long, conflicts_with = "interpreter")]
     pub find_interpreter: bool,
 
+    /// Also build for prerelease (alpha/beta/rc) interpreters found by
+    /// `--find-interpreter`, which are skipped by default.
+    #[arg(long)]
+    pub allow_prereleases: bool,
+
     /// Which kind of bindings to use.
     #[arg(short, long)]
     pub bindings: Option<Bindings>,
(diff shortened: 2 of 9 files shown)


tientnc:
The one red check, `Test (windows-11-arm, 3.15-dev)`, failed during environment setup, before the build or test steps ran, with:

```
Fatal error. Internal CLR error. (0x80131506)
pwsh.exe failed with exit code 3221225477   # 0xC0000005, access violation
```

This happened while extracting a tool archive on the ARM Windows runner, so it looks like a transient runner issue rather than anything in this change. The same `3.15-dev` path passes on `Test (ubuntu-22.04-arm, 3.15-dev)`. I suppose that job can be re-run when convenient? Happy to rebase or adjust if you'd prefer!

PR #3277, opened 2026-07-24
add uv cache keys to project template

davidhewitt:
I find myself often using `uv run` to build / test projects these days.

Every time I forget that `uv` has its own caching layer that is by default not quite right for Rust projects. Adding `cache-keys` configuration to be derived from the Rust src helps a lot with this.

Might not be quite right for "python" project layout - we might want to tweak this before merge? Just wanted to kick off the discussion.

### DIFF
--- a/src/templates/pyproject.toml.j2
+++ b/src/templates/pyproject.toml.j2
@@ -35,3 +35,13 @@ exclude = [
     "**/*.pyo"
 ]
 {% endif -%}
+
+[tool.uv]
+cache-keys = [
+    { file = "pyproject.toml" },
+    { file = "Cargo.toml" },
+    { file = "Cargo.lock" },
+    { file = "**/*.rs" },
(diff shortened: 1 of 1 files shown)


src/templates/pyproject.toml.j2
```diff
@@ -35,3 +35,13 @@ exclude = [
     "**/*.pyo"
 ]
 {% endif -%}
+
+[tool.uv]
+cache-keys = [
+    { file = "pyproject.toml" },
+    { file = "Cargo.toml" },
+    { file = "Cargo.lock" },
+    { file = "**/*.rs" },
```
konstin, line 44: That glob would make uv iterate over all files and directories in the project dir, we should use something more targeted: https://github.com/astral-sh/uv/blob/3010295ae7ff572de459987ad70db315a62ecd61/pyproject.toml#L117-L124

PR #3282, opened 2026-07-30
fix(generate-stubs): write stubs into the module's package directory

jonasdedden:
## Problem

`maturin generate-stubs --out DIR` ignores the package path in `module-name`, so the stub is written somewhere no type checker will look.

With

```toml
[tool.maturin]
python-source = "python"
module-name = "disco.parquet._lib"
```

`maturin generate-stubs -o stubs` writes:

```
stubs/_lib.pyi
```

but the extension is imported as `disco.parquet._lib`, and `maturin build --generate-stubs` puts the very same file at `disco/parquet/_lib.pyi` inside the wheel. The `--out` tree therefore cannot be copied over the python source tree, which is the obvious thing to want from a command whose only job is to write stub files to a directory.

The cause is that `commands/generate_stubs.rs` builds its output path from `project_layout.extension_name` alone (`_lib`), while the wheel path derives a module directory from `rust_module` relative to the python source root (`disco/parquet`).

## Fix

Give `ProjectLayout` the `base_path()` / `module_dir()` accessors that `binding_generator` was computing inline, and use them from `generate-stubs` so the two cannot drift. The command now lays files out exactly as `--generate-stubs` lays them out inside a wheel:

| layout | before | after |
|---|---|---|
| mixed, single stub | `{ext}.pyi` | `{module_dir}/{ext}.pyi` |
| mixed, several stubs | `{ext}/…` | `{module_dir}/{ext}/…` |
| pure Rust, several stubs | `{ext}/…` | `{ext}/…` (unchanged) |
| pure Rust, single stub | `{ext}.pyi` | `{ext}/__init__.pyi` |

The last row is a second, smaller inconsistency the same code path had. For a pure Rust project maturin generates the package *around* the extension, so the wheel contains `{ext}/__init__.py`, `{ext}/{ext}.so` and `{ext}/__init__.pyi` — the stub is that package's own `__init__.pyi` and is not inlined. `generate-stubs` inlined it unconditionally, which only happened not to show up because the existing test crate emits two stub files and so never took that branch. Fixing only the missing directory would have made this case worse (`{ext}/{ext}.pyi`), so both are handled together.

`binding_generator` behaviour is unchanged; it just calls the extracted accessors now.

## Tests

`abi3_generate_stubs_mixed_py_subdir` covers a dotted `module-name` using the existing `pyo3-stub-generation-mixed-py-subdir` crate, which until now was only exercised through the wheel tests. It fails on `main`:

```
thread 'integration::abi3_generate_stubs_mixed_py_subdir' panicked at tests/common/other.rs:816:5:
assertion failed: `(left == right)`
```

and passes with this change. The existing `abi3_generate_stubs` case is unaffected — that crate is pure Rust with two stub files, which is the one row of the table that does not move.

### DIFF
--- a/src/binding_generator/mod.rs
+++ b/src/binding_generator/mod.rs
@@ -390,23 +390,8 @@ where
         .context("Failed to add the python module to the package")?;
     }
 
-    let base_path = context
-        .project
-        .project_layout
-        .python_module
-        .as_ref()
-        .map(|python_module| python_module.parent().unwrap().to_path_buf());
-
-    let module = match &base_path {
-        Some(base_path) => context
-            .project
-            .project_layout
-            .rust_module
-            .strip_prefix(base_path)
-            .unwrap()
-            .to_path_buf(),
-        None => PathBuf::from(&context.project.project_layout.extension_name),
-    };
+    let base_path = context.project.project_layout.base_path();
+    let module = context.project.project_layout.module_dir();
 
     {
         let mut dest = InstallDest::new(writer, context.project.editable, base_path.as_deref());
--- a/src/commands/generate_stubs.rs
+++ b/src/commands/generate_stubs.rs
@@ -1,6 +1,7 @@
 use anyhow::Result;
 use fs_err as fs;
 use maturin::{BuildOptions, BuildOrchestrator, CargoOptions, OutputOptions, PythonOptions};
+use std::collections::HashMap;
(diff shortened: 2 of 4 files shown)


Tpt (approved):
Make sense to me! Thank you!

@messense this is technically a breaking change but this is an experimental feature,  shipping it in a minor release is fine imho

PR #3284, opened 2026-08-01
Fix group not cwd

### LINKED ISSUE #3283: maturin develop `--manifest-path` bug when `pyproject` contains `dependency-groups`
### Bug Description

```
maturin develop --manifest-path subfolder/Cargo.toml
🐍 Found CPython 3.14 at /Github/maturin-manifest-path-group-dev-bug/venv/bin/python
🔗 Found pyo3 bindings
ERROR: pyproject.toml not found. Cannot resolve '--group' option.
💥 maturin failed
  Caused by: pip install --group finished with "exit status: 1"
```
if you remove

```
[dependency-groups]
dev = [
    "maturin>=1.14.1",
]
```

from the `pyproject.toml`, it works again. Maturin should pass to pip the path to the pyproject.toml

0.14.1

3.14.5

26.2

pyo3

### Does `cargo build` work?

- [x] Yes, it works

- [ ] Yes

### Steps to Reproduce

MRE, clone and follow the readme https://github.com/kraktus/maturin-manifest-path-group-dev-bug

kraktus:
fix https://github.com/PyO3/maturin/issues/3283, by always adding the absolute path of the pyproject.toml to the group

The code was mainly authored by deepseek v4 flash, but reviewed personally.

And personally tested that the local implementation resulting from the fix, fix my issue, with and without the workaround used before (pre-pending the absolute path to the group), making the change BC compatible.

### DIFF
--- a/src/pgo.rs
+++ b/src/pgo.rs
@@ -215,15 +215,13 @@ impl PgoContext {
                 .and_then(|p| p.dependency_groups.as_ref())
                 .is_some_and(|dg| dg.0.contains_key("dev"));
             if has_dev_group {
-                let project_dir = build_context
-                    .project
-                    .pyproject_toml_path
-                    .parent()
-                    .context("Failed to get project directory")?;
+                let pyproject_group = format!(
+                    "{}:dev",
+                    dunce::simplified(&build_context.project.pyproject_toml_path).display()
+                );
                 debug!("Installing dev dependency group");
                 let status = Command::new(&venv_python)
-                    .args(["-m", "pip", "install", "--group", "dev"])
-                    .current_dir(project_dir)
+                    .args(["-m", "pip", "install", "--group", &pyproject_group])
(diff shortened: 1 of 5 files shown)


src/pgo.rs
kraktus, line 210: note that UV now support groups https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-groups, so could be resolved, but didn't touch because don't know which versions of uv is supported

kraktus:
the ruff failure is due to ruff update 0.16 enabling by default many more lints

messense (commented):
Can we reuse pyo3-pure instead? Adding the a new test crate to test this does not pull its weight. Thanks

kraktus:
Will do

messense (approved):

messense:
Looks like github actions is having issue.

PR #3296, opened 2026-08-26
Fix spurious cross-compilation metadata warnings

woodruffw:
I hit this warning when doing some cross-compilation of wheels with maturin.

TL;DR: when cross-compiling (e.g. Windows-on-Linux), Cargo may compile host-only dependencies omitted from target-filtered cargo metadata. Maturin would spuriously warn about their artifacts. The fix is to collect artifacts only for the package being packaged, identified by its package ID.

I've added a wasm cross build to the tests that ensures that the warning no longer occurs.


messense:
The Alpine test is failing, needs to install the Rust target.

PR #3298, opened 2026-09-01
Fix MinGW stable ABI library names in generated PyO3 config

### LINKED ISSUE #3291: abi3 Windows builds with pyo3 >= 0.29 are unimportable on MSYS2/MinGW (python3.dll vs libpython3.dll)
I ran into this issue during a dependebot update (https://github.com/ariostas/cygv/pull/86). I looked into it a bit with Claude, and it found that some other repos have ran into the same issue. The output from Claude is attached below.

:robot: _AI text below_ :robot:

## Summary

Building an abi3 extension module against pyo3 >= 0.29 in an MSYS2/MINGW64 environment produces a `.pyd` that links and installs without any error but cannot be imported:

```
ImportError: DLL load failed while importing cygv: The specified module could not be found.
```

The extension's import table names `python3.dll`, which does not exist on MSYS2 — it ships `libpython3.dll`. The same project builds and imports fine with pyo3 0.28.

## Environment

- maturin 1.14.1 (`mingw-w64-x86_64-python-maturin`), driven by `pip install --no-build-isolation .`
- pyo3 0.29.2, `features = ["abi3-py310"]`
- MSYS2 MINGW64, `mingw-w64-x86_64-python` 3.14.7, `mingw-w64-x86_64-rust` 1.97.0
- Failing run: https://github.com/ariostas/cygv/actions/runs/31948115189/job/95167262967

## Root cause

`set_env_vars` computes `force_target_abi = stable_abi.is_some() && pyo3_ver >= 0.29.0` ([`src/compile.rs:872-873`](https://github.com/PyO3/maturin/blob/v1.14.1/src/compile.rs#L872-L873)) and, when it is set, writes its own `PYO3_CONFIG_FILE` rather than letting pyo3 probe the interpreter ([`src/compile.rs:916-944`](https://github.com/PyO3/maturin/blob/v1.14.1/src/compile.rs#L916-L944)). With pyo3 0.28 that branch is inert, so the interpreter got probed as usual.

The generated config ([`src/python_interpreter/config.rs:460-534`](https://github.com/PyO3/maturin/blob/v1.14.1/src/python_interpreter/config.rs#L460-L534)) carries `implementation`, `version`, `target_abi`, `shared`, `build_flags` and `pointer_width` — but no `lib_name`, and nothing that records that the interpreter is a MinGW one. `lib_name` is only written for Android (lines 519-526).

pyo3 then fills the gap:

```rust
// pyo3-build-config 0.29.2, src/impl_.rs:692
if config.lib_name.is_none() {
    config.lib_name = Some(default_lib_name_for_target(config.target_abi, target));
}
```

and `default_lib_name_for_target` hardcodes `mingw = false`:

```rust
// src/impl_.rs:2372
fn default_lib_name_for_target(abi: PythonAbi, target: &Triple) -> String {
    if target.operating_system == OperatingSystem::Windows {
        default_lib_name_windows(abi, false, false).unwrap()
```

For abi3 that yields `python3`. pyo3 0.29 made `raw-dylib` the default on Windows and `python3` is in its known-DLL list, so the extension gets `#[link(name = "python3", kind = "raw-dylib")]` and the DLL name goes straight into the import table. raw-dylib needs no import library, so **nothing fails at build time** — the failure only shows up at `import`.

With pyo3 0.28 the same `python3` name went through `-lpython3`; the MinGW linker resolved that to `/mingw64/lib/libpython3.dll.a`, and the import table got the correct `libpython3.dll`. That is why this regressed silently.

This is not the same bug as PyO3/pyo3#6157, and it survives its fix (PyO3/pyo3#6185, released in pyo3 0.29.2). That fix teaches `default_lib_name_windows` to add the `lib` prefix when `mingw` is true, but the flag is only ever true when pyo3 probes a live interpreter — which is exactly what the config file generated here replaces. The run linked above is pyo3 0.29.2, i.e. with #6185 already applied.

## Reproducer

In an MSYS2 MINGW64 shell, with any pyo3 project using an `abi3-py3*` feature:

```bash
pip install --no-build-isolation .
python -c "import your_module"   # ImportError: DLL load failed
objdump -p .../your_module.pyd | grep 'DLL Name'   # python3.dll
```

## Suggested fix

Write `lib_name` for MinGW interpreters the way it is already written for Android. maturin has the signal in the same struct: `InterpreterConfig::ext_suffix` is `.cp314-mingw_x86_64_msvcrt_gnu.pyd` on MSYS2 (and `SOABI` likewise starts with `cp314-mingw`), so something along the lines of

```rust
if target.is_windows() && self.ext_suffix.contains("-mingw") {
    let lib_name = if stable_abi.is_some() {
        "libpython3".to_string()
    } else {
        format!("libpython{}.{}", self.major, self.minor)
    };
    write!(content, "\nlib_name={lib_name}").unwrap();
}
```

would give pyo3 the name it would have derived from the interpreter itself. (pyo3 cannot infer this from the target triple: `x86_64-pc-windows-gnu` is also used to cross-compile against a regular MSVC CPython, where `python3.dll` is right.)

## Downstream

MSYS2 has been hitting this: msys2/MINGW-packages#30442 (`_polars_runtime.pyd` links against `python3.dll` instead of `libpython3.dll`) for a maturin abi3 build, and msys2/MINGW-packages#29942 for the setuptools-rust equivalent. Both PKGBUILDs now carry `export PYO3_USE_RAW_DYLIB=0`.

## Workaround

Setting `PYO3_USE_RAW_DYLIB=0` restores the pyo3 0.28 behaviour and fixes it.

Udonburo:
## Summary

Fixes #3291.

For PyO3 0.29+ stable ABI builds, maturin generates a `PYO3_CONFIG_FILE`
instead of leaving PyO3 to probe the selected interpreter directly.

On native MSYS2/MinGW, maturin retains the interpreter's
`sysconfig.get_platform()` value, but the generated config previously lost
that distinction. PyO3 then selected the ordinary Windows `python3.dll`
raw-dylib path instead of the MinGW stable ABI library.

This change uses the selected interpreter's platform metadata when generating
the config:

- MinGW + abi3: `lib_name=libpython3`
- MinGW + abi3t: `lib_name=libpython3t`

It does not infer MinGW from the Rust target triple or `EXT_SUFFIX`.
Non-MinGW Windows interpreters and the generic `InterpreterConfig` output are
unchanged.

## Tests

Added regression coverage for:

- MinGW CPython + abi3
- MinGW CPython + abi3t
- a non-MinGW interpreter platform using the same
  `x86_64-pc-windows-gnu` Rust target

Local checks:

- `cargo test --lib` (257 passed)
- `cargo check --all-features --all-targets`
- `cargo fmt --all -- --check`
- `git diff --check`

I also validated the abi3 case on native MSYS2/MINGW64 with PyO3 0.29.2.
Before the change, the installed `.pyd` referenced `python3.dll` and failed to
load. After the change, the generated config contained
`lib_name=libpython3`, the installed `.pyd` referenced `libpython3.dll`, and
the module loaded successfully.

Native MSYS2 pip does not accept the generated `win_amd64` wheel tag directly,
so the exact unchanged wheel was installed with an explicit platform override
for this loader-level validation. The abi3t case is covered by unit tests only.

On this MSYS2 toolchain, the full clippy command fails on the unchanged
upstream base with the same 11 unused-import diagnostics in
`src/module_writer/mod.rs`.

### DIFF
--- a/src/python_interpreter/mod.rs
+++ b/src/python_interpreter/mod.rs
@@ -1,8 +1,8 @@
 pub use self::config::InterpreterConfig;
-use crate::Target;
 use crate::auditwheel::PlatformTag;
 use crate::bridge::{ABI3T_MINIMUM_PYTHON_MINOR, StableAbiKind};
 use crate::target::WheelTag;
+use crate::{StableAbi, Target};
 use anyhow::{Result, bail};
 use std::borrow::Cow;
 use std::ffi::OsStr;
@@ -151,6 +151,39 @@ impl PythonInterpreter {
             StableAbiKind::Abi3 => self.has_abi3(),
             StableAbiKind::Abi3t => self.has_abi3t(),
         }
+    }
+
+    /// Generate PyO3 config file content for this interpreter.
+    ///
+    /// When maturin supplies `PYO3_CONFIG_FILE`, PyO3 does not live-probe the
+    /// interpreter. Use `platform` (from `sysconfig.get_platform()`) to preserve
+    /// the `libpython3` / `libpython3t` names PyO3's native MinGW discovery
+    /// would otherwise select.
+    pub(crate) fn pyo3_config_file(
+        &self,
+        target: &Target,
+        stable_abi: Option<StableAbi>,
+        use_target_abi: bool,
+    ) -> String {
+        let mut content = self
+            .config
+            .pyo3_config_file(target, stable_abi, use_target_abi);
+        if target.is_windows()
+            && self.interpreter_kind == InterpreterKind::CPython
+            && self
+                .platform
+                .as_deref()
+                .is_some_and(|platform| platform.starts_with("mingw"))
+            && let Some(stable_abi) = stable_abi
+        {
+            let lib_name = match stable_abi.kind {
+                StableAbiKind::Abi3 => "libpython3",
+                StableAbiKind::Abi3t => "libpython3t",
+            };
+            content.push_str("\nlib_name=");
+            content.push_str(lib_name);
+        }
+        content
     }
 
     /// Formats this interpreter as a value for uv's `--python` argument.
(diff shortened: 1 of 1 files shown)


Udonburo:
The `ubuntu-latest, 3.15t-dev` job failed while two tests were creating Conda environments concurrently, before exercising this change. Both failures came from missing `.conda.partial` files in the shared package cache (`create_conda_env`); all other jobs, including the Windows and other 3.15t jobs, passed.

I don't have permission to rerun Actions in the upstream repository. Could a maintainer please rerun the failed job?

messense:
I don't quite understand, the new functions are only used in tests but not wired in other places?

Udonburo:
Thanks for checking. The existing call in `configure_pyo3_env` was resolving to the new method on `PythonInterpreter`, but that wasn't clear from the diff. I've renamed it to `pyo3_config_file_with_platform` and updated the call site to make the connection explicit.

messense (approved):

PR #3299, opened 2026-09-09
properly normalize / validate abiflags and include in environment signature

davidhewitt:
`maturin` currently generates `PYO3_ENVIRONMENT_SIGNATURE` without accounting for abiflags. This in particular means that builds switching between regular and free-threaded Python can fail to invalidate `pyo3-ffi`'s cache and lead to compilation selecting the wrong ABI.

This PR corrects the generated environment signature, and attempts to make abiflag handling correct throughout `maturin`. In particular I found the following other deficiencies:
- Parsing `PYO3_CONFIG_FILE` did not infer (or validate) missing ABI flags from `Py_GIL_DISABLED` or `Py_DEBUG`.
- The default CPython extension suffix generated from `PYO3_CONFIG_FILE` omitted free-threaded and debug flags. (This would produce a wheel with an invalid extension for the free-threaded/debug interpreter.)
- Cross-compilation sysconfig parsing did not infer missing ABI flags from build settings.
- Interpreter discovery generally did not account for the possibility of the `d` debug abiflag.

### DIFF
--- a/src/python_interpreter/abiflags.rs
+++ b/src/python_interpreter/abiflags.rs
@@ -8,12 +8,39 @@ use anyhow::{Result, bail, ensure};
 
 use super::discovery::InterpreterMetadataMessage;
 
+pub(super) fn normalize_abiflags(
+    abiflags: &mut String,
+    debug: bool,
+    gil_disabled: bool,
+) -> Result<()> {
+    if debug && !abiflags.contains('d') {
+        abiflags.insert(0, 'd');
+    }
+    if gil_disabled && !abiflags.contains('t') {
+        abiflags.push('t');
+    }
+    validate_abiflags(abiflags, debug, gil_disabled)
+}
+
+pub(super) fn validate_abiflags(abiflags: &str, debug: bool, gil_disabled: bool) -> Result<()> {
+    for (flag, enabled, name) in [
+        ('d', debug, "Py_DEBUG"),
+        ('t', gil_disabled, "Py_GIL_DISABLED"),
+    ] {
+        ensure!(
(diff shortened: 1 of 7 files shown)


src/python_interpreter/abiflags.rs
```diff
@@ -137,6 +164,39 @@ pub(super) fn calculate_abi_tag(ext_suffix: &str) -> Option<String> {
...
+    #[test]
+    fn test_infer_abiflags() {
+        for (flags, debug, gil_disabled, expected) in [
+            ("", false, false, ""),
+            ("", true, false, "d"),
+            ("", false, true, "t"),
+            ("", true, true, "dt"),
+            ("t", true, true, "dt"),
+            ("d", true, true, "dt"),
+            ("dt", true, true, "dt"),
```
messense: https://github.com/pypa/packaging/blob/10590c194edb33c82f84a127883d6097c56b7840/src/packaging/tags.py#L376-L390

we should match `packaging` order? like it should be `td` instead of `dt`?

```python
yield f"cp{version}{threading}{debug}{pymalloc}{ucs4}"
```

messense (commented):
should we also update here? https://github.com/PyO3/maturin/blob/faad068feb480028e53073c17c1973226854e408/src/cross_compile.rs#L65-L74

davidhewitt:
Thanks, addressed both points. To keep the ordering consistent I simplified to either generate abiflags matching packaging if no flags present, otherwise I just validate the user-supplied flags at least match the sysconfigdata keys.

davidhewitt:
conda bug looks possibly unrelated?

messense (approved):

PR #3300, opened 2026-09-10
Re-order maturin.rs for clarity

0atman:
I found it a little confusing to figure out what maturin is from reading maturin.rs, I think re-ordering the page could solve this easily:

I've moved the sponsors to the bottom of the page, where sponsors are typically placed, and removed the introduction paragraph that tells the user how to use a web page, but not what maturin.rs is.

The readme is fantastic, starting with "formally known as", and then gets straight to an introduction. I think that should go FIRST on the page?


0atman:
TL;DR: maturin

Is now:

image

PR #3302, opened 2026-09-12
fix: lossy-decode external command output in interpreter discovery

### LINKED ISSUE #3301: Windows interpreter discovery panics on non-UTF-8 py/conda output (from_utf8 unwrap in find_all_windows)
## Summary

Windows interpreter discovery panics on non-UTF-8 child output. `find_all_windows` calls `str::from_utf8(&output.stdout).unwrap()` on the stdout of `py --list-paths` and `conda info -e` — both of which echo install paths that routinely contain non-UTF-8 bytes on localized Windows setups — aborting the entire build instead of skipping an undecodable line.

## Location

- File: `src/python_interpreter/discovery.rs`
- Function: `find_all_windows`
- Sites: `str::from_utf8(&output.stdout).unwrap()` on the `py --list-paths` output and on the `conda info -e` output

Relevant code path (static analysis of current `main`):

```rust
if let Ok(output) = execution {
    ...
    let stdout = str::from_utf8(&output.stdout).unwrap();  // panics on non-UTF-8
    for line in stdout.lines() { ... }
}
...
let conda_info = Command::new("conda").arg("info").arg("-e").output();
if let Ok(output) = conda_info {
    let lines = str::from_utf8(&output.stdout).unwrap().lines();  // same
    ...
}
```

## Problem

`Command::output` stdout is `Vec<u8>` with no encoding guarantee; on Windows the console codepage output of `py`/`conda` reflects filesystem paths (e.g. `C:\Users\<name>\...` with a CJK or accented username under a legacy codepage) that are not valid UTF-8. `from_utf8().unwrap()` turns one such path into a process-wide panic in the middle of `maturin build`, even though the undecodable line could simply be skipped (or lossily decoded) — the loop already skips non-matching and nonexistent paths line by line.

## Trigger / Reproduction

Based on static analysis (no Windows run performed):

1. On Windows with a Python install under a path whose bytes are not valid UTF-8 (localized username/install dir), run any `maturin build` that triggers interpreter discovery.
2. `py --list-paths` prints that path; `from_utf8` fails; `.unwrap()` panics and the build aborts before compiling anything.

Note: this is a static-analysis finding; I did not execute discovery on such a machine.

## Expected Behavior

Undecodable output should degrade gracefully — `String::from_utf8_lossy` plus the existing per-line skip logic — so one exotic path cannot kill discovery of all other interpreters.

## Actual Behavior

A single non-UTF-8 byte in either command's stdout panics the whole build.

## Impact

- `maturin build` unusable on affected Windows machines with no actionable error (panic message points at an internal `unwrap`, not at the offending path).
- Same latent crash in both the `py` launcher and conda enumeration paths.

## Suggested Direction

- Replace both `str::from_utf8(&output.stdout).unwrap()` calls with `String::from_utf8_lossy(&output.stdout)`; the downstream line loop already tolerates junk lines. No behavior change for valid UTF-8 output.

## Evidence

- Source via API: `discovery.rs::find_all_windows` shows both `unwrap()` sites on external-command bytes, surrounded by defensive per-line handling (`if let Some(capture)`, `if !exists { continue }`) that highlights the gap.
- Duplicate check: issue search for `from_utf8 UTF-8 panic windows discovery` returns one unrelated closed feature (delvewheel re-implementation); `non-UTF8 conda py launcher` returns `total_count: 0` — no apparent duplicate.

## Classification

- FACT: two `from_utf8().unwrap()` calls decode external command stdout (verified in source via API).
- INFERENCE: non-UTF-8 install paths therefore panic discovery and abort the build.
- HYPOTHESIS: lossy decoding preserves all current behavior on valid input while removing the crash.

devtechedge:
On Windows, `find_all_windows()` ran `str::from_utf8(&output.stdout).unwrap()` on the raw stdout of `py --list-paths` and `conda info -e`. Localized installations, for example a CJK or accented Windows username under a legacy console codepage, can make those commands emit non-UTF-8 bytes. In that case the `unwrap()` panicked and aborted the whole build instead of just skipping the interpreter.

This replaces both call sites with `String::from_utf8_lossy`, matching the idiom already used in the same file for stderr and error contexts (lines 316 and 389). For valid UTF-8 output nothing changes. Undecodable bytes become U+FFFD, which cannot match the py-launcher or conda path regexes, so affected lines are skipped by the existing per-line logic just like any other junk line.

Added a focused unit test next to the existing discovery tests that feeds non-UTF-8 bytes through the same decode step and asserts the old input would have failed `str::from_utf8` (the panic condition) while the lossy decode preserves the line structure and valid output is decoded unchanged. The `find_all_windows` call sites themselves shell out to external commands, so they are not directly unit-testable; the fail-before/pass-after repro is `str::from_utf8(&[0xff]).unwrap()` panicking on any non-UTF-8 byte.

Verified with `cargo fmt --all -- --check`, `cargo clippy --tests --all-features` (no new warnings from this change), and the `python_interpreter::discovery` test suite.

Fixes #3301


devtechedge:
Quick note on the red checks here, in case they hold up review.

`Test (ubuntu-latest, 3.15-dev)` and `Test (ubuntu-latest, 3.15t-dev)` fail in `cargo test` at `tests/common/mod.rs:428` with `Failed to create conda environment: exit status 1`, from a conda `CondaMultiError` renaming `/usr/share/miniconda/pkgs/*.conda.partial` files.

The `conclusion` job just mirrors those two, so there are only two real failures.

This looks like runner package-cache drift rather than anything in this diff: #3303 fails identically and touches no Rust code, and neither test gets past creating the environment.

Every other leg is green, including the macOS, ARM, and Windows 3.15-dev ones.

messense (approved):

devtechedge:
Thanks for merging, @messense. 🙌

Matching the `String::from_utf8_lossy` idiom already used for stderr in `discovery.rs` kept the fix to two lines, a useful reminder that consistency with the surrounding file is what this repo rewards.

Thanks too for merging through the two red 3.15-dev conda legs; those turned out to be runner cache drift rather than anything in this diff.

Issue #3196, opened 2026-05-22
pixi environment fails with pip installed and uv installed globally


hoxbro:
### Bug Description

I have UV installed outside the pixi environment (in my case with `pixi global`, but should be a general problem). This was likely introduced in https://github.com/PyO3/maturin/pull/3165, I think their needs to be a check if the uv is installed inside the pixi environment before trying to use it over pip.

Take the following `pixi.toml` file. 

``` toml
[workspace]
channels = ["conda-forge"]
name = "maturin-pixi"
platforms = ["linux-64", "osx-arm64", "win-64"]

[tasks]
build = "maturin develop"

[environments]
new-pip = ["maturin-new", "pip"]
new-uv = ["maturin-new", "uv"]
old-pip = ["maturin-old", "pip"]
old-uv = ["maturin-old", "uv"]

[feature.maturin-new.dependencies]
maturin = "==1.13.2"

[feature.maturin-old.dependencies]
maturin = "==1.13.1"

[feature.pip.dependencies]
pip = "*"

[feature.uv.dependencies]
uv = "*"
```

And run the combination
``` bash
#!/usr/bin/env bash
set -x
which uv
pixi run -e new-pip build
pixi run -e new-uv build
pixi run -e old-pip build
pixi run -e old-uv build
```

Will reduce this output:
``` bash
+ which uv
/home/shh/.local/pixi/bin/uv
+ pixi run -e new-pip build
✨ Pixi task (build in new-pip): maturin develop
🐍 Found CPython 3.14 at /home/shh/projects/maturin-pixi/.pixi/envs/new-pip/bin/python
🔗 Found pyo3 bindings
   Compiling pyo3-build-config v0.28.3
   Compiling pyo3-ffi v0.28.3
   Compiling pyo3-macros-backend v0.28.3
   Compiling pyo3 v0.28.3
   Compiling pyo3-macros v0.28.3
   Compiling maturin-pixi v0.1.0 (/home/shh/projects/maturin-pixi)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 3.94s
📦 Built wheel for CPython 3.14 to /tmp/.tmpNRZ9wn/maturin_pixi-0.1.0-cp314-cp314-linux_x86_64.whl
💥 maturin failed
[...]

PR #3306, opened 2026-09-21
fix: avoid global uv fallback in pixi environments

1998LJ:


messense:
Thanks for the PRs, could you write a brief pr description, like reference existing issue?

1998LJ:
Thanks. This fixes #3196.

Brief summary: in pixi environments, automatic uv discovery is now
scoped to the active environment so a globally installed uv cannot
hijack an environment that only provides pip. Explicit --uv usage and
uv-created virtual environments keep the existing global fallback
behavior.

The change also adds coverage for the pixi/uv discovery decision and
PATH filtering.

